PR #157, opened 2025-04-12
ast: Add default recursive visitor and refactor AstDumper

flofriday:
In the past we had some problems where multiple visitors  were implemented which all re implemented the traversing through the tree. The Problems arose when a visitor skipped parts of the sub-tree.

By providing a single recursive implementation for the traversing we can minimize bugs since there are less implementations where bugs can hide and simplifying the code since the default `RecursiveAstVisitor` has only the task of traversing the tree and now the `AstDumper` has mostly to focus on dumping which minimizes the mental load during development.

The default implementation should be able to be used for the `AstDumper`, `SymbolCollector`, `SymbolResolver` and maybe: `Ungrouper` and `ModelRemover`.
More complex traversals through the tree where the order is more important and traversals produce results will still rely on their own implementation like the `Typechecker` and `ViamLowering`.

For now this PR only refactors the AstDumper as a proof of concept, but I want to do the Symboltable in a different PR since this is already getting quite large.


Jozott00 (commented):
Looks good to me!

I would suggest refactoring formats, such that format fields are also definitions. I don't see any reason why they shouldn't be definitions.

vadl/main/vadl/ast/AstDumper.java
```diff
@@ -119,286 +149,6 @@ public Void visit(FormatDefinition definition) {
     return null;
   }
```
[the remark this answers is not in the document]
flofriday: Yes, you are right :)

vadl/main/vadl/ast/Definition.java
```diff
@@ -333,257 +348,265 @@ public Type type() {
...
+ * {@code
+ * format ABC : Bits<8> =
+ *   { A  : Bits<2>
+ *   , B  : Bits<6>
+ *   , C = A as Bits<4>    // This is a derived format field
+ *   }
+ * }
+ * </pre>
+ */
+class DerivedFormatField extends FormatField {
```
Jozott00, line 515: Maybe we should rename it `AccessField` or `AccessFunctionField`? As the official term is `AccessFunction`.

vadl/main/vadl/ast/Definition.java
```diff
@@ -333,257 +348,265 @@ public Type type() {
...
+}
+
+class FormatDefinition extends Definition implements IdentifiableNode, TypedNode {
+  IdentifierOrPlaceholder identifier;
+  @Child
+  TypeLiteral typeLiteral;
+  @Child
+  List<FormatField> fields;
+  @Child
+  List<AuxiliaryField> auxiliaryFields;
```
Jozott00, line 598: minor: Maybe we should add a comment that auxiliary fields are the encoding and predicate block fields.

vadl/main/vadl/ast/Definition.java
```diff
@@ -1291,13 +1294,75 @@ public ConcreteRelationType type() {
...
+      }
+      var that = (RelationTypeLiteral) obj;
+      return Objects.equals(this.argTypes, that.argTypes)
+          && Objects.equals(this.resultType, that.resultType);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(argTypes, resultType);
+    }
```
Jozott00, line 1332: suggestion: In many cases the equals and hashcode methods are trivial as they’re just wrapping the children. We could use your annotation processor to also generate the `equals` and `hashcode` functions automatically. 

For fields that are no children, we could have a `collectValues` method that returns a (nullable) list of all values that are no children, which can be used for equals/hashcode computation. (or something similar)

java-annotations/main/vadl/javaannotations/ast/Child.java
```diff
@@ -14,27 +14,21 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 /**
- * A simple VADL parser entry point to collect timings etc.
+ * The Child annotation indicates that this field is a child of this node and is owned by it.
+ * Identifiers are only children if they aren't used to identify the definition.
+ *
+ * <p>Annotating with this annotaion will include the field in the NodeChildrenRegistry, which
+ * is generated by the {@link ChildAnnotationProcessor}.
```
Jozott00, line 29: minor: This could have a bit more documentation about the generated `NodeChildrenRegistry` and how to use it. E.g. several nodes have a custom `children()` implementation as they’re too complicated for the generator. New users wouldn't know the limitations of the `@Child` annotation.

vadl/test/vadl/ast/AstDumpTests.java
```diff
@@ -41,29 +42,43 @@ public class AstDumpTests {
...
-          .map(path -> DynamicTest.dynamicTest("Parse " + path.getFileName(),
-              () -> assertDumpEquality(path)));
+    var unitTestStream = loadVadlFiles("dumps").stream()
+        .filter(path -> path.getFileName().toString().endsWith(".vadl"))
+        .map(path -> DynamicTest.dynamicTest("Parse " + path.getFileName(),
+            () -> assertDumpEquality(path, dumpPath(path))));
+
+    var realWorldTestStream = Stream.of(
+            Pair.of(
+                getResourcePath("testSource/sys/risc-v/rv64im.vadl"),
```
Jozott00, line 52: major: I am not sure if we should use the reference implementation for the snapshot test. It might be changed frequently, so we need to update the reference snapshot on every change. And in that case, we wouldn't really be able to check if the changes in the reference snapshot are actually correct.

I would suggest having a duplicated version of the `rv64im.vadl`, used only by the frontend. In the `resources/macros` there is already a `rv3264im.vadl` that can be used/extended.

vadl/main/vadl/ast/Definition.java
flofriday: There is quite a large comment where the class is defined (which is only 12 lines below the highlighted line) so do you think we need one here too? 🤔

java-annotations/main/vadl/javaannotations/ast/Child.java
flofriday: True, I should definitely describe that better. 👍

vadl/test/vadl/ast/AstDumpTests.java
flofriday: I'm also not sure. 
I like that we test the example specifications we provide so that they never are out of sync with what the tool can understand.

However, maybe we should only do that with more complete tests (like where we run the whole frontend on the specs to verify it doesn't crash and produces a VIAM), and keep copies for tests that would otherwise require frequent adaption.

Jozott00: I mean, the frontend and VIAM generation are already part of the integration test with the ISS and LCB. If the frontend produces a wrong output or crashes, they won't complete either.

vadl/main/vadl/ast/Definition.java
flofriday: I'm not sure, there are quite a few cases (with identifiers) where they are part of the `equals` and `hashCode` and not part of the children.

And I'm not convinced that it would provide enough help to justify adding a method that is only used by two utility methods.

PR #160, opened 2025-04-13
frontend: Add exception definition support

Jozott00:
This PR adds parameters to Exceptions and supports raising them. 

```
    register mstatus: Regs

    enumeration ExcCode: Bits<64> =
    { 
    , BREAKPOINT = 0x3
    }

    exception Exc(code: Bits<64>) = {
      mcause := code
    }

    instruction TEST : Itype =
      raise Exc(ExcCode::BREAKPOINT)
```

### DIFF
--- a/vadl/main/vadl/ast/BehaviorLowering.java
+++ b/vadl/main/vadl/ast/BehaviorLowering.java
@@ -17,6 +17,8 @@
 package vadl.ast;
 
 
+import static vadl.error.Diagnostic.error;
+
 import com.google.errorprone.annotations.concurrent.LazyInit;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -263,7 +265,7 @@ class BehaviorLowering implements StatementVisitor<SubgraphContext>, ExprVisitor
     // FIXME: Wrap input and output in casts
     // FIXME: Add conditions based on annotations
     var regFile = (RegisterFile) viamLowering.fetch(regFileDef).orElseThrow();
-    var regfileRead = new WriteRegFileNode(
+    var regfileWrite = new WriteRegFileNode(
         regFile,
         new FuncParamNode(indexParam),
         new FuncParamNode(valueParam),
@@ -271,7 +273,7 @@ class BehaviorLowering implements StatementVisitor<SubgraphContext>, ExprVisitor
         null
     );
 
-    var end = graph.addWithInputs(new ProcEndNode(new NodeList<>(regfileRead)));
+    var end = graph.addWithInputs(new ProcEndNode(new NodeList<>(regfileWrite)));
     graph.addWithInputs(new StartNode(end));
 
     // FIXME: Modify based on annotations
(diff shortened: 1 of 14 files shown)


AndreasKrall (dismissed):
Seems fine to me. Perhaps one of the suggested reviewers also take a look on the diffs.

AndreasKrall (approved):
Fine.

vadl/main/vadl/ast/BehaviorLowering.java
```diff
@@ -1199,15 +1203,39 @@ public SubgraphContext visit(PlaceholderStatement statement) {
...
+    } else {
+      // FIXME: Add to a global store so that ISA can get a list of all exceptions
+      var name = statement.viamId + "::anonymousException";
+      exception = new ExceptionDef(
+          viamLowering.generateIdentifier(name, statement.statement),
+          new vadl.viam.Parameter[] {},
+          new BehaviorLowering(this.viamLowering).getProcedureGraph(statement.statement, name),
+          ExceptionDef.Kind.ANONYMOUS
+      );
+    }
```
Jozott00, line 1235: @flofriday it is important that these anonymous exceptions are included in the VIAM as a definition. However, this might require a structural change, so I haven’t implemented this.

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -1739,11 +1740,13 @@ private void visitIdentifiable(Expr expr) {
...
       innerName = identifier.name;
       fullName = identifier.name;
+      identifier.target = origin;
     } else if (expr instanceof IdentifierPath path) {
       origin = requireNonNull(expr.symbolTable).findAs(path, Node.class);
       var segments = path.pathToSegments();
       innerName = segments.get(segments.size() - 1);
       fullName = path.pathToString();
+      path.target = origin;
     } else {
```
Jozott00, line 1750: @flofriday This adds the nodes as target to the `Identifier` and `IdentifierPath`. It makes it a lot easier to find them later (e.g. when handling `raise Exc` in the lowering).

AndreasKrall (approved):
Looks fine to me, interesting that there were Read/Write naming errors.

vadl/main/vadl/ast/BehaviorLowering.java
```diff
@@ -17,6 +17,8 @@
 package vadl.ast;
 
 
+import static vadl.error.Diagnostic.error;
```
flofriday, line 20: **question:** Do we need that import? Is it really that more useful or have we started doing that in other parts of the codebase? 

And if so shouldn't we replace all `Diagnostic.error` with `error`?

vadl/main/vadl/ast/BehaviorLowering.java
flofriday: It does but it should be easy to add 👍

vadl/main/vadl/ast/BehaviorLowering.java
Jozott00: In the backend I always directly import the `error` method. 
So it was because I was used to it. However, I can remove it in the frontend.

flofriday: Nonono,  then we will switch to that to make it more consistent with the backend.

PR #162, opened 2025-04-13
chore: Tooling Java Minimum Major Version Check

ppaulweber:
This PR introduces the following changes:

* set minimum required Java 17 major version fix in Gradle setup (relates to docs changed in a905882a450cc9ee53d4b7a42e98e2d2c30b8083)

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -141,8 +141,8 @@ location of your Java installation."
 fi
 
 # Add minimum Java major version check, see: https://github.com/OpenVADL/openvadl/pull/162
-JAVA_VERSION_STRING=$("$JAVACMD" -version 2>&1)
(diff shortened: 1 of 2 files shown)


AndreasKrall:
I do not know too much about gradle. But is it a good idea to use `sed` in a gradle script, then there could be another dependency.

ppaulweber:
> I do not know too much about gradle. But is it a good idea to use `sed` in a gradle script, then there could be another dependency. And I was once hurt by the reference manual `sed` script which changed all my files on my system.

@AndreasKrall I've provided a fixup in 11c61b1dce65e30700230af7ce5b314d1a871ea7 to use `grep` and `cut` to avoid the usage of `sed`.

BTW: there is a usage of `sed` in the `gradlew` script unrelated to this PR, see:
https://github.com/OpenVADL/openvadl/blob/d178e9b9982ed3a3a071d70dd97d3719f8be1ae2/gradlew#L245

Jozott00 (approved):

PR #176, opened 2025-04-17
frontend: Refactor symbol collecting

flofriday:
This only refactors the collection of symbols and distribution of symboltables, the actual resolving is left as future work.


vadl/test/vadl/ast/AsmDescriptionTests.java
```diff
@@ -83,8 +83,8 @@ application binary interface ABI for ISA = {}
   void asmDescriptionDuplicateModifierString() {
     var prog = """
           instruction set architecture ISA = {
-            relocation HI ( symbol : Bits <32> ) -> Bits <16> = ( symbol >> 16 ) & 0xFFFF
-            relocation LO ( symbol : Bits <32> ) -> Bits <12> =   symbol         & 0xFFF
+            relocation HI ( symbol : Bits <32> ) -> Bits <16> = (( symbol >> 16 ) & 0xFFFF) as Bits<16>
+            relocation LO ( symbol : Bits <32> ) -> Bits <12> = (  symbol         & 0xFFF ) as Bits<16>
```
flofriday, line 87: @benjaminkasper99 I put the tests into the check command which also does typechecking and found this issue, I know it's not necessary for this test to be correctly typed but I think we should still aim for all tests to be correctly typed if that's not an aspect we want to test.

vadl/test/vadl/ast/TypecheckerProcessorTest.java
```diff
@@ -41,13 +41,9 @@ void shouldThrow_duplicatedDefinitions() {
...
     var spec = base.formatted(body);
-    var ast = Assertions.assertDoesNotThrow(() -> VadlParser.parse(spec), "Cannot parse input");
-    var typechecker = new TypeChecker();
-    var throwable = assertThrows(DiagnosticList.class, () -> typechecker.verify(ast));
-    TestUtils.assertErrors(throwable,
-        "Contains multiple `firmware` definitions",
-        "Contains multiple `start` definitions"
+    var throwable = Assertions.assertThrows(DiagnosticList.class, () -> VadlParser.parse(spec)
     );
+    TestUtils.assertErrors(throwable, "Symbol name already used: start");
```
flofriday, line 46: @benjaminkasper99 The problem here is that this issue is now already found during symbol resolution instead of the typechecking.

I don't know if you still want to leave this test in but because the old symbol collector skipped some parts of the trees it previously was left as an task for the typechecker.

vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -380,489 +380,264 @@ private void reportAlreadyDefined(String error, SourceLocation location,
...
-        }
+    @Override
+    public void afterTravel(Definition definition) {
+      viamPath.pollLast();
+      symbolTables.pollLast();
+    }
+
+    @Override
+    public Void visit(AsmDescriptionDefinition definition) {
+      // More complex tasks like this require custom handling.
```
flofriday, line 463: **thought:** Only the more complex edge cases must be overriden and they all contain a comment on why it wasn't possible to use the default implementation

vadl/test/vadl/ast/AsmDescriptionTests.java
benjaminkasper99: I agree, but I think the second cast here should then be `as Bits<12>`?

vadl/main/vadl/ast/SymbolTable.java
benjaminkasper99: I think that's a fine solution. Also the adaptations for the `Asm...` definitions look good to me.

vadl/test/vadl/ast/AsmDescriptionTests.java
flofriday: Yes you are right the second one should be 12 Bits

Jozott00 (approved):

benjaminkasper99 (approved):

PR #178, opened 2025-04-17
viam: Refactor Register and RegisterFile to RegisterTensor

Jozott00:
This PR refactors `Register` and `RegisterFile` to a more generic `RegisterTensor`, which may consist of arbitrary many dimensions. 

To achieve this, the `Register` and `RegisterFile` classes become subtypes of `RegisterTensor`, to avoid breaking current code. 

Then all generators must adopt the new `RegisterTensor` and remove usages of all `Register` and `RegisterFile` with their corresponding nodes (read, write). 

We should do the refactoring in separate branches and then merge it into this one.
- [x] Set up new `RegisterTensor` classes
- [x] Adopt in ISS @Jozott00 
- [x] Adopt in LCB @kper 
- [x] Adopt in ASM @benjaminkasper99 
- [x] Adopt in Hardware @linushdot 
- [x] Adopt in VIAM lowering @Jozott00


AndreasKrall:
In the comment you write:

```
 * <p>The dimension list is sorted as in the VADL specification: outer dimensions are at the start,
 * inner ones are at the end.
 * The minimal accessible unit is the innermost dimension as a whole.</p>
```
We need to have also have access to single bits, but this can be done with slicing / concatenation and it can be that the innermost dimension is of format type and not of Bits type. I would extend the comment by one sentence explaining how to access single bits, bit slices or the format fields. We have to support things like that:

```
format f    : Bits<64> = {a : Bits<32>, b : Bits<2><16>}

register r1 : Bits<32><64>   // or Bits<5> -> Bits<64>
register r2 : f<32>          // or Bits<5> -> f

// register accesses
r1
r1(rs1)
r1(rs1)(63)
r1(rs1)(31..0)
r1(rs1)(63,6..0)

r2
r2(rs1)
r2(rs1).a
r2(rs1).b
r2(rs1).b(0)
r2(rs1).b(1)(15)
r2(rs1).b(1)(15,6..0)
```
The register has to support the definition from the beginning, the more complex accesses can be implemented on demand when we need them in an architecture specification (system registers).

vadl/main/vadl/viam/RegisterTensor.java
```diff
@@ -0,0 +1,166 @@
...
+ * A register tensor is a generic representation of all kinds of registers.
+ * It is defined by a list of {@link Dimension}s. This list defines the result types
+ * and index types of registers.
+ *
+ * <p>The dimension list is sorted as in the VADL specification: outer dimensions are at the start,
+ * inner ones are at the end.
+ * The minimal accessible unit is the innermost dimension as a whole.
+ * If users access more specific areas of a register by using formats, slices, or an index on the
+ * innermost dimension, these accesses are resolved by the frontend by using slice
+ * nodes in the graph.</p>
```
Jozott00, line 38: @AndreasKrall I extended the comment to clarify how more specific register accesses are handled.

AndreasKrall:
```
  /**
   * Returns whether this register tensor represents a single register.
   * This the case if the number of dimensions is 1.
   */
  public boolean isSingleRegister() {
    return dimCount() == 1;
  }
```
I am not sure what a single register is. Eventually this depends also on the base type. If the base type is a format, then it is not only necessary that the dimCount is 1, but it is also necessary, that the number of elements in this dimension is 1.
```
register status1 : StatusFormat
register status2 : StatusFormat<3>
```
The first is for sure a single register. But the second has one dimension and for me it is a tensorRegister or register file.
What happens if the base type is Bool? `register b : Bool` is for sure a single register. Is `register b: Bool<4>` a single register? It probably depends on the use.

Is a register of type Bits<1><64> a single register or a TensorRegister?
Is a register of type Bits<64><1> a single register or a TensorRegister?

In mathematics also a point is a tensor.

What is the usage and application of this function?

Jozott00:
@AndreasKrall when a register's type is a format, from a VIAM perspective the register's type is actually the bits type of the format.

So lets say `StatusFormat` has a type of `Bits<64>` then from a VIAM perspective, there is no difference between
```
register status: StatusFormat
register status: Bits<64>
```
This is also a single register, as it can't be accessed by index (on the VIAM level). 

In the case of your second example (`StatusFormat<3>`), it is not a single register, but a two-dimensional one.
So it can be accessed by an index, so it is the same as currently a register file. 

In the case of `register b: Bool`, it is just a one-dimension register with `Bits<1>`.
The `register b: Bool<4>` is two-dimensional, as on the VIAM side it is `Bits<4><1>`. Of course, if the frontend detects that it is always accessed as a whole, it can turn it into `Bits<4>`, but this would require analysis and is not the default behavior.

In the case of the outermost dimension having only one element (`Bits<1><64>`), it is the frontend's responsibility to turn it into a `register a: Bits<64>`.

The `isSingleRegister()` is just a utility method as an alternative to `tensor.dimCount() == 1`.

AndreasKrall:
> EDIT: One thing we have to discuss is the possibility to access a register across multiple dimensiosn. Let's say we have
> 
> ```
> register A: Bits<4><4><32>
> ```
> 
> Is it possible (on the VIAM level) to access `A` with `A(2)` ( and just `A`), which would return a concatenation of the underlying dimensions?
>

Yes, this is possible. But on the type system level this would return an element at the lower level. I am not quite sure if we should call that concatenation, but perhaps in the ISS and hardware implementation this would be a big bit vector, on the type system the result is again a tensor. Just take another annotation:
```
using Int32   = SInt<32>
using Int32x1 = Int32<4>
using Int32x2 = Int32x1<4>

register A: Int32x2
register B: Int32x2
register C: Int32x1

B := A
C := A(2)
```
The type system makes it clear if you access `A` the result is a bit vector of type `Int32x2` and if you take `A(2)` the result is a bit vector of type Int32x1.

> Currently this is allowed, which means that generators have to take care of such register tensor accesses. This makes handling of registers harder for generators, however, it could lead to better performance (at least in the ISS).
> 
> The alternative would be, that the frontend generates multiple reads, and concatenates them.
> 
> I tend to the first (and currently used) approach, as it provides more flexibility. If a generator does not want to handle such cases, we could write a transformation pass in the VIAM, that turns those accesses into multiple reads and concatenations.

In my opinion only the first implementation is the way to go. It is necessary to handle these accesses as large bit vectors and in the specification of tensor instructions you will need. It is also necessary to support the casting of such bit vectors. This would be a legal cast which could be optimized away:
```
A(1) as Bits<128> as Int32x1
```
We will see a lot of similar casts and `alias register` which should be handled as large bit vectors.

benjaminkasper99:
Adoption in ASM is covered by the LCB refactoring by @kper

kper (approved):

Issue #148, opened 2025-04-10
lcb: Wrong label for `%pcrel_lo` in -fPIC


kper:
When compiling the following program:

~~~
double Coeff = 0.0;

int main() {
        return !(Coeff - Coeff <= 0.0001);
}
~~~

The LCB creates the following assembly snippet:

~~~
        AUIPC a0,%got_pcrel_hi(Coeff)
        LD a0,%pcrel_lo(Coeff)(a0)
~~~

While, upstream generates:

~~~
.Lpcrel_hi0:
        auipc   a0, %got_pcrel_hi(Coeff)
        ld      a0, %pcrel_lo(.Lpcrel_hi0)(a0)
~~~

Upstream creates a temporary label for the lower bits. It's done here https://github.com/llvm/llvm-project/blob/llvmorg-20.1.2/llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp#L3385

kper:
As discussed, this is blocked until the frontend supports it. For now, it is hardcoded.

flofriday:
I somewhat dislike the creation syntax for labels as it requires more custom rules when we could do it without.

### The problem
```
newlabel(hi_label)
```

For syntax breaks many assumptions we have currently. 

First `hi_label` would be treated as any other symbol and since we cannot resolve it (there is no definition/expression that explicitly definied it) the symbol resolver would throw an error here and reject the program. This would require a hack in the symbol resolver to already resolve this function name (even though functions are ususally resolved later in the typechecker) and not check the children symbols fro this function.

Second, every expression has to have a type but defining a type for `hi_label` would require on injecting custom constraints in the bidirectional typechecker (but so far we only have suggestions for types top-down). Also we would rewrite the typechecker probably again writing a hack because now we need to first resolve the buitlin and then the arguments when we ususally first resolve the arguments and then the builtin.

### Solution 1 - A new definition

I think a cleaner implementation is creating a new definition for label creation, just as Andi [initially mentioned](https://github.com/OpenVADL/openvadl/issues/148#issuecomment-2791850472)

```
newlabel hi_label
```

We still have to create a new type `LabelType`. But it would fit right in the existing infrastructure and no more hacks wood be needed.

### Solution 2 - Just strings

```
pseudo instruction LGA_64( rd: Index, symbol: Bits<32> ) = {
      newlabel ( "hi_label" )
      AUIPC { rd = rd, imm = got_hi( symbol ) }
[...]

PR #179, opened 2025-04-18
embench: WIP to support riscv64im

kper:
This PR contains changes to support partially embench for riscv64.

Changes:

- support for clang types in the ABI: the compiler generator also generates a clang frontend. Types like long and pointers have to have a certain width and alignment which is configurable.
- `LlvmUMulLoHiSD` has to be expanded for RISCV64 because we do not match it in a tablegen pattern.
- The `LLA` and `LGA` pseudo instructions require labels during the expansion. This is not supported by frontend and has been hardcoded in a separate pass.
- Label generation was not supported in InstExpander and has been refactored. The InstExpander has now two functions which can be invoked to emit a label or instruction.
- The RISCV spike images use the ISS's built toolchain, so it can reuse existing image layers.
- Hardcoded clang macros for riscv
- Progress on embench riscv64

References #148


vadl/main/vadl/ast/vadl.ATG
```diff
@@ -860,6 +871,43 @@ PRODUCTIONS
...
+  (
+    POINTER (
+      WIDTH SYM_EQ expression<out Expr expr1, BIN_OPS>               (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.POINTER_WIDTH, expr1); .)
+      |  ALIGN SYM_EQ expression<out Expr expr2, BIN_OPS>            (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.POINTER_ALIGN, expr2); .)
+    )
+    | LONG (
+      WIDTH SYM_EQ expression<out Expr expr3, BIN_OPS>               (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.LONG_WIDTH, expr3); .)
+      |  ALIGN SYM_EQ expression<out Expr expr4, BIN_OPS>            (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.LONG_ALIGN, expr4); .)
+    )
+  )
```
kper, line 893: @flofriday I am open to suggestions if you have any to make it more concise haha

vadl/main/vadl/ast/Definition.java
```diff
@@ -3697,6 +3701,108 @@ enum Purpose {
...
 
+/**
+ * The compiler does not only generate a compiler backend but also a clang frontend.
+ * This frontend requires information about the types like: What is the alignment of an integer?
+ * The difference between {@link AbiClangTypeDefinition} and {@link AbiClangNumericTypeDefinition}
+ * is that {@link AbiClangNumericTypeDefinition} requires an integer as property.
+ */
+class AbiClangNumericTypeDefinition extends Definition {
+  AbiClangNumericTypeDefinition.TypeName typeName;
+  Expr size;
```
flofriday, line 3712: **issue:** The expression should be annotated with the `@Child` annotation, otherwise the dumping and symbol resolving won't work here.

vadl/main/vadl/ast/Definition.java
```diff
@@ -3697,6 +3701,108 @@ enum Purpose {
...
+  SyntaxType syntaxType() {
+    return BasicSyntaxType.INVALID;
+  }
+
+  @Override
+  void prettyPrint(int indent, StringBuilder builder) {
+    builder.append(prettyIndentString(indent)).append("clang numeric type: ").append(typeName)
+        .append(" with");
+    size.prettyPrint(indent + 1, builder);
+  }
```
flofriday, line 3750: **chore:** Missing are `equals` and `hashCode`, which is need for tests, (both inspect all values but the location).

vadl/main/vadl/ast/Definition.java
```diff
@@ -3697,6 +3701,108 @@ enum Purpose {
...
+  SyntaxType syntaxType() {
+    return BasicSyntaxType.INVALID;
+  }
+
+  @Override
+  void prettyPrint(int indent, StringBuilder builder) {
+    builder.append(prettyIndentString(indent)).append("clang type: ")
+        .append(typeName).append(" with ").append(typeSize);
+  }
+}
```
flofriday, line 3804: **chore:** Again, `equals` and `hashcode`.

vadl/main/vadl/ast/Definition.java
```diff
@@ -3697,6 +3701,108 @@ enum Purpose {
...
+  enum TypeName {
+    POINTER_WIDTH,
+    POINTER_ALIGN,
+    LONG_WIDTH,
+    LONG_ALIGN
+  }
+
+  public AbiClangNumericTypeDefinition(SourceLocation loc,
+                                       AbiClangNumericTypeDefinition.TypeName typeName,
+                                       Expr size) {
```
flofriday, line 3724: **nitpick:** ~~All~~ Almost all node constructures take the location in last.

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -706,6 +706,18 @@ public Void visit(UsingDefinition definition) {
...
+  @Override
+  public Void visit(AbiClangTypeDefinition abiClangTypeDefinition) {
+    // Check nothing on purpose
+    return null;
+  }
+
+  @Override
+  public Void visit(AbiClangNumericTypeDefinition abiClangNumericTypeDefinition) {
+    // Check nothing on purpose
+    return null;
```
flofriday, line 718: **question:** Why is the `size` field unchecked?

vadl/main/vadl/ast/Ungrouper.java
```diff
@@ -328,6 +328,18 @@ public Void visit(UsingDefinition definition) {
...
+  public Void visit(AbiClangTypeDefinition definition) {
+    ungroupAnnotations(definition);
+    return null;
+  }
+
+  @Override
+  public Void visit(AbiClangNumericTypeDefinition definition) {
+    ungroupAnnotations(definition);
+    return null;
+  }
```
flofriday, line 341: **question:** Probably same reason but why isn't the `size` expression ungrouped?

vadl/main/vadl/ast/ViamLowering.java
```diff
@@ -1466,6 +1476,24 @@ public Optional<vadl.viam.Definition> visit(UsingDefinition definition) {
...
+    return Optional.of(
+        new Abi.AbstractClangType.ClangType(typeName, typeSize, definition.location()));
+  }
+
+  @Override
+  public Optional<vadl.viam.Definition> visit(
+      AbiClangNumericTypeDefinition definition) {
+    var typeName =
+        Abi.AbstractClangType.NumericClangType.TypeName.valueOf(definition.typeName.name());
+    var value = constantEvaluator.eval(definition.size).toViamConstant().intValue();
```
flofriday, line 1492: **chore:** The `constantEvaluator.eval(definition.size)` might produce errors/crashes because the `size` isn't typedchecked

vadl/main/vadl/ast/vadl.ATG
```diff
@@ -860,6 +871,43 @@ PRODUCTIONS
                                                 } .)
   .
 
+  clangTypeDefinition<out Definition def>                     (. def = DUMMY_DEF; .)
+  =
+  (
+    SIZE_T TYPE SYM_EQ typeSizeDefinition<out AbiClangTypeDefinition.TypeSize size1> (. def = new AbiClangTypeDefinition(loc(), AbiClangTypeDefinition.TypeName.SIZE_TYPE, size1); .)
+    | INT MAX TYPE SYM_EQ typeSizeDefinition<out AbiClangTypeDefinition.TypeSize size2> (. def = new AbiClangTypeDefinition(loc(), AbiClangTypeDefinition.TypeName.INT_MAX_TYPE, size2); .)
+  )
+  .
```
flofriday, line 880: **issue:** `loc()` only returns the location of the last consumed token which means in those cases it will only return the token of the last think in the `typedSizeDefinition`. 

What you want though is to also include the first keywords in your location. Many other definitions already have the pattern where `lookaheadLoc()` is used before the first token is consumed and latter is joined using `startLoc.join(loc)`. ;)

vadl/main/vadl/ast/vadl.ATG
```diff
@@ -860,6 +871,43 @@ PRODUCTIONS
...
+      |  ALIGN SYM_EQ expression<out Expr expr2, BIN_OPS>            (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.POINTER_ALIGN, expr2); .)
+    )
+    | LONG (
+      WIDTH SYM_EQ expression<out Expr expr3, BIN_OPS>               (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.LONG_WIDTH, expr3); .)
+      |  ALIGN SYM_EQ expression<out Expr expr4, BIN_OPS>            (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.LONG_ALIGN, expr4); .)
+    )
+  )
+  .
+
+  // See https://clang.llvm.org/doxygen/structclang_1_1TransferrableTargetInfo.html#a30e275c7308708d1b55cd9d009265e33
```
flofriday, line 896: **praise:** I love comments pointing to docs/specs 😍

vadl/main/vadl/ast/vadl.ATG
```diff
@@ -860,6 +871,43 @@ PRODUCTIONS
...
+  clangNumericTypeDefinition<out Definition def>                     (. def = DUMMY_DEF; .)
+  =
+  (
+    POINTER (
+      WIDTH SYM_EQ expression<out Expr expr1, BIN_OPS>               (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.POINTER_WIDTH, expr1); .)
+      |  ALIGN SYM_EQ expression<out Expr expr2, BIN_OPS>            (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.POINTER_ALIGN, expr2); .)
+    )
+    | LONG (
+      WIDTH SYM_EQ expression<out Expr expr3, BIN_OPS>               (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.LONG_WIDTH, expr3); .)
+      |  ALIGN SYM_EQ expression<out Expr expr4, BIN_OPS>            (. def = new AbiClangNumericTypeDefinition(loc(), AbiClangNumericTypeDefinition.TypeName.LONG_ALIGN, expr4); .)
```
flofriday, line 891: **chore:** Again the use of the `loc()`.

**thought:** Maybe we should rename that to something like `lastTokenLoc` to make misuse less common. 🤔

vadl/main/vadl/ast/MacroExpander.java
```diff
@@ -566,6 +566,24 @@ public Definition visit(UsingDefinition definition) {
...
+        abiClangTypeDefinition.typeSize
+    );
+  }
+
+  @Override
+  public Definition visit(AbiClangNumericTypeDefinition abiClangNumericTypeDefinition) {
+    return new AbiClangNumericTypeDefinition(
+        copyLoc(abiClangNumericTypeDefinition.loc),
+        abiClangNumericTypeDefinition.typeName,
+        abiClangNumericTypeDefinition.size
```
flofriday, line 583: **issue:** The size expression needs to be exapanded and cannot be reused.

flofriday (commented):
Overall it looks good, there are just some tricky parts in the frontend that are difficult to get right 😅

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -706,6 +706,25 @@ public Void visit(UsingDefinition definition) {
...
+  @Override
+  public Void visit(AbiClangTypeDefinition abiClangTypeDefinition) {
+    // Check nothing on purpose
+    return null;
+  }
+
+  @Override
+  public Void visit(AbiClangNumericTypeDefinition abiClangNumericTypeDefinition) {
+    check(abiClangNumericTypeDefinition.size);
+    var ty = requireNonNull(abiClangNumericTypeDefinition.size.type);
```
flofriday, line 718: **nitpick:** There is the `type()` getter on many Nodes that implement `Typed Node` that returns the type as a not nullable object.

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -706,6 +706,25 @@ public Void visit(UsingDefinition definition) {
...
+    if (!(ty instanceof ConstantType)) {
+      throw error("Type Mismatch", abiClangNumericTypeDefinition.size)
+          .description("Expected a number as data type")
+          .build();
+    }
```
flofriday, line 723: **thought:** I'm not sure this is what you want.

Even if the number can be known during compile-time it's type isn't necessary `ConstantType`.

For example if you want use the following it wouldn't work:
```
constant Size: Bits<8> = 32
```

What you probably want is to evaluate the expression and see if you can get a number from that (that is known during compilation).

Maybe we should even introduce a `isConstantValue` for the ConstantEvaluator for such checks.

vadl/main/vadl/ast/MacroExpander.java
```diff
@@ -566,6 +566,24 @@ public Definition visit(UsingDefinition definition) {
...
+        abiClangTypeDefinition.typeSize,
+        copyLoc(abiClangTypeDefinition.loc)
+    );
+  }
+
+  @Override
+  public Definition visit(AbiClangNumericTypeDefinition abiClangNumericTypeDefinition) {
+    return new AbiClangNumericTypeDefinition(
+        abiClangNumericTypeDefinition.typeName,
+        expandExpr(abiClangNumericTypeDefinition.size),
```
flofriday, line 582: **chore:** This still isn't quite right as you need to assign the returned expanded Node.

flofriday (changes requested):
Almost done 👍

vadl/main/vadl/ast/MacroExpander.java
kper: Assign to what?

vadl/main/vadl/ast/TypeChecker.java
kper: I can see the benefit but I think this is too much for this PR since my priority lies with the implementation of RISCV64 and the PR is already quite large. I created a new issue for that #200 but using a constant is good enough for me at the moment.

flofriday: ok

vadl/main/vadl/ast/vadl.ATG
flofriday: I don't see anything at the moment and we can still refactor it when we discover something 👍😉

flofriday (approved):

PR #180, opened 2025-04-19
parser: Refactor symbol resolver

flofriday:


vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -647,466 +692,404 @@ public Void visit(ForallThenExpr expr) {
...
+      beforeTravel(definition);
+
+      // Only the attribute needs special treatment here, and if it is set none of the other fields
+      // are set (this is basically an enum).
+      if (definition.attribute != null) {
+        // If attrSymbol is not null, attribute refers to local variable
+        // Else attribute is handled by matching in the AsmParser
+        var attrSymbol = definition.symbolTable().resolveNode(definition.attribute.name);
+        definition.isAttributeLocalVar = attrSymbol instanceof AsmGrammarLocalVarDefinition;
+      } else {
```
benjaminkasper99, line 1029: major: An `AsmGrammarElementDefinition` can have multiple of its children be non-null. So this if-else fails to iterate other non-null children if `attribute` happends to be non-null.
So after dealing with the special case of `attribute`, we still have to iterate the other children. This should also fix the failing test case in `TypecheckerAsmTest`.

vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -647,466 +692,404 @@ public Void visit(ForallThenExpr expr) {
...
+      return null;
+    }
+
+    @Override
+    public Void visit(AsmGrammarLocalVarDefinition definition) {
+      beforeTravel(definition);
+
+      // FIXME: @benjaminkasper99 should we maybe make "null" a symbol that is always in the
+      // symboltable so we can avoid this special treatment here?
+      if (definition.asmLiteral.id != null && !definition.asmLiteral.id.name.equals("null")) {
```
benjaminkasper99, line 1044: thought: Wouldn't this then allow `null` to be used anywhere, where a defined identifier is expected?

vadl/test/vadl/ast/TypecheckerAsmTest.java
```diff
@@ -149,7 +149,7 @@ void jalrInstruction() {
               mnemonic = 'JALR' @operand
               tmp = Register @operand
               [ COMMA rs1 = tmp
-                tmp = Register @operand ]
+                tmp2 = Register @operand ]
```
benjaminkasper99, line 152: This should not work as `tmp2` isn't defined anywhere. But it is never checked because of the issue decribed above.

Jozott00 (dismissed):

benjaminkasper99 (approved):

vadl/main/vadl/ast/SymbolTable.java
Jozott00: question: Where is null used in the specification? Isn’t “none” a better name for something not existing, as it probably does not reference a pointer?

Jozott00 (approved):

vadl/main/vadl/ast/SymbolTable.java
benjaminkasper99: The only point null can be used in the asm grammar is when declaring a local variable, e.g.: `var tmp = null` in this example:
```
JalrInstruction : var tmp = null @operand
   inst = (
   mnemonic = 'JALR' @operand
   tmp = Register @operand
   [ COMMA rs1 = tmp
     tmp = Register @operand ]
   rd = tmp
  ) @instruction
;
```

benjaminkasper99 (approved):

Jozott00 (approved):

PR #186, opened 2025-04-23
docs: Adjust assembly sections

benjaminkasper99:
This PR includes some adjustments to the assembly sections of the tutorial and the assembly related tables in the refmanual.

I also noticed there was still an `AsmType`, which I thought I had already removed - so I  removed it now.


AndreasKrall (approved):
looks good

PR #195, opened 2025-04-28
frontend: Replace `register file` by `register`

Jozott00:
This changes how users define register files.
Previously `register file` was used to specify a register using a mapping type.
As we move into the direction of multidimensional types, also on registers, register files are now also defined using the `register` definition.

#198 is an alternative PR. If we merge #198 we will close this PR without merging it.


flofriday (dismissed):
Looks good to me 👍

kper (approved):

PR #198, opened 2025-04-29
frontend: Replace `register file` by `register` with tensor types

Jozott00:
This PR is an *alternative* PR to #195. 

In comparison to #195, this PR uses a more generic approach by using a new `TensorType` in the frontend that represents both, mapping and multi dimensional types. 

Because of this, this PR additionally adds rudimentary support for multi-dimensional registers:
```
instruction set architecture TEST = {
        
        register A: Bits<5><32>
        register B: Bits<5> -> Bits<5><32>
        register C: Bits<32><5><32>
        
        function t1 -> Bits<32> = A(0)
        function t2 -> Bits<32> = B(A(0) as Bits<5>)(2)
        function t3 -> Bits<32> = B(2, A(0) as Bits<3>)
        function t4 -> Bits<32> = C(B(A(0) as Bits<5>, 4) as Bits<5>, 2)        
        function t5 -> Bits<1> = C(B(A(0) as Bits<5>, 4) as Bits<5>, 2)(2)
}
```

This `TensorType` could be used as a frontend super type for all `DataType` related operations for a more complete multi dimensional type support.

### DIFF
--- a/vadl/main/vadl/ast/AstUtils.java
+++ b/vadl/main/vadl/ast/AstUtils.java
@@ -30,6 +30,7 @@ import vadl.types.Type;
 import vadl.types.UIntType;
 
 class AstUtils {
+
   @Nullable
   static BuiltInTable.BuiltIn getBuiltIn(String name, List<Type> argTypes) {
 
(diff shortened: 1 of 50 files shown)


Jozott00:
Closed in temporary favor of #195. However, when tensor types are generally implemented, the approach of this PR will probably be adapted.

PR #204, opened 2025-05-01
lcb: Print pseudo instructions

kper:
Before this PR, all pseudo instructions have been expanded before printing. This is not required anymore. However, constant sequence and register adjustment sequences do not have an assembly printing function. Therefore, we still need to expand those.


vadl/main/vadl/viam/graph/HasRegisterTensor.java
```diff
@@ -21,11 +21,11 @@
 /**
  * Interface to indicate that the implementing class has register file.
  */
-public interface HasRegisterFile {
+public interface HasRegisterTensor {
```
Jozott00, line 24: I think this marker interface got pretty useless, as there are only the `ReadRegTensor` and `WriteRegTensor` nodes.

vadl/main/vadl/viam/graph/dependency/FuncParamNode.java
```diff
@@ -76,4 +77,22 @@ public Definition definition() {
...
+    if (obj == null) {
+      return false;
+    }
+
+    if (obj instanceof FuncParamNode that) {
+      return Objects.equals(parameter, that.parameter);
+    } else {
+      return false;
+    }
+  }
```
Jozott00, line 97: major: why is this required? I think we shouldn't start implementing `equals` and `hashcode` for nodes. There are generic methods that already do that for any kind of nodes.

vadl/main/vadl/viam/passes/functionInliner/FunctionInlinerPass.java
```diff
@@ -103,7 +103,7 @@ private UninlinedGraph handleMainBehavior(Instruction instruction) {
               Arrays.stream(functionCall.function().parameters()), Pair::new)
           .forEach(
               pair -> behaviorCopy.getNodes(FuncParamNode.class)
-                  .filter(n -> n.parameter() == pair.right())
+                  .filter(n -> n.parameter().equals(pair.right()))
```
Jozott00, line 106: If the parameter is a different object than what is used in the FuncParamNode, then there is probably a bug during lowering. Equals shouldn’t be necessary here, as every definition only exists ones.

vadl/main/vadl/viam/Parameter.java
```diff
@@ -83,4 +84,23 @@ public void setParent(@SuppressWarnings("NullableProblems") Definition parent) {
...
+      return false;
+    }
+
+    if (obj instanceof Parameter that) {
+      return Objects.equals(identifier, that.identifier)
+          && Objects.equals(type, that.type);
+    } else {
+      return false;
+    }
+  }
```
Jozott00, line 105: major: This shouldn’t be necessary. A definition only exists ones in the whole VIAM, so there are never duplicated objects of the same parameter.

Jozott00 (changes requested):

vadl/main/vadl/viam/graph/HasRegisterTensor.java
kper: And yeah, that's the problem
Two classes but I would like to have one check.

vadl/main/vadl/viam/graph/dependency/FuncParamNode.java
kper: It's required because they aren't the same across behavior and assembly printing function. We've got two instances of the FuncParamNode for example.

Jozott00 (approved):

PR #205, opened 2025-05-01
grammar: Rename `micro processor` to `processor`

Jozott00:
With this PR users will use `processor` instead of `micro processor` in their specifications.
This breaks previous specifications.


flofriday (approved):

flofriday:
Looks good to me 🎉

PR #212, opened 2025-05-03
iss: Add `memory region` to `processor` (discussion #138)

Jozott00:
This PR implements the `memory region` definition proposed in #138.


vadl/main/vadl/ast/SymbolTable.java
Jozott00, line 1: @flofriday I refactored the existing `resolve` methods in the `SymbolTable` to take an `Identifier, IdentifierPath and IsId` instead of a String. This makes it more generic and easier to use. However, the most important advantage is that now, the resolved node is automatically set as target to the `Identifier(Path)`, which makes it much more reliable. 

Now it isn't necessary anymore to have a nullable field for a resolved target for some definition, but instead one can just add a method that uses the identifier's target node and cast it down to whatever they want. (see `CpuMemoryRegionDefinition` that has a reference to a `MemoryDefinition` without having a separate property that is set during symbol resolving).

vadl/main/vadl/ast/Definition.java
```diff
@@ -3959,6 +3961,12 @@ Stream<CpuFunctionDefinition> findCpuFuncDef(CpuFunctionDefinition.BehaviorKind
         .map(e -> (CpuFunctionDefinition) e);
   }
 
+  Stream<CpuMemoryRegionDefinition> findMemoryRegionDefs() {
+    return definitions.stream()
+        .filter(e -> e instanceof CpuMemoryRegionDefinition region)
+        .map(e -> (CpuMemoryRegionDefinition) e);
+  }
+
```
flofriday, line 3969: **question:** Wouldn't it be more useful to have a generic `find` method like:

```java
<T extends Definition> Stream<T> findDefsOf(Class<T> klass) {
   // implementation left as an exercise to the reader :P
}
```

vadl/main/vadl/ast/Definition.java
```diff
@@ -4139,6 +4147,109 @@ public int hashCode() {
...
+ * [ base: 0x8000000 ]
+ * memory region [RAM] DRAM in MEM
+ *
+ * memory region [ROM] MROM in MEM = {
+ *   MEM<4>(0x1000 as Bits<64>) := 0x00000297  // auipc t0, 0x0
+ *   MEM<4>(0x1004 as Bits<64>) := 0x02828613  // addi a2, t0, 40
+ * }
+ * }</pre>
+ */
+class CpuMemoryRegionDefinition extends Definition implements IdentifiableNode {
```
flofriday, line 4163: **praise:** Amazing documentation ✨

vadl/main/vadl/ast/MacroExpander.java
```diff
@@ -801,6 +801,13 @@ public Definition visit(CpuFunctionDefinition definition) {
     ).withAnnotations(expandAnnotations(definition.annotations));
   }
 
+  @Override
+  public Definition visit(CpuMemoryRegionDefinition definition) {
+    return new CpuMemoryRegionDefinition(definition.id, definition.kind, definition.memoryRef,
```
flofriday, line 806: **chore:** I think you also need to expand the `id` and `memoryRef` (because it's location needs to be adapted).

vadl/main/vadl/ast/Definition.java
```diff
@@ -4139,6 +4147,109 @@ public int hashCode() {
...
+
+    String keyword() {
+      return name();
+    }
+  }
+
+  Identifier id;
+  MemKind kind;
+  @Child
+  Identifier memoryRef;
```
flofriday, line 4176: **nitpick:** The way `id` and `memoryRef` are written they can not be expanded from an macro, which might be fine, depending on the usecase you are expecting.

vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -744,7 +742,8 @@ public Void visit(AnnotationDefinition definition) {
 
       if (definition.annotation == null) {
         // FIXME: Add a more descriptive error message and a body.
-        requireNonNull(definition.symbolTable).errors.add(
+        // FIXME: Add a list of available annotations (usage strings) as help to the error.
+        definition.symbolTable().errors.add(
```
flofriday, line 746: **question:** What's missing to implement this?

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -165,6 +165,9 @@ private void verifyAnnotations(Definition def) {
             .build();
       }
 
+      // check annotation definition itself
+      check(annotation);
```
flofriday, line 169: **chore:** Can we move this into the `visit(Annotation definition)` method?

vadl/main/vadl/ast/SymbolTable.java
flofriday: Sounds good, historically Identifiers and Paths didn't have a target so the API wasn't designed around them, but I'm also quite unhappy with the API of the Symboltable and a propper redesign is in place.

But your contribution definitely helped here.

vadl/main/vadl/ast/TypeChecker.java
Jozott00: I think this is actually the call to the visit. Without this, the visit isn't invoked at all.

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -486,15 +521,105 @@ public void applyViam(vadl.viam.Definition definition, List<Annotation> annotati
...
-    private static class GroupCheckContext<D> {
-      final D targetDefinition;
-      final List<Annotation> annotations;
+    private static class GroupContext<D> {
+      final D astTargetDef;
+      // annotations set by the user
+      final LinkedHashMap<String, Annotation> annotations;
+      // holds the annotation factories of this group.
+      // might be used to get declared annotations.
+      private final Map<String, Supplier<Annotation>> factories;
```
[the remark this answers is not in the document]
Jozott00: I think as it is a private field, it should not have to make sense to users (as they don't "see" it). I can also abstract it by passing a `declarationProvider` to the group, that is just a `Function<String, AnnotationDeclaration>`.

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -659,6 +840,11 @@ void typeCheck(AnnotationDefinition definition, TypeChecker typeChecker) {
       isEnabled = typeChecker.constantEvaluator.eval(valueExpr).value().equals(BigInteger.ONE);
     }
   }
+
+  @Override
+  public String usageString() {
+    return "[ " + name + " ]";
+  }
```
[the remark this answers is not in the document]
Jozott00: This is part of the `EnableAnnotation`, so it can't be abstract. The `Annotation` superclass does not implement this method.

vadl/main/vadl/ast/AnnotationTable.java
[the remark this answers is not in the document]
Jozott00:
```
error: Missing annotation
     ╭──[file:///Users/jozott/development/uni/dipl/iss-dev/code/open-vadl/sys/risc-v/rv64im.vadl:136:3]
     │
 136 │   [ firmware ]
     │   ^^^^^^^^^^^^ Requires the [ base : <expr> ] annotation
     │ 
     If [ firmware ] was specified, the definition also requires [ base : <expr> ].
```

```
error: Missing annotation
     ╭──[file:///Users/jozott/development/uni/dipl/iss-dev/code/open-vadl/sys/risc-v/rv64im.vadl:136:3]
     │
 136 │   [ firmware ]
     │   ^^^^^^^^^^^^ Requires the `base` annotation
     │ 
     If [ firmware ] was specified, the definition also requires `base` annotation.
```

I think the most useful cases are annotations like `EnumAnnotation` (`[ name : opt1 | opt2 | opt3 ]`) and `ZeroConstraintAnnotation`.

However, it's fine for me to remove it.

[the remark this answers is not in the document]
Jozott00: One thing I noticed: We could make checks safer with such annotations. E.g. consider 
```
ctx.verifyIFThen("anno xy", "anno abc")
```
It can easily happen that someone make a typo when checking annotation names. However, the check itself could verify that the given name is part of the declared annotations (even if the annotation is not set by the user). This way, we can be sure that the annotation checks are actually tiggered.

The same can be done for things like `ctx.get("anno xy", EnableAnnotation.class)`, which could always throw an `IllegalStateException` if the annotation is of different type, even if the annotation is not set by the user. This would nearly prevent such bugs (with subsequent crashes) in the distributed version.

vadl/main/vadl/ast/Definition.java
Jozott00: I think the reason for the `memoryRef` was, that the `Identifier` itself has a target node. 
Do you think it would be fine to add a `@Nullable Node target()` to the `IsId` interface? This way I could just use `IsId`, which would allow macro extension right?

And if I understood correctly, after macro expansion, the `IsId` is always either an `Identifier` or an `IdentifierPath`, is that right?

vadl/main/vadl/ast/MacroExpander.java
Jozott00: Alright, I did it the same way as other definitions in the expander (that also don't expand their id and references). But if extending these two properties is generally better, I will update the implementation :)

kper (approved):

PR #218, opened 2025-05-05
ci: Added image QEMU riscv64

kper:
For convince I created a qemu image which already has riscv64 prebuild.


Jozott00 (approved):
Note: I used that before, however the build process got stuck when I used as the testing base image. (I think because it got confused with the redis cache, but I am not sure why exactly). 

So you will probably not be able to use this in your tests.

PR #221, opened 2025-05-05
asm: Emit correct kind of relocations for unknown symbols

benjaminkasper99:
This PR adjusts the the `MCCodeEmitter` to emit the correct type of relocations when unknown symbols are used as immediate operands. If the instruction's behavior adds the immediate operand to the program counter, a relative relocation is emitted, else an absolute relocation is emitted.

### DIFF
--- a/vadl/main/vadl/gcb/passes/RelocationKindCtx.java
+++ b/vadl/main/vadl/gcb/passes/RelocationKindCtx.java
@@ -1,0 +1,46 @@
+// SPDX-FileCopyrightText : © 2025 TU Wien <vadl@tuwien.ac.at>
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package vadl.gcb.passes;
+
+import java.util.Map;
+import vadl.gcb.passes.relocation.model.CompilerRelocation;
+import vadl.viam.Definition;
+import vadl.viam.DefinitionExtension;
+import vadl.viam.Format;
+import vadl.viam.Instruction;
+
+/**
+ * An extension for {@link Instruction} to store which kind of relocations
+ * should be emitted for its immediate operands.
+ */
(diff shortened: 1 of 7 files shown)


vadl/main/vadl/lcb/template/lib/Target/MCTargetDesc/EmitMCCodeEmitterCppFilePass.java
```diff
@@ -209,4 +225,44 @@ private List<Map<String, Object>> generateSymbolRefFixupMappings(
...
         }).toList();
   }
+
+  /**
+   * Determines the relocation kind for a given instruction and immediate operand.
+   * If the instruction's behavior adds the immediate to the Program Counter (PC),
+   * then the relocation kind is RELATIVE.
+   */
+  private CompilerRelocation.Kind fittingRelocationKindByBehavior(
+      Instruction instruction, Format.FieldAccess fieldAccess) {
```
kper, line 235: question: I do not understand, why do you need the relative relocations here?

kper (changes requested):

vadl/main/vadl/lcb/template/lib/Target/MCTargetDesc/EmitMCCodeEmitterCppFilePass.java
benjaminkasper99: An example would be branch or jump instructions in RISCV like `BEQ x1, x2, .lbl` or `JAL x1, .lbl`.
Here is the upstream LLVM implementation: https://github.com/llvm/llvm-project/blob/f49ee00ec47f596269639987f5ea8755a358ca16/lld/ELF/Arch/RISCV.cpp#L288

kper: ok, so any instruction which references the PC requires to emit relative relocations.
But I don't think this belongs into this class. Can you implement this in an extra pass and use the result or attach an extension to the instruction how the (instruction, field_access) maps to a kind.

vadl/main/vadl/gcb/passes/RelocationKindCtx.java
```diff
@@ -0,0 +1,46 @@
...
+import vadl.viam.Format;
+import vadl.viam.Instruction;
+
+/**
+ * An extension for {@link Instruction} to store which kind of relocations
+ * should be emitted for its immediate operands.
+ */
+public class RelocationKindCtx extends DefinitionExtension<Instruction> {
+
+  private final Map<Format.Field, CompilerRelocation.Kind> immediateRelocationKindMap;
```
kper, line 32: nitpick: You have here a "leaking" abstraction. Of course, you can only relocate immediates but it's not clear from the type since it only depends on what you have put into the map. Therefore, I would propose to rename the property to `fieldToKind`.

kper (changes requested):

kper (approved):

Jozott00 (approved):

PR #223, opened 2025-05-05
iss: Support `rv32csr.vadl` specification

Jozott00:
This PR adds support for the `rv32csr.vadl` specification.

The most important change that was necessary was the implementation of the null register annotation.

### DIFF
--- a/vadl/main/vadl/ast/AnnotationTable.java
+++ b/vadl/main/vadl/ast/AnnotationTable.java
@@ -40,7 +40,9 @@ import vadl.utils.SourceLocation;
 import vadl.utils.WithLocation;
 import vadl.utils.functionInterfaces.TriConsumer;
 import vadl.viam.AssemblyDescription;
+import vadl.viam.Constant;
 import vadl.viam.MemoryRegion;
+import vadl.viam.RegisterTensor;
 import vadl.viam.Relocation;
 import vadl.viam.annotations.AsmParserCaseSensitive;
 import vadl.viam.annotations.AsmParserCommentString;
@@ -73,9 +75,13 @@ class AnnotationTable {
         // FIXME: Apply to AST
         .build();
 
-    annotationOn(RegisterDefinition.class, "zero", ExprAnnotation::new)
-        // FIXME: Typecheck
-        // FIXME: Apply to VIAM
+    annotationOn(RegisterDefinition.class, "zero", ZeroConstraintAnnotation::new)
+        .applyViam((def, annotation, lowering) -> {
+          var viamDef = (RegisterTensor) def;
+          var indices = annotation.indices.stream().map(ConstantValue::toViamConstant).toList();
+          var zero = Constant.Value.of(0, viamDef.resultType(indices.size()));
+          viamDef.addConstraint(new RegisterTensor.Constraint(indices, zero));
+        })
         .build();
 
     groupOn(RelocationDefinition.class)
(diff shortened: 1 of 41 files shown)


vadl/main/vadl/ast/Expr.java
```diff
@@ -1639,6 +1639,10 @@ void replaceArgsFor(int index, List<Expr> newArgs) {
     args.values.addAll(newArgs);
   }
 
+  List<Expr> argIndicesFlatten() {
+    return argsIndices.stream().flatMap(a -> a.values.stream()).toList();
+  }
```
flofriday, line 1644: **thought:** Unfortunately this is most of the times semantically wrong to use, and implementing it will make prone to being used, since how you write calls does make a semantically difference.

```
// Call the function with three arguments and slice the result
abc(1,2,3)(8..16)
// Typeerror since 4 arguments are provided for a function that only takes 3
abc(1,2,3, 8..16)


// A slice followed by a bit index results in a Bits<1>
x(16..32)(8)

// A slice and a bit index concatinated, resulting in a Bits<17>
x(16..32, 8)
```

I think it's better not to have this function at all since its simply not that useful.

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -1045,4 +1051,63 @@ void typeCheck(AnnotationDefinition definition, TypeChecker typeChecker) {
   public String usageString() {
     return "[ " + name + " : <expr> ]";
   }
+
 }
+
+/**
+ * The {@code [ zero : <register>(<expr>) ]} annotation.
+ * This is its own class, as the typechecking is rather complex and determines
+ * new properties.
```
flofriday, line 1060: **nitpick:** We should probably add some examples in to the docs.

vadl/main/vadl/ast/BehaviorLowering.java
```diff
@@ -335,6 +335,8 @@ private ExpressionNode fetch(Expr expr) {
     var result = expr.accept(this);
     result.setSourceLocationIfNotSet(expr.location());
     expressionCache.put(expr, result);
+    result.ensure(!(result.type() instanceof ConstantType),
+        "Constant types must not exist in the VIAM");
```
flofriday, line 339: **thought:** There are actually more types that cannot exist in the VIAM. I once wrote the `toViamType` (if I remember correctly) method that should cast them correctly or throw errors if no propper cast exists.

**question:** What does ensure do? Does it throw an error or crash because I think we should crash here since there is nothing in the user can do if that happens and it's a bug in the compiler if a constant type ever escapes to the VIAM.

vadl/main/vadl/ast/BehaviorLowering.java
```diff
@@ -535,8 +537,8 @@ public ExpressionNode visit(GroupedExpr expr) {
 
   @Override
   public ExpressionNode visit(IntegerLiteral expr) {
-    throw new RuntimeException(
-        "The behavior generator doesn't implement yet: " + expr.getClass().getSimpleName());
+    return new ConstantNode(Constant.Value.fromInteger(expr.number,
+        ((ConstantType) expr.type()).closestBits()));
```
flofriday, line 541: **question:** Did your ever trigger that code?

Because if I remember correctly there is an optimization that converts all constant types to Viam constants and since the type of an `IntegerLiteral` should always be constant there shouldn't be a path that triggers this code at all.

Though I'm sure a comment explaining this would have made that clearer.

vadl/main/vadl/ast/ViamLowering.java
```diff
@@ -781,8 +781,7 @@ public Optional<vadl.viam.Definition> visit(CounterDefinition definition) {
 
     // FIXME: Further research for the parameters (probably don't apply to counter)
     var reg = new RegisterTensor(identifier,
-        List.of(dimFromType(0, resultType)),
-        new RegisterTensor.Constraint[] {}
+        List.of(dimFromType(0, resultType))
```
flofriday, line 784: **chore:** You should wrap the type in `toViamType` to avoid leaking frontend only types.

vadl/main/vadl/ast/ViamLowering.java
```diff
@@ -1373,8 +1372,7 @@ public Optional<vadl.viam.Definition> visit(RegisterDefinition definition) {
     var resultType = (DataType) getViamType(definition.type());
     var reg = new RegisterTensor(
         generateIdentifier(definition.viamId, definition.identifier()),
-        List.of(dimFromType(0, resultType)),
-        new RegisterTensor.Constraint[] {}
+        List.of(dimFromType(0, resultType))
```
flofriday, line 1375: **chore:** You should wrap the type in `toViamType` to avoid leaking frontend only types.

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -1045,4 +1051,63 @@ void typeCheck(AnnotationDefinition definition, TypeChecker typeChecker) {
...
+ */
+class ZeroConstraintAnnotation extends ExprAnnotation {
+  @LazyInit
+  List<ConstantValue> indices;
```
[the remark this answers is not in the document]
Jozott00: Yes, especially with multi dimensional registers.

vadl/main/vadl/ast/BehaviorLowering.java
Jozott00: Yeah, we should extend this to every type that cannot be part of the VIAM. 
`ensure` checks the condition and crashes with an VIAM error and context information. This is similar to an `assert`.

vadl/main/vadl/ast/ViamLowering.java
Jozott00: The `resultType` is already retrieved from a `getViamType` call.

vadl/main/vadl/ast/ViamLowering.java
Jozott00: The `resultType` is already retrieved from a `getViamType` call.

flofriday (approved):

vadl/main/vadl/ast/BehaviorLowering.java
flofriday: True, I will adress the issue in another PR :)

PR #234, opened 2025-05-07
sys:var-risc Fix opcode encoding of SYS insn

rascmatt:
The instruction set has a conflict for the `BAL` and `SYS` instruction (`opcode: 62`, `func5: 31`) which currently have the same instruction patterns

This looks to me like a mistake in the macro definition of `SysCallInstr`, which should use `opcode 61` (`MemCode`) according to the op/fun-code table in the comment.

I fixed it, but please check if that's the correct replacement opcode.


AndreasKrall (approved):
Thank you for fixing this

PR #237, opened 2025-05-08
decoder: Fix issues with split bit splices encoding

rascmatt:
There was a problem when converting encodings of fixed fields into the corresponding bit pattern for the decoder generation algorithm.

For example this happened for instructions of the `AddSubExtFormat` format in the `AArch64` spec, where the `opcode` field is split across multiple bit slices.


Jozott00 (approved):
Many thanks for the quick fix!

PR #238, opened 2025-05-08
frontend: Support multi-inheritance on ISA definitions

Jozott00:
This implements the proposal discussed in #40.
It enables a very modular design of the ISA extension with a high degree of flexibility.

In addition to the implementation of multiple inheritance, this PR fixes some problems in the parser:
- ISA identifiers and extensions can be macro expansions
- Fixes a bug that an ISA can only use the macro definition in its direct super ISA, but not e.g. in its super super ISA (the macro symbol scope is now recursively extended)
- Fixes a bug where conflicting macro definitions did not cause an error. This was because the error list in the `macroTable` was never checked and the errors from child `macroTables` were not passed on to the parent table. (this is now handled by the new method `SymbolTable#pop()`)
- Fix `SymbolTable#extendBy` method that blindly added all symbols to the given table even if there were conflicts. With the implementation, there can be no conflicting symbols coming from an extended scope (e.g. `with ABI` or `extending ISA` or `implementing ISA`).
- Fix `SymbolTable#defineSymbol` that caused a `symbol already defined` error in diamond patterns (e.g. `ISA - ABI - SuperISA`). If the symbol is already defined, but the source node is the same object, no error is now registered.


vadl/test/resources/testSource/sys/risc-v/rv3264im.vadl
Jozott00, line 1: @AndreasKrall is the RISC-V specification refactoring fine?

AndreasKrall: Yes, but now with multiple inheritance I would rename `rv32csr` just to `rvcsr`. The `rvcsr` works both for the 32 bit and 64 bit implementation, it just depends on how `Regs` is defined. Then it is either a 32 bit or 64 bit register.

vadl/main/vadl/ast/Expr.java
```diff
@@ -757,6 +757,18 @@ public int hashCode() {
...
+ * A placeholder is some macro expr that cannot be directly expanded to an {@link Identifier}, as
+ * itself is part of a macro definition.
+ *
+ * <p>For instance in the following {@code $constId} is such a placeholder:
+ * <pre>{@code
+ * model Test (constId: Id) : Defs = {
+ *   constant $constId = 4
+ * }
+ * }</pre></p>
+ */
```
flofriday, line 771: **praise:** This looks amazing, thank you for improving the documentation here.

Just a followup since you added docs here, is there really a need or difference between this and `IsId`?
Because I kinda lost track on why both are useful. Also I hope we can come up with a better name than `IsId`.

vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -39,7 +39,7 @@ class SymbolTable {
   final Map<String, AstSymbol> macroSymbols = new HashMap<>();
   List<Diagnostic> errors = new ArrayList<>();
 
-  interface Symbol {
+  sealed interface Symbol {
```
flofriday, line 42: **question:** What's the benefit of making it a `sealed interface` ?

vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -109,6 +109,21 @@ SymbolTable createChild() {
...
+   * @return the parent symbol table
+   * @throws IllegalStateException if parent is null
+   */
+  SymbolTable pop() {
+    if (parent == null) {
+      throw new IllegalStateException("Tried to pop symbol table, but parent is null");
+    }
+    parent.errors.addAll(this.errors);
+    return parent;
+  }
```
flofriday, line 125: **question:** Maybe I'm missremembering this but I thought that the `errors` field of all `Symboltable` instances pointed to the same list.

So only one list of `errors` ever existed.

To be fare I don't love that design but I haven't come up with a much better approach yet.

vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -247,6 +262,15 @@ Macro getMacro(String name) {
...
+   * This will add an error to {@link #errors} if the identifier couldn't be resolved
+   * to a node of the given type.
+   * In this case it will return null, so the user must check the result before continuing.
+   *
+   * @param usage the identifier that should be resolved
+   * @param type  the type that the resolved node must have
+   * @return the resolved node, or null if it could not be resolved with the given type
+   */
   // FIXME: I don't like how it's called require but still returns null
   <T extends Node> @Nullable T requireAs(IsId usage, Class<T> type) {
```
flofriday, line 275: **question:** Now that you have had more contact with the symbol table, do you think it makes sense to keep this method or would it be better to only rely on `findAs`?

Also should that method be private, because if called after the symbolresolver the errors will get added to the symboltable but no other code checks if there are new errors in the symboltable, right?

vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -301,9 +325,28 @@ SyntaxType requireSyntaxType(Identifier recordName) {
...
 
+  /**
+   * Copies all symbols of the given symbol table into this symbol table.
+   * It internally calls {@link #defineSymbol(String, Node)}, so it
+   * will register an error in {@link #errors} if there are symbol name conflicts.
+   */
   void extendBy(SymbolTable other) {
-    symbols.putAll(other.symbols);
-    macroSymbols.putAll(other.macroSymbols);
+    // we have to check for each symbol that is is not already in this symbol table
```
flofriday, line 334: **chore:** typo in "that _it_ is"

vadl/main/vadl/ast/ViamLowering.java
```diff
@@ -1239,21 +1240,17 @@ public Optional<vadl.viam.Definition> visit(ProcessorDefinition definition) {
...
-      mergeInto(definition, nodeList, processedIsas);
-    }
-
-    var identifier = findIsaIdentifier(definitions);
-    var location = findIsaLocation(definitions);
+    mergeInto(isaDef, nodeList, processedIsas);
 
     // create new isa ast node
-    return new InstructionSetDefinition(identifier, null, nodeList, location);
+    return new InstructionSetDefinition(isaDef.identifier, List.of(), nodeList,
```
flofriday, line 1252: **question:** What is the empty list here, will that be filled at some point in the future?

vadl/main/vadl/ast/MacroExpander.java
```diff
@@ -776,7 +776,7 @@ definition.purpose, definition.exprs, copyLoc(definition.loc)
   @Override
   public Definition visit(ProcessorDefinition definition) {
     var definitions = expandDefinitions(definition.definitions);
-    return new ProcessorDefinition(definition.id, definition.implementedIsas, definition.abi,
+    return new ProcessorDefinition(definition.id, definition.implementedIsa, definition.abi,
```
flofriday, line 779: **question:** Do you need to expand `definition.id`, `definition.implementedIsa`and `definition.abi`?

vadl/main/vadl/ast/MacroExpander.java
flofriday, line 1: **thought:** Since the Instructionsetdefinition changed to allow placeholders, don't you need to adapt some code to expand them properly?

vadl/main/vadl/ast/Expr.java
Jozott00: `IsId` can also be an `IdentifierPath`, while `IdentifierOrPlaceholder` can only expand to a single identifier. Essentially `IsId` is a unknown AST node that eventually expands to either `Identifier` or `IdentifierPath`.

But yeah, I would also suggest thinking of a better name and maybe a common interface for `Identifier` and `IdentifierPath`.

vadl/main/vadl/ast/SymbolTable.java
Jozott00: Type save/exhaustive switch over all subtypes of `Symbol`. (Like an enum in swift and rust)

flofriday: True, but doesn't that also need a `permits` clause or is that optional? 🤔

vadl/main/vadl/ast/SymbolTable.java
Jozott00: Oh yeah, you are right, this is not needed. I will remove the error propagation, leave the method, because it is a bit more readable in the ATG.

vadl/main/vadl/ast/SymbolTable.java
Jozott00: I think it does make sense, otherwise every caller would have to add an error manually. Making it private is a good idea, however, the Parser needs the method to find ISA definitions (this is a special case, as macros from super ISAs are available in a sub-ISA). Maybe for the parser we should add a new method that makes clear that this is only used by the parser. 

All other usages of this method can be replaced by a call to the `target()` of the identifier.

vadl/main/vadl/ast/MacroExpander.java
Jozott00: Those are only identifiers which where already expanded during parsing (afaik).

vadl/main/vadl/ast/SymbolTable.java
Jozott00: I think, if it is a inner class (or a class in the same file) it is optional. 
(otherwise it won't compile ^^)

vadl/main/vadl/ast/SymbolTable.java
flofriday: Oh I guess, and if the parser doesn't find it it also should throw an error...
Well then we might keep it but still unhappy that after the symbolresolver the method is still visible but shouldn't be used.

Jozott00: I now refactor the symbol table, so almost all methods are private, which should hopefully prevent misuse.

flofriday (approved):

PR #248, opened 2025-05-14
iss: Fix Risc-V test runner

rascmatt:
The qemu simulation attempts to decode multiple invalid instructions, as it runs off into invalid memory:

Test to reproduce:

```yaml
tests:
- id: BEQ
  debug: true
  asm_core: |-  
    li t0, 123
    add x3, x0, t0
```

The test results report illegal instructions:

```yaml
simLogs:
  stdout: []
  stderr:
  - '[EState] use tohost addr: 0x0'
  - 'qemu-system-rv64im: RV64IM translate, illegal instr, pc: 0x80000018 , insn: 0x0000'
  - ''
  - 'qemu-system-rv64im: RV64IM translate, illegal instr, pc: 0x80000018 , insn: 0x0000'
  - ''
  - '[REG] x0: 0x0000000000000000'
  - '[REG] x1: 0x0000000000000001'
  - '[REG] x2: 0x0000000080001000'
 [...]
```

@Jozott00 Pls check if that makes sense or if there's another reason for this.

### DIFF
--- a/vadl/test/resources/scripts/iss_qemu/compilers/riscv_compiler.py
+++ b/vadl/test/resources/scripts/iss_qemu/compilers/riscv_compiler.py
@@ -57,6 +57,8 @@ async def build_assembly(id: str, core: str) -> Path:
   la x2, tohost
   sd x1, 0(x2)
   
+  1: j 1b # Loop to avoid running off into invalid memory
+  
   .section .tohost, "aw", @progbits
   .align 6; 
   .global tohost; 


vadl/test/resources/scripts/iss_qemu/compilers/riscv_compiler.py
```diff
@@ -57,6 +57,8 @@ async def build_assembly(id: str, core: str) -> Path:
   la x2, tohost
   sd x1, 0(x2)
   
+  1: j 1b # Loop to avoid running off into invalid memory
```
Jozott00, line 60: When Thomas started with the CAS, we removed the loop because it makes the instruction counting non-deterministic. However, when Thomas left, I removed support for instruction counting and the corresponding tests.

In general, it doesn't matter if illegal instructions are found at the end of the test, as long as the test terminates before the timeout (1 second). However, since it is no longer necessary to determine the number of instructions executed, it is fine to put this back in as it is less confusing.

Jozott00 (approved):

vadl/test/resources/scripts/iss_qemu/compilers/riscv_compiler.py
rascmatt: The original issue which prompted this PR was that the ISS ran into an assertion failure due to not correctly incrementing ctx->base.pc_next upon raising an illegal-instruction error.

This will be resolved (as discussed) by incrementing pc_next by 1 byte in case of an illegal instruction to avoid this assertion. (As pc_next should not be used any more due to the translation block being terminated).

So I think the reason to make the instruction count deterministic is reasonable, even if the test output may be confusing at times. So I'd suggest to leave it as is and abandon this change.

PR #250, opened 2025-05-15
lcb: Run the optimizer on derived graphs and enhance dumps

kper:
It is sometimes necessary to create multiple graphs based on the original viam graph. This PR adds the functionality s.t these graphs get also optimised. Additionally, they will be dumped so that a developer can see why a LLVM graph was not lowered properly.

For example, the ADDI has a VIAM behavior, a lowered LCB graph and an additional graph where the immediate is replaced by a frame index.
Screenshot 2025-05-15 at 12 52 32

### DIFF
--- a/vadl/main/vadl/dump/BehaviorTimelineDisplay.java
+++ b/vadl/main/vadl/dump/BehaviorTimelineDisplay.java
@@ -1,0 +1,27 @@
+// SPDX-FileCopyrightText : © 2025 TU Wien <vadl@tuwien.ac.at>
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package vadl.dump;
+
+import vadl.viam.graph.Graph;
+
+/**
+ * Allows the display of the {@link Graph} in a timeline.
+ */
+public interface BehaviorTimelineDisplay {
+  String passId();
+  String passName();
+}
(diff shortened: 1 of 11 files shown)


vadl/main/vadl/dump/InfoUtils.java
```diff
@@ -90,7 +89,7 @@ public static Info.Modal createGraphModal(String title,
   @SuppressWarnings("LineLength")
   public static Info.Modal createGraphModalWithTimeline(String title,
                                                         String modalTitle,
-                                                        List<Pair<PassResults.SingleResult, String>> passGraphs) {
+                                                        List<Pair<BehaviorTimelineDisplay, String>> passGraphs) {
```
Jozott00, line 92: Assuming we implement the suggested change, this would become a `List<BehaviorTimelineDisplay>`.

vadl/main/vadl/dump/infoEnrichers/LcbEnricherCollection.java
```diff
@@ -98,6 +100,78 @@ public class LcbEnricherCollection {
...
+                @Override
+                public String passId() {
+                  return "BehaviorRewrittenPassId";
+                }
+
+                @Override
+                public String passName() {
+                  return "BehaviorRewritten";
+                }
+              }
```
Jozott00, line 151: This would be avoided with the suggested approach.

vadl/main/vadl/dump/BehaviorTimelineDisplay.java
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Get the pass key id.
+   */
+  String passId();
+
+  /**
+   * Get the human-readable string from the timeline.
+   */
+  String passName();
+}
```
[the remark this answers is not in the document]
kper: I see your point! But I think, the timeline is a feature of the pass. Imho having it as an interface to a result is the more natural choice.

vadl/main/vadl/dump/infoEnrichers/LcbEnricherCollection.java
kper: I would actually keep it that way, so you can see the "awkwardness". The "awkwardness" is that I run multiple passes inside of another pass.

vadl/main/vadl/dump/BehaviorTimelineDisplay.java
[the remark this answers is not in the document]
Jozott00: Ok, then let's keep it

Jozott00 (dismissed):

PR #259, opened 2025-05-18
viam: Allow multiple field references in FieldAccess

Jozott00:
Until now, we assumed that there is exactly one field a field access is referring to. 
However, there might be multiple fields referenced in a single field access. E.g. the AArch64 spec defines
```
  format BitFieldMoveFormat: Instr =  // bit field move format
    { ...
    , immr       : Bits6              // rotate right count (bit field low position)
    , imms       : Bits6              // bit field length -1 or high position
    , ...
    , rightWSize = BitfieldRightWSize (imms(4..0), immr(4..0))
    , rightXSize = BitfieldRightXSize (imms, immr)
    }
```

The `Format.FieldAccess` VIAM definition now holds a list of field references (instead of a single one). The `fieldRef()` method is still available as we currently don't have a good way to handle multiple field references in the LCB.

Additionally, the decoder now correctly handle multiple field references. (@rascmatt)

### DIFF
--- a/vadl/main/vadl/cppCodeGen/common/AccessFunctionCodeGenerator.java
+++ b/vadl/main/vadl/cppCodeGen/common/AccessFunctionCodeGenerator.java
@@ -16,12 +16,16 @@
 
 package vadl.cppCodeGen.common;
 
+import static java.util.function.Function.identity;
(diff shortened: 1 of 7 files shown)


kper (approved):

kper:
Note that this won't work for the LCB.

PR #262, opened 2025-05-18
cli: Add default ISA lowering of root ISA

Jozott00:
This PR lowers the root ISA, if there was no generator entry definition (such as `processor`) and a single root ISA could be determined.

Additionally the VDT generator is now executed on `check`, to validate that there are no overlapping instructions.

Currently, the VDT generator throws runtime exceptions if there are problems with the specification. @rascmatt will fix this by throwing detailed diagnostic user errors.

### DIFF
--- a/vadl-cli/main/vadl/cli/CheckCommand.java
+++ b/vadl-cli/main/vadl/cli/CheckCommand.java
@@ -16,13 +16,11 @@
 
 package vadl.cli;
 
-import java.io.IOException;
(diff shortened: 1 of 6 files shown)


Jozott00:
@AndreasKrall when this is merged, `check` on the current `aarch64.vadl` spec will fail, as there are overlapping instruction encodings.

flofriday (approved):
Looks good to me 👍

PR #266, opened 2025-05-19
ci: Added aarch64 images

kper:
This PR adds an image for aarch64 qemu and an image with an aarch64 toolchain (bare metal and linux gnu).

### DIFF
--- a/docker/iss/aarch64-toolchain/Dockerfile
+++ b/docker/iss/aarch64-toolchain/Dockerfile
@@ -1,0 +1,63 @@
+# Base build stage for all architectures
+FROM ubuntu:22.04 AS build
+
+WORKDIR /work
+
+RUN apt update
+RUN apt install -y wget tar xz-utils
+
+WORKDIR /work
+RUN mkdir /opt/aarch64
+
+RUN set -ex; \
+    ARCH=$(uname -m); \
+    case "$ARCH" in \
+        x86_64) SUFFIX="x86_64-aarch64-none-linux-gnu.tar.xz" ;; \
+        aarch64) SUFFIX="aarch64-aarch64-none-linux-gnu.tar.xz" ;; \
+        *) echo "Unsupported architecture: $ARCH" >&2; exit 1 ;; \
+    esac; \
+    wget -O /work/output.tar.xz "https://developer.arm.com/-/media/Files/downloads/gnu/14.2.rel1/binrel/arm-gnu-toolchain-14.2.rel1-$SUFFIX"; \
+    tar --strip-components=1 -xf /work/output.tar.xz -C /opt/aarch64
+
+FROM ubuntu:22.04 AS final
+COPY --from=build /opt/aarch64 /opt/aarch64
+
+ENV PATH=/opt/aarch64/bin:$PATH
+
+RUN apt update && \
+    apt install -y python3  \
+    python3-pip libmpc-dev libmpfr-dev libgmp-dev build-essential  \
+    gperf libtool patchutils bc zlib1g-dev libexpat-dev       \
+    libglib2.0-dev libslirp-dev
+
+# Testing toolchain commands
+
+RUN aarch64-none-linux-gnu-addr2line --help; \
+ aarch64-none-linux-gnu-ar --help; \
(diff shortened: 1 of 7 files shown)


docker/iss/aarch64-toolchain/README.md
```diff
@@ -0,0 +1,56 @@
...
+On both machines execute the following
+
+```bash
+docker buildx build \
+  -t <username>/riscv-toolchain \
+  -f Dockerfile \
+  --platform linux/arm64 \
+  --output push-by-digest=true,type=image,push=true \
+  .
+```
```
Jozott00, line 21: I think the platform flag should be removed, as the built image should use the host machine's architecture.

docker/iss/aarch64-toolchain/Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+RUN apt update
+RUN apt install -y wget tar
+
+WORKDIR /work
+RUN wget https://developer.arm.com/-/media/Files/downloads/gnu/14.2.rel1/binrel/arm-gnu-toolchain-14.2.rel1-x86_64-aarch64-none-elf.tar.xz
+RUN tar -xf arm-gnu-toolchain-14.2.rel1-x86_64-aarch64-none-elf.tar.xz
+
+RUN wget https://developer.arm.com/-/media/Files/downloads/gnu/14.2.rel1/binrel/arm-gnu-toolchain-14.2.rel1-x86_64-aarch64-none-linux-gnu.tar.xz
+RUN tar -xf arm-gnu-toolchain-14.2.rel1-x86_64-aarch64-none-linux-gnu.tar.xz
+
```
Jozott00, line 15: This currently only installs the x64 compiled ARM-toolchain. It is always copied to the final image even if the image is uses the linux/arm64 platform.

docker/iss/aarch64-toolchain/README.md
kper: ah ok
I copied it from the other README

docker/iss/aarch64-toolchain/Dockerfile
kper: How do I make it conditional?

Jozott00: There are some variables that can be used, but I don't know how to conditionally copy something 🧐 Also we should make the usage of the toolchain the same on both platforms, so we would need to rename the arm toolchain on the linux/arm64 image to have the same names as on the x64 image.

Jozott00: https://github.com/OpenVADL/openvadl/blob/master/docker/iss/qemu/Dockerfile has some conditional execution depending on the image platform

kper: > There are some variables that can be used, but I don't know how to conditionally copy something 🧐 Also we should make the usage of the toolchain the same on both platforms, so we would need to rename the arm toolchain on the linux/arm64 image to have the same names as on the x64 image.

ah yeah
makes sense

Jozott00: You could also use arch variables to conditionally import from different build stages.

This is from an old version of the riscv toolchain image:
https://github.com/OpenVADL/openvadl/blob/d1d5ec734329d86b6d98f30956564193c9d006f4/docker/iss/riscv-toolchain/RiscvToolchain.Dockerfile

docker/iss/aarch64-toolchain/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+RUN apt install -y wget tar xz-utils
+
+WORKDIR /work
+RUN mkdir /opt/aarch64
+
+RUN set -ex; \
+    ARCH=$(uname -m); \
+    case "$ARCH" in \
+        x86_64) SUFFIX="x86_64-aarch64-none-linux-gnu.tar.xz" ;; \
+        aarch64) SUFFIX="darwin-arm64-aarch64-none-elf.tar.xz" ;; \
```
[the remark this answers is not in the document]
kper: ah yes
I must not look under the MacOs section

Jozott00 (approved):

PR #272, opened 2025-05-20
decode: Extract and evaluate format fields from encoding

rascmatt:
Allow in memory access and evaluation of format fields and field accesses. This is in preparation of the simple in-memory disassembler proposed by @Jozott00 for test-case generation.

There already exist in-memory decoding functionality (`vadl.vdt.target.common.DecisionTreeDecoder`), but until now there was no possibility to actually extract the decoded fields. This change provides utilities to extract both actual encoded format fields as well as evaluation of field access functions.

- TODO: implement full builtin library


vadl/main/vadl/utils/evaluate/impl/PureFunctionEvaluator.java
```diff
@@ -0,0 +1,228 @@
...
+  protected void handle(EvaluationContext ctx, BuiltInCall toHandle) {
+
+    final List<Constant.Value> args = toHandle.arguments()
+        .stream()
+        .map(this::evaluate)
+        .toList();
+
+    var result = VadlBuiltInEvaluator.evaluate(toHandle, args);
+    ctx.setResult(result);
+  }
```
Jozott00, line 163: You don't need your own built-in evaluator. A built-in has the `#compute` method which computes the method given constant params. So you can remove your custom `VadlBuiltInEvaluator`. 

There is also the `Canonicalizer` with the `#canonicalizeSubGraph` method. It takes a node (in your case the return value expression) and evaluates/transforms it.

rascmatt: Ok, thanks for the hint :).

As discussed, I removed the evaluation of field access functions entirely for now. 
Also, see the `DecisionTreeDecoderTest` on an example on how to use the decoder & extract fields from the result.

Jozott00 (approved):

PR #282, opened 2025-05-22
chore: Fix checkstyle IDEA plugin scan scope

linushdot:
Make checkstyle IDEA plugin scan all java files _including tests_ in order to match the checks done in gradle task checkstyleAll (runs in CI)

This also adds `config/checkstyle/checkstyle.xml` to the active configuration used by default when running checks with the plugin. By this, the VADL configuration does not have to be selected manually. Not sure if this has any drawbacks.


kper:
Sry, I didn't intented to close this

PR #285, opened 2025-05-23
ast: Fixes !283

kper:
Allows the usage of `X(1)` in `SequenceCallExpr`.


vadl/main/vadl/ast/Expr.java
```diff
@@ -2712,13 +2715,13 @@ enum Operation {
 class SequenceCallExpr extends Expr {
 
   @Child
-  Identifier target;
+  Expr target;
```
Jozott00, line 2718: I think the most restrictive type that includes `Identifier` and `CallIndexExpr` is `IsCallExpr`. It can be either a `CallIndexExpr` or a `IsSymExpr`.

vadl/main/vadl/ast/Expr.java
```diff
@@ -2749,3 +2752,58 @@ void prettyPrintExpr(int indent, StringBuilder builder, Precedence parentPrec) {
     }
   }
 }
+
+class ExpandedSequenceCallExpr extends Expr {
```
Jozott00, line 2756: I think it would be a good Idea to make this a sealed class, so one can use an exhaustive switch to always handle all possible cases.

vadl/main/vadl/ast/Expr.java
```diff
@@ -2749,3 +2752,58 @@ void prettyPrintExpr(int indent, StringBuilder builder, Precedence parentPrec) {
...
+
+  @Override
+  void prettyPrintExpr(int indent, StringBuilder builder, Precedence parentPrec) {
+    target.prettyPrint(0, builder);
+  }
+}
+
+class ExpandedRegisterSequenceCallExpr extends ExpandedSequenceCallExpr {
+  // a register is not a child because `X1` does not exist as a symbol.
+  Identifier register;
```
Jozott00, line 2789: question: Why is this identifier needed? Is it possible to write something like `X(1){1..4}`?

vadl/main/vadl/ast/ParserUtils.java
```diff
@@ -765,14 +777,33 @@ static List<SequenceCallExpr> expandSequenceCalls(Parser parser, List<SequenceCa
...
       }
     }
     return expandedCalls;
   }
 
+  /**
+   * Map an expression to an {@link Identifier}. It can be already an {@link Identifier}.
+   * Or {@code X(1)} then we map the {@code expr} to {@code X1}.
+   */
+  private static Identifier mapToIdentifier(Expr expr) {
```
Jozott00, line 793: I would make the type a `IsCallExpr`.
Additionally, you should rename this method, as `mapToIdentifier` is rather generic while the implementation is very specific.

vadl/main/vadl/ast/ParserUtils.java
```diff
@@ -765,14 +777,33 @@ static List<SequenceCallExpr> expandSequenceCalls(Parser parser, List<SequenceCa
...
+    if (expr instanceof Identifier ident) {
+      return ident;
+    } else if (expr instanceof CallIndexExpr callIndexExpr) {
+      var indexLit = callIndexExpr.argsIndices.get(0).values.get(0);
+      var index = new ConstantEvaluator().eval(indexLit);
```
Jozott00, line 798: This will lead to a crash if the index can’t be constantly evaluated. You can catch the `EvaluationError` to gracefully handle this situation.

vadl/main/vadl/ast/ParserUtils.java
```diff
@@ -765,14 +777,33 @@ static List<SequenceCallExpr> expandSequenceCalls(Parser parser, List<SequenceCa
...
 
+  /**
+   * Map an expression to an {@link Identifier}. It can be already an {@link Identifier}.
+   * Or {@code X(1)} then we map the {@code expr} to {@code X1}.
+   */
+  private static Identifier mapToIdentifier(Expr expr) {
+    if (expr instanceof Identifier ident) {
+      return ident;
+    } else if (expr instanceof CallIndexExpr callIndexExpr) {
+      var indexLit = callIndexExpr.argsIndices.get(0).values.get(0);
```
Jozott00, line 797: This assumes that there is a single argument group with a single entry. However, the user might write more arguments than one.

vadl/main/vadl/ast/vadl.ATG
```diff
@@ -1917,25 +1917,33 @@ PRODUCTIONS
...
       }
       SYM_BRACK_CLOSE
-    )                                                     (. list = expandSequenceCalls(this, list); .)
+    )                                                             (. list = expandSequenceCalls(this, calls); .)
   .
 
-  sequenceCallExpr<out SequenceCallExpr call> (. var start = nextTokenLoc(); .)
-  = identifier<out Identifier id>             (. Expr range = null; .)
+  sequenceCallExpr<out SequenceCallExpr call>     (. var start = nextTokenLoc(); .)
+  = callOrBinaryExpression<out Expr id, false>    (. Expr range = null; .)
```
Jozott00, line 1934: major: The `callOrBinaryExpression` non-terminal is very powerful.
It supports things like `a::b<3>(1)(1,2)(3,4).c()`. 

However, in this case, the expression is a lot simpler, limited to `<ident>{"(" <expr> 
")"}`. So there are no size expressions (`<3>`), slices (`(1,2)`) or sub calls (`.c()`), but only a number of groups, where each group can have exactly one argument. 
Therefore, I would define a new non-terminal in the grammar that represents such a call, which makes the whole syntax much less error-prone. This is probably also useful in other places in the grammar.

Jozott00 (changes requested):
I think having an extra class `ExpandedSequenceCallExpr` that represents the expanded sequence call is a good approach.

vadl/main/vadl/ast/Expr.java
```diff
@@ -2749,3 +2752,58 @@ void prettyPrintExpr(int indent, StringBuilder builder, Precedence parentPrec) {
...
+    super(registerFile, loc);
+    this.register = register;
+    this.callIndexExpr = callIndexExpr;
+  }
+
+
+  @Override
+  void prettyPrintExpr(int indent, StringBuilder builder, Precedence parentPrec) {
+    target.prettyPrint(0, builder);
+    register.prettyPrint(0, builder);
```
Jozott00, line 2807: major: I think the output of this pretty print can't be used as input again, right?

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -3062,6 +3062,18 @@ public Void visit(SequenceCallExpr expr) {
...
+    throwUnimplemented(expr);
+    return null;
+  }
+
+  @Override
+  public Void visit(ExpandedAliasDefSequenceCallExpr expr) {
+    throwUnimplemented(expr);
+    return null;
+  }
+
```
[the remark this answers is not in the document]
Jozott00: Outdated, as you typecheck during VIAM lowering.

Jozott00 (approved):

vadl/main/vadl/ast/Expr.java
kper: Yes, this should work

In our call, we have said `IsSymExpr` and that didn't work

Jozott00 (approved):

Issue #82, opened 2025-03-23
Empty or Invalid Identifier Names should be disallowed


AndreasKrall:
The following VADL specification passes the frontend and it seems that it is possible to have an empty identifier name which works fine:
```
constant ExtendId( "" ) = 3
constant b = ExtendId( "" ) + 3
```
The AST dump shows that it works fine:
```
ConstantDefinition
. Identifier name: ""
. IntegerLiteral literal: 3 (3), type: Const<3>
ConstantDefinition
. Identifier name: "b"
. BinaryExpr operator: +, type: Const<6>
. : Identifier name: "", type: Const<3>
. : IntegerLiteral literal: 3 (3), type: Const<3>
```
But the expanded macros from `check --expand-macros` cannot be parsed any more:
```
constant  = 3
constant b = ( + 3)
```
I believe it is a funny feature to have empty identifiers which are working, but I think we should forbid them as the expanded macros cannot be parsed any more.
```
constant ExtendId( "3" ) = 3
constant b = ExtendId( "3" )
```
It seams that anything works:
```
ConstantDefinition
. Identifier name: "3"
. IntegerLiteral literal: 3 (3), type: Const<3>
ConstantDefinition
. Identifier name: "b"
. Identifier name: "3", type: Const<3>
```
We should secure `ExtendId` to check if the result is a valid identifier. Of course we also could fix `--expand-macros` to enclose invalid identifier names in an `ExtendId`. The advantage of the second solution would be that it is possible to use `mnemonic` also for identifiers like `add.s` which are no valid VADL identifiers.

flofriday:
In an meeting we discussed that all keywords that don't cause an LL1 conflict should be added to the allowed keywords list.

Even more, there only should be a single source of truth for the list so that the two lists cannot get out of sync in the future.

flofriday:
I couldn't find a good way to unify the lists, but I added comments that clarify the linkage between the two implementations.

PR #303, opened 2025-05-28
parser: Fix empty identifiers

flofriday:
Identifiers expanded from ExtendId can no longer be empty. #82

## Example
```vadl
constant ExtendId( "" ) = 3
```

```
error: Invalid Empty Identifier
     ╭──[file:///Users/flo/src/open-vadl/tmp.vadl:1:10]
     │
   1 │ constant ExtendId( "" ) = 3
     │          ^^^^^^^^^^^^^^ This expands to an empty identifier name.
     │ 
     note: Empty identifiers are forbidden as they needlessly obfuscate the code.
```


AndreasKrall (approved):
looks fine

PR #308, opened 2025-05-28
parser: Fix inherit records from extended ISA

flofriday:
Records can now also be inherited from the extened ISA.

### DIFF
--- a/vadl/main/vadl/error/DiagnosticBuilder.java
+++ b/vadl/main/vadl/error/DiagnosticBuilder.java
@@ -18,6 +18,8 @@ package vadl.error;
 
 import com.google.errorprone.annotations.FormatMethod;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
 import vadl.utils.SourceLocation;
 import vadl.utils.WithLocation;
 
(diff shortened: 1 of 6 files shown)


vadl/main/vadl/error/DiagnosticBuilder.java
```diff
@@ -145,6 +146,25 @@ public DiagnosticBuilder help(String content, Object... args) {
     return this;
   }
 
+  /**
+   * Attaches a help message of available options.
+   *
+   * @param options to display the user.
+   * @return the builder itself.
+   */
+  public DiagnosticBuilder suggestions(List<String> options) {
```
flofriday, line 155: **question:** Does this method and it's signature make sense? @Jozott00 

I implemented suggestions a couple of PR's ago but wasn't quite happy with the way the code was written, as conditionally adding something to a diagnostic is somewhat verbose/cumbersome with the Diagnostic Builder.

And since suggestions are a somewhat common thing in the frontend I thought of adding them to the builder, where conditions a much cleaner.

One of my concerns though is if this method should be a rather simple building block like it is right now where the caller already has filtered through all available possibilities and selected the ones that they want to display.
Or should the method receive all available options and filter them out, and therefore take some control away from the caller?

Jozott00: I think the method makes sense. However, it would be helpful if the method itself sorts the options. We could make the signature `suggetsions(List<String> options, boolean keepOrder)`, so the user can choose whether to sort them or not.

Jozott00 (dismissed):

vadl/main/vadl/error/DiagnosticBuilder.java
flofriday: Well, you _also_ need to reduce the number of options to something manageable so `Leivenshtein.suggestion` already sorts it and reduces the number so it can be printed in error messages.

So I think, either the caller should sort and reduce before calling the diagnostic builder or the diagnostic builder should do it but I wouldn't split it up.

And right now we are always do the same `Levenshtein.suggestions` but maybe in the future there might be a day where you want to have more controll over what is suggested and in which order? 🤔

flofriday: In an private discussion we decided to keep the API as it is for now and stay open to further changes if the requirements change.

PR #314, opened 2025-05-29
frontend: Rename ExtendId and IdToStr to AsId and AsStr

flofriday:
As discussed in #103 `AsStr` now also takes multiple arguments like `AsId`.

The PR also addresses an issue with `AsId` and now only allows valid identifiers. #82


AndreasKrall (approved):
is fixed and fine

PR #318, opened 2025-05-30
decoder: Support subsumed instructions in irregular VDT generator

rascmatt:
This commit adds the option to try to automatically synthesize some constraints on the encodings, specifically to automatically detect subsumed encodings, i.e. two (or more) instructions of the format, where one instructions is a more specific (i.e. specifies more fixed bits) encoding of the other instruction.

We now automatically exclude the more specific encoding from the general case, so the general instruction is only matched if none of the more specific ones do.

E.g. the `ADDW` and `ADDWLSL` instruction definitions in the current `Aarch64` Spec, where `ADDW` is the more specific case of `ADDWLSL` because it additionally specifies the encoding of `imm6`:

Encoding for `ADDWLSL`: `op = 0x16`, `sf = 0x0`, `ff = 0x0`, `shift = 0x0`
Encoding for `ADDW`: `op = 0x16`, `sf = 0x0`, `ff = 0x0`, `shift = 0x0`, `imm6 = 0x0`

---

TODO: Extend the CLI to let the user disable/enable this behaviour and or switch between decoder generators (regular vs. irregular)

### DIFF
--- a/vadl/main/vadl/vdt/impl/irregular/IrregularDecodeTreeGenerator.java
+++ b/vadl/main/vadl/vdt/impl/irregular/IrregularDecodeTreeGenerator.java
@@ -28,6 +28,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
+import vadl.error.Diagnostic;
 import vadl.vdt.impl.irregular.model.DecodeEntry;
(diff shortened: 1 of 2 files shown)


rascmatt:
Also improved the exception handling a bit. The irregular decision tree generator is now enabled by default. In the future we'll allow to select the generator by a command line argument.

PR #320, opened 2025-05-31
vdt: Add encoding constraint

Jozott00:
This PR adds the encoding constraint annotation (#319) and implements it in the VDT generator.


benedikt-l-huber (approved):
LGTM

PR #321, opened 2025-05-31
typechecker: Fix wrong long multiply type

flofriday:
Because the typechecker didn't return after handling one specific special case it sometimes continued with the generic case, resulting in the wrong type.


AndreasKrall (approved):
looks fine

PR #325, opened 2025-06-03
lcb: Generate the OpenVADL Linker

benjaminkasper99:
This PR implements the linker in the LCB and tests assembler + linker with the `rv32im` specfication. 

To be able to assemble output from clang / gcc I had to make some changes to the  `rv32im` / `rv3264im` specfications:
- Add the `JR` pseudo instruction
- Add directives and modifiers to the assembly description
- Adjust pseudo instructions to use the correct machine instruction and relocations
- Allow the assembler to parse modifiers for the immediate operand of `JALR`

Note: This PR seems large, but most of the reported changes are due to the two template files `lld/ELF/Driver.cpp` and `lcb/lld/ELF/Target.h` which I have updated from the LLVM 17 version to their LLVM 19 version.


vadl/main/resources/templates/lcb/lld/ELF/Arch/ImmediateUtils.h
```diff
@@ -2,11 +2,13 @@
...
+#include "vadl-builtins.h"
+
 
 // collection of all available immediates
 
-«FOR immediate : immediates»
-    #include "«processorName»Immediates/«immediate.loweredImmediate.identifier».hpp"
-«ENDFOR»
+[# th:each="function : ${encodeFunctions}" ]
+static [(${function})]
```
kper, line 11: minor: `static` means that the encoding function will not be exported in the object file and thus if you include it then it will not available. Is that really what you want?

vadl/main/resources/templates/lcb/lld/ELF/Arch/Target.cpp
```diff
@@ -10,7 +10,7 @@
 #include "Symbols.h"
 #include "SyntheticSections.h"
 #include "Target.h"
-#include "[(${namespace})]ManualEncoding.hpp"
+#include "[(${namespace})]ManualEncoding.cpp"
```
kper, line 13: minor: you're including the `.cpp` file

vadl/main/resources/templates/lcb/lld/ELF/Driver.cpp
```diff
@@ -82,2019 +84,2043 @@ Ctx elf::ctx;
...
+      warn(mb.getBufferIdentifier() + ": duplicate ordered symbol: " + s);
+
+  return names.takeVector();
+}
+
+static bool getIsRela(opt::InputArgList &args) {
+  // The psABI specifies the default relocation entry format.
+  bool rela = is_contained({EM_AARCH64, EM_AMDGPU, EM_HEXAGON, EM_LOONGARCH,
+                            EM_PPC, EM_PPC64, EM_RISCV, EM_S390, EM_X86_64},
+                           config->emachine);
```
kper, line 1180: major: I think our `EM_` is missing here

vadl/main/vadl/gcb/passes/relocation/model/AutomaticallyGeneratedRelocation.java
```diff
@@ -170,4 +170,9 @@ public Format.Field field() {
   public Format format() {
     return format;
   }
+
+  @Override
+  public String llvmKind() {
+    return kind.llvmKind();
+  }
```
kper, line 177: minor: just add a method to return the kind and all the users can call llvmKind themselves.

vadl/main/vadl/gcb/passes/relocation/model/HasRelocationComputationAndUpdate.java
```diff
@@ -74,4 +74,9 @@ public interface HasRelocationComputationAndUpdate {
    * Generates and returns the name of the ELF relocation.
    */
   ElfRelocationName elfRelocationName();
+
+  /**
+   * Get the LLVM kind of the relocation.
+   */
+  String llvmKind();
```
kper, line 81: minor: just add a method to return the kind and all the users can call llvmKind themselves.

vadl/main/vadl/gcb/passes/relocation/model/ImplementedUserSpecifiedRelocation.java
```diff
@@ -125,4 +125,9 @@ public ElfRelocationName elfRelocationName() {
             + "_" + field.identifier.simpleName()
     );
   }
+
+  @Override
+  public String llvmKind() {
+    return kind.llvmKind();
+  }
```
kper, line 132: minor: just add a method to return the kind and all the users can call llvmKind themselves.

vadl/main/vadl/lcb/template/lld/ELF/Arch/EmitLldArchFilePass.java
```diff
@@ -59,12 +81,26 @@ protected Map<String, Object> createVariables(final PassResults passResults,
...
+    var elfRelocations = output.elfRelocations().stream().map(
+        r -> new ElfRelocationInfo(r.elfRelocationName().value(),
+            r.llvmKind(),
+            r.valueRelocation().functionName().lower(),
+            r.fieldUpdateFunction().functionName().lower(),
+            r instanceof AutomaticallyGeneratedRelocation
+                ? requireNonNull(encodingFunctions.get(r.field())).functionName().lower()
+                : ""
+        )
+    ).toList();
```
kper, line 96: major: Why do you exclude the other relocations?

kper (changes requested):
I think that it is great that you have assembly tests that are linked and executed. But is there a reason why you also run the c inputs to object files, link them and execute them. Is the assembler not working?

vadl/test/resources/images/lcb_execution_test_rv32im/base_image/asm.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -x
+
+/src/llvm-final/build/bin/llvm-mc -filetype=obj -arch=${TARGET} /src/inputs/$INPUT -o /tmp/main.o
+/src/llvm-final/build/bin/llvm-mc -filetype=obj -arch=${TARGET} /helper/init.S -o /tmp/init.o
+
+/src/llvm-final/build/bin/ld.lld -static -T/helper/link.ld /tmp/main.o /tmp/init.o -o /tmp/main
```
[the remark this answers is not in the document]
benjaminkasper99: Did you maybe try to link libraries compiled with gcc using the llvm linker? Because I ran into this incompatibility between libraries and the OpenVADL linker tests.

vadl/main/resources/templates/lcb/lld/ELF/Driver.cpp
benjaminkasper99: fixed

benjaminkasper99:
> Is there a reason why you have this `vadl/test/resources/llvm/riscv/asm/rv32im/qemu_c/` folder? We already have a `vadl/test/resources/llvm/riscv/asm/rv32im/spike` folder. Why isn't that sufficient?
> 
The `.../qemu_c` directory only contains the subset of the tests in `.../spike` which do not require libc. As mentioned above the libc from the riscv toolchain is compiled and assembled with gcc and therefore incompatible / cannot be linked.
> Also you don't test it. You only have added a test for the asm.
> 
They are used in the `LcbRiscv32SimulationTest`.

kper:
> The .../qemu_c directory only contains the subset of the tests in .../spike which do not require libc. As mentioned above the libc from the riscv toolchain is compiled and assembled with gcc and therefore incompatible / cannot be linked.

hmmm
while, I understand the problem, I just don't want to maintain two folders with the same files. We can gladly keep them separate but I think then you need to remove them from `spike` and make the previous simulation tests also run the `qemu_c` files. Then we get both what we want. I have one source of truth and you have your separation.
You probably also need to rework the folder structure since it must be the same for riscv64.

> They are used in the LcbRiscv32SimulationTest.

ah yes, now I see it

PR #342, opened 2025-06-08
iss: Add TB-Layer to cosimulation

Giftzwerg02:
This PR implements cosimulation at the translation-block layer, both `tb` and `tb-strict` are added as values to the `testing.protocol.layer` configuration option:

- `tb-strict`: Assumes that equivalent translation-blocks are generated by all QEMU-clients - any divergences result in a test-fail.
- `tb`: Translation-blocks do not have to be equivalent, rather the PC is compared. This allows for clients with differently generated TBs but equivalent instructions. 

For `tb`, the clients currently synchronize as follows:

1. Each client gets one execution step, the PC before the exec and after is collected and added to a list.
2. After every client ran once, a call to `sync_clients` is made:
3. The largest PC (after exec-step) is selected.
4. All clients with a after-exec PC smaller than the maximum are added into a queue.
5. While there are clients in the queue:
5.1 Run the client once and check the PC after it has run
5.2 if it is still **smaller** than the max PC, re-add it to the queue
5.3 if it is **equal** than the max PC, then the client is synchronized and not re-added to the queue
5.4 if it is **larger**  than the previous max PC, then this is currently not accounted for and the test will fail

The synchronization is done once the queue is empty - leading to a state-comparison and then going back to step 1.

The synchronization currently assumes that there will always be a common available sync-point between all clients, this is usually some sort of jump instruction since TBs should never contain a jump in the middle of the block.

However, if there is no TB from one client which is a "superset" of the other TBs from the other clients then this sync-algorithm will fail, specifically at step 5.4.

If step 5.4 never gets executed then this synchronization is probably ideal, since it would maximize the number of state comparisons - which is desired for lock-stepping. Otherwise (if this becomes an issue) I would need re-synchronize the clients based on the new max PC - which might lead to very late synchronization points at worst, which would hardly make it a lock-step test.

### DIFF
--- a/vadl-cosim/src/annotate-fields.py
+++ b/vadl-cosim/src/annotate-fields.py
@@ -23,9 +23,77 @@ The generated __init__ method is printed to stdout, simply add it to the class-s
 
 import re
 from typing import Optional
+from dataclasses import dataclass
+from pprint import pprint
 
 fields = '[("size", c_size_t), ("buffer", c_uint8 * MAX_INSN_DATA_SIZE)]'
 
+@dataclass
+class CField:
+    ctype: str
+    name: str
+    array_mod: Optional[str]
+
+@dataclass
+class CStruct:
+    name: str
+    cfields: list[CField]
+
+def parse_cfield(input: str) -> CField:
+    expr = re.compile(r"\s*(.*?)\s+(.*?)\s*(\[(.*?)\])?\s*;")
+    found = expr.findall(input)[0]
+    ctype = found[0]
+    name = found[1]
+    array_mod = found[3]
+    if array_mod.strip() == "":
+        array_mod = None
+    cfield = CField(ctype, name, array_mod)
+    return cfield
+
+def parse_cstructs(input: str) -> list[CStruct]:
+    expr = re.compile(r"\s*typedef\s*struct\s*{((.|\s)*?)}\s*(.*?)\s*;", re.MULTILINE)
+    found = expr.findall(input)
+    cstructs = []
+
+    for f in found:
+        cfields = filter(lambda s: len(s) > 0, f[0].split("\n"))
+        cfields = list(map(parse_cfield, cfields))
+        name = f[2]
+        cstructs.append(CStruct(name, cfields))
+
+    return cstructs
+
+def into_python_str(cstruct: CStruct) -> str:
+    out = ""
+
+    for field in cstruct.cfields:
+        if field.array_mod is not None:
+            out += f"{field.array_mod} = <TODO>\n"
+
+    out += "_fields_ = "
+
+    def format_field(field: CField) -> str:
+        if field.array_mod is not None:
+            return f"(\"{field.name}\", {c_typ(field.ctype)} * {field.array_mod})" 
+        else:
+            return f"(\"{field.name}\", {c_typ(field.ctype)})"
+
+    fs = ", ".join([format_field(f) for f in cstruct.cfields])
+    out += f"[{fs}]\n\n"
+
+    out += "def __init__(self, *args: Any, **kw: Any) -> None:\n"
+    out += "\tsuper().__init__(*args, **kw)\n"
+
+    def format_init_field(field: CField) -> str:
+        if field.array_mod is not None:
+            return f"\tself.{field.name}: Annotated[{python_array_typ(field.ctype)}, {c_typ(field.ctype)} * {field.array_mod}]"
+        else:
+            return f"\tself.{field.name}: Annotated[{python_typ(field.ctype)}, {c_typ(field.ctype)}]"
+
+    fs = "\n".join([format_init_field(f) for f in cstruct.cfields])
+    out += fs
+    return out
+        
 
 def array_typ(typ: str) -> Optional[tuple[str, str]]:
     expr = re.compile(r"^(.*?)\*(.*?)$")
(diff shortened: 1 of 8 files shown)


Jozott00:
> However, if there is no TB from one client which is a "superset" of the other TBs from the other clients then this sync-algorithm will fail, specifically at step 5.4.

As you already stated, your algorithm fails under certain conditions. 
E.g. consider the example:
client1: 
```
PC
0: start: j start
```
client2:
```
PC
0: nop
1: start: j start
```

When you execute one TB on each client, client1 will be at `PC: 0` while client2 will be at `PC: 1`. From your algorithm description, client2 will "try" to catch up with client1 as long as its PC is smaller than the PC of client1. This will always be the case, as `client2` is in an infinite loop and never catches up. So your algorithm would not detect that client2 and client1 diverged and won't terminate either.

Now my suggestion is to detect and operate on executed basic block.
This means, instead of letting one client run a single TB, you let it run until there was a jump. As you know the TB's start and end address, you can detect when a TB executes at a non-continues address (`(tb-1).endAddress + 4 != tb.startAddress`). 
Now if there was a jump, you stop executing the client and you do the same on the other one. 

Now you’re at a synchronization point where you can compare the PCs of both clients. If they don't match, then they diverged at some point.

To provide a more specific diverging location, you can compare the TB end address of each executed block on the second client with the jump PC of the first one. (similar to what you described in your algorithm)

With this approach you would always detect diverging execution.

(I hope I’ve understood your algorithm correctly)

vadl-cosim/src/annotate-fields.py
```diff
@@ -35,41 +104,69 @@ def array_typ(typ: str) -> Optional[tuple[str, str]]:
...
+typedef struct {
+  int init_mask;
+  SHMCPU cpus[MAX_CPU_COUNT];
+  TBInfo tb_info;
+} BrokerSHM_TB;
+"""
+
+cstructs = parse_cstructs(s)
+into_python = [into_python_str(cstruct) for cstruct in cstructs]
+print("\n\n".join(into_python))
```
Jozott00, line 145: question: What is this doing? I’m not sure what it is used for.

vadl-cosim/src/cosimulation_broker.py
```diff
@@ -291,16 +107,154 @@ class Report:
...
+
+            diffs.extend(diff_register(c1reg, c2reg, cpu_index, reg_index, config))
+
+    for reg_index in range(min(cpu1.registers_size, cpu2.registers_size)):
+        c1reg = cpu1.registers[reg_index]
+        c2reg = cpu2.registers[reg_index]
+
+        diffs.extend(diff_register(c1reg, c2reg, cpu_index, reg_index, config))
+
+    return diffs
```
Jozott00, line 204: question: What is the loop in line `198` doing? Isn't it already checked with the two loops above? Also, the loop assumes that the register indices are the same on both CPUs, but this might not be the case. So I think `cpu1: {pc, rd}` and `cpu2: {rd, pc}` would cause a diff in that for loop check.

vadl-cosim/src/cosimulation_broker.py
```diff
@@ -413,40 +365,145 @@ def compare_client_state() -> list[ClientDiff]:
...
+                        client.sem_client.release()
+
+                        # wait at most 0.1 second, if an error occurs: assume that the client finished (crashing = finishing)
+                        client.sem_server.acquire(0.1)
+                    except ipc.BusyError:
+                        logger.debug(
+                            f"BusyError: noticed that client #{client.id} shutdown. marking as closed"
+                        )
+                        client.is_open = False
+
```
Jozott00, line 388: minor: I would extract this into its own function.

vadl-cosim/src/cosimulation_broker.py
```diff
@@ -413,40 +365,145 @@ def compare_client_state() -> list[ClientDiff]:
...
+            diffs += compare_client_state()
+
+            # early exit for lockstepping
+            if len(diffs) > 0:
+                return report_from_diffs(diffs)
+    else:
+        # TB-Syncing:
+        # Why?: Necessary to ensure that on a TB-level cosimulation, the state of all clients is compared at the same PC
+        # Syncing is only done for the "tb" layer, the "tb-strict" layer assumes that all TBs across all clients are equal.
+        def sync_clients(clients_tbs: list[tuple[int, int, int]]):
```
Jozott00, line 404: minor: Maybe move this out of the else. It gets a bit hard to read this way.

vadl-cosim/src/cosimulation_broker.py
```diff
@@ -413,40 +365,145 @@ def compare_client_state() -> list[ClientDiff]:
...
+            if len(clients_tbs) == 0:
+                return
+
+            # NOTE: These asserts are assumptions by the syncing algorithm
+            #       If they fail then that indicates a bug in the algorithm and not in any of the QEMU-clients
+            for i, client_tbs in enumerate(clients_tbs):
+                assert client_tbs[0] < client_tbs[1], (
+                    f"Jumps should not occur inside a TB: #{i}: ({client_tbs})"
+                )
```
Jozott00, line 413: question: I’m not sure if I understand this check. Why is it impossible that the PC of translation block 2 is less than translation block 1? The message `Jumps should not occur inside a TB` is also misleading, as this is not a jump with a TB, but between TBs. 

Example:
```
start: j label2

label1: j end

label2: j label1

end:
```
This would produce 4 translation blocks (each for each label). They are executed in the order `start, label2, label1, end`. `PC(label1) < PC(label2)` even-though `label1` was executed after `label2`.

Jozott00 (changes requested):

vadl-cosim/src/cosimulation_broker.py
```diff
@@ -413,40 +365,145 @@ def compare_client_state() -> list[ClientDiff]:
...
+                client_entry = clients_queue.popleft()
+                client = clients[client_entry[2]]
+
                 try:
+                    tb1 = client.shm_struct.shm_tb.tb_info.pc
                     client.sem_client.release()
 
                     # wait at most 0.1 second, if an error occurs: assume that the client finished (crashing = finishing)
                     client.sem_server.acquire(0.1)
+                    tb2 = client.shm_struct.shm_tb.tb_info.pc
```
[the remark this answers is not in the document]
Giftzwerg02: Good point, I generally refactored all the ipc-related code to its own file now - I think it should be cleaner now.

vadl/main/resources/templates/iss/contrib/plugins/cosimulation.c
```diff
@@ -122,31 +124,18 @@ typedef struct {
...
-  const gchar *client_id;
-  ExecMode mode;
-  const gchar *client_name;
-  gboolean client_name_set;
-} Arguments;
-
-typedef struct {
-  size_t size;
-  TBInfo infos[TBINFO_ENTRIES];
+  int init_mask;
```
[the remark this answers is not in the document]
Giftzwerg02: I need to keep track which CPUs were actually used during the execution of a QEMU-client, since I don't want to compare uninitialized data in the shared memory between clients. Therefore, when an exec-callback in the QEMU-Plugin uses a `vcpu_index` I add the (used) index to the `init_mask` to flag the CPU as used and therefore include it in the diffing.

vadl-cosim/src/cosimulation_broker.py
```diff
@@ -328,218 +243,203 @@ def run_lockstep(config: Config, traces: deque[Trace]) -> Report:
...
+
+@dataclass
+class ClientSyncInfo:
+    start_pc: int
+    end_pc: int
+    tb_size: int
+    client_idx: int
+
+    def is_jump(self) -> bool:
+        return self.start_pc + self.tb_size * 4 != self.end_pc
```
Jozott00, line 436: minor: Fine for now, but in future the instruction size must be settable in the config. For risc-v and aarch64 a 4-byte instruction width is correct, but not necessarily for every architecture.

vadl-cosim/src/cosimulation_broker.py
```diff
@@ -328,218 +243,203 @@ def run_lockstep(config: Config, traces: deque[Trace]) -> Report:
...
-                    client.sem_client.release()
-
-                    # wait at most 0.1 second
-                    # if a client already closes in this phase then simple mark it as closed
-                    # since, in case the client closed / crashed prematurely, this will be found when checking the client's state
-                    client.sem_server.acquire(0.1)
-                except ipc.BusyError:
-                    logger.debug(f"BusyError: noticed that client #{client.id} shutdown. marking as closed")
-                    client.is_open = False
+                client.run()
```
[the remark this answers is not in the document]
Giftzwerg02: Yes, `client.run()` runs a TB in `tb-strict` and `tb`, which is most likely multiple assembly instructions. I thought this was (at least in some cases) the desired behavior: Should `skip`ping always only work on instructions? If so, then I would probably want to change this in another PR since this would probably introduce more complicated semaphore logic.

vadl-cosim/src/cosimulation_broker.py
```diff
@@ -328,218 +243,203 @@ def run_lockstep(config: Config, traces: deque[Trace]) -> Report:
...
+        while any_client_open(clients):
+            for client in clients:
+                if client.is_open:
+                    client.run()
+
+            if not execute_remaining:
+                if stop_after > 0:
+                    stop_after -= 1
+                else:
+                    return report_from_diffs(diffs)
```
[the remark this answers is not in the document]
Jozott00: Ok, eventually `stop_after_n_instructions` should also work on instruction level in the TB mode. However, I guess this is fairly easy to implement, as you know how many instructions where executed between sync points. It does not have to stop exactly after `n` instructions, but as soon as the number executed instructions exceeds `n` at some sync point.

PR #344, opened 2025-06-10
viam: Add forall(do, tensor, fold) construct to VIAM

Jozott00:
This adds 5 nodes to the VIAM.
Two control nodes: `ForallNode, ForallEndNode`
Three dependency nodes: `FoldNode, TensorNode, ForIdxNode`


benedikt-l-huber (approved):
LGTM

PR #345, opened 2025-06-10
iss: Add `[undefined when]` annotation

Jozott00:
The annotation can be set on instructions and constraints field and fieldAccesses. 
If the condition of the annotation is met, ISS will treat the instruction like an instruction that wasn’t able to be decoded.

The ISS AArch64 test generator uses the annotation to determine whether a generated instruction is valid or not. This makes it possible to avoid generating invalid instructions during test generation.

The annotation is further discussed in #277.


AndreasKrall (approved):
Looks fine

PR #347, opened 2025-06-12
lcb: Make `plt` based on `isPositionIndepedent`

kper:
We had the problem that our compiler attaches the `plt` suffix to all function calls, even when it's not compiled with `PIC`. We added now a check whether we use `PIC` or not.


kper:
Ops, I branched off from another feature branch. That's why we have so many commits here.

Jozott00 (approved):

PR #348, opened 2025-06-12
iss: Fix CPU state register overflow

Jozott00:
The CPU state register type was potentially smaller than the corresponding TCG variable, which caused an overflow when QEMU synchronized the TCG variable with the CPU state.


AndreasKrall (approved):
Looks fine to me

AndreasKrall (approved):
looks fine to me. Just a question, how many different imports from google classes do we have?

Jozott00:
Currently, the project uses the [Google Guava](https://github.com/google/guava) library, which is a common utility library for Java.

PR #352, opened 2025-06-13
frontend: Add forall expression and statement

flofriday:


flofriday:
@Jozott00 can you quickly look over the failing iss instructions and tell me if thats something on your part or (much more likely) that I messed up in the lowering?

Maybe we can sit down and investigate the lowering togehter?

PR #367, opened 2025-06-23
lcb: Refactored predicates

kper:
The compiler generator generated value ranges for the instruction selection. This ranges were based on the field size and were not correct. This has been refactored now. Additionally, we are generating predicate functions only for the field access functions which we can also generate an encoding for. If the user specifies a custom encoding then she has to also define a custom predicate function.


vadl/test/resources/llvm/riscv/asm/rv32im/codeemitter/JType.s
```diff
@@ -3,8 +3,5 @@
 JAL x16, 10
 # CHECK: # encoding: [0x6f,0x08,0xa0,0x00]
 
-JAL x16, 524287
-# CHECK: # encoding: [0x6f,0xf8,0xf7,0x7f]
-
```
kper, line 8: same here

Jozott00 (approved):

PR #369, opened 2025-06-24
lcb: Removed warning

kper:
Removed it because it fills the entire screen with warnings for AArch64.


kper:
> Later we might want to add a CLI option to enable such warnings.

Yeah, but at the moment it's a problem because the errors are at the top and the warnings fill my buffer so I don't see them.

PR #373, opened 2025-06-25
frontend: Add micro architecture

flofriday:


flofriday:
@Jozott00 this is the first time in months that I added some VIAM nodes. Can you look over that part, cause it's quite likely I missed something there or didn't find the cleanest solution.

flofriday:
@linushdot The last test that is failing seems to be from your backend that is now confused with the new VIAM nodes I added. Do you have any suggestions how to tackle this?

Maybe we can introduce a quick and dirty workaround till you fix it in a future PR or should we already implement it in this one? For the later would it make more sense for you to implement it or is it easy enough to explain that I can do it?

vadl/main/vadl/ast/BehaviorLowering.java
```diff
@@ -1154,6 +1192,26 @@ private ExpressionNode visitSubCall(CallIndexExpr expr, ExpressionNode exprBefor
         var indexing =
             new TupleGetFieldNode(subCall.computedStatusIndex, resultExpr, Type.bool());
         resultExpr = visitSliceIndexCall(indexing, Type.bool(), subCall.argsIndices);
+      } else if (exprBeforeSubcall.type() == MicroArchitectureType.instruction()) {
+        // There is weired way to call functions on instructions
+        var builtin =
+            BuiltInTable.builtIns().filter(b -> b.name().equals(subCall.id.name)).findFirst().get();
+        var call = new MiaBuiltInCall(builtin, new NodeList<>(exprBeforeSubcall),
```
Jozott00: minor: The created call has no location

vadl/main/vadl/ast/ViamLowering.java
```diff
@@ -1711,8 +1738,10 @@ public Optional<vadl.viam.Definition> visit(SpecialPurposeRegisterDefinition def
 
   @Override
   public Optional<vadl.viam.Definition> visit(StageDefinition definition) {
-    throw new RuntimeException("The ViamGenerator does not support `%s` yet".formatted(
-        definition.getClass().getSimpleName()));
+    var identifier = generateIdentifier(definition.viamId, definition.identifier());
+    var behaivor = new BehaviorLowering(this).getStageGraph(definition.statement,
+        definition.viamId + "::behavior");
```
Jozott00: This should probably be `definition.viamId.append("behavior")`

vadl/main/vadl/viam/graph/dependency/StageEffectNode.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  @Override
+  public Node shallowCopy() {
+    return new StageEffectNode(miaCall, condition);
+  }
+
+  @Override
+  public <T extends GraphNodeVisitor> void accept(T visitor) {
+    visitor.visit(this);
```
Jozott00: This can be empty. The `GraphNodeVisitor` is deprecated and new nodes don't have to implement it.

PR #381, opened 2025-07-01
lcb: Refactored operand detection

kper:
Before this PR, the generation of tableGen operands was part of the lowering. This means that we have already pruned non-default-branches and lowered the nodes. Unfortunately, this had a major drawback since there are some instructions in Aarch64 which have to behavior after pruning. For these cases, the generation of operands did not work. Now, it is part of the GCB and no pruning is done before the operands haven't been determined.


vadl/test/resources/snapshots/rv64im/InstrInfoTableGen.td
```diff
@@ -1730,7 +1730,7 @@ let Size = 4;
 let CodeSize = 4;
 
 let OutOperandList = ( outs X:$rd );
-let InOperandList = ( ins X:$rs1, X:$rs2 );
+let InOperandList = ( ins X:$rs2, X:$rs1 );
```
kper, line 1733: fyi @benjaminkasper99 because the operands are determined much earlier the order of the operands has changed. However, the emission is still correct. Just keep that in mind when writing the parser tests.

benedikt-l-huber (approved):
LGTM

PR #383, opened 2025-07-03
lcb: Support raw field immediates in assembly string

### LINKED ISSUE #380: lcb: Discussion on immediate fields in assembly strings
### Discussed in https://github.com/OpenVADL/openvadl/discussions/377

<div type='discussions-op-text'>

<sup>Originally posted by **kper** June 29, 2025</sup>
It is possible to define machine instructions with an assembly directive. This assembly directive is used by the compiler for printing an assembly representation of an instruction.

An example would like that

```
  model UtypeInstr (name : Id, opcode : Bin, rhsEx : Ex) : IsaDefs = {
    instruction $name : Utype =                        // upper immediate instructions
      X(rd) := $rhsEx
    encoding $name = {opcode = $opcode}
    assembly $name = (mnemonic, " ", register(rd), ",", hex(imm))
    }

$UtypeInstr (AUIPC; 0b001'0111 ; PC + immUp)         // add upper immediate to PC
$UtypeInstr (LUI  ; 0b011'0111 ;      immUp)         // load upper immediate
```

Note that in the case of `LUI` the behavior is `X(rd) := immUp`. This means that the `LUI` instruction has two operands: `rd` and `immUp`. The assembly directive uses the both the `rd` which is ok since it is an operand and `imm` because `immUp` encodes `imm`. However, `hex(imm)` uses `imm` which is not an operand and instead refers to the binary representation of the instruction. This makes it necessary to encode `immUp` to `imm`. In this particular case, this is quite easy since field access functions and fields have a direct 1:1 relationship. But, if the encoding function requires multiple field access functions, the implementation becomes quite complicated and weird. (1) It's not clear why the assembly view and binary encoding are mixed. (2) It's not clear for the user that an encoding function is applied.
~~I propose to disallow the use of fields which are not operands. In this case `imm` would throw an error, while `rd` is ok.~~ 
A correct specification would be:
```
 assembly $name = (mnemonic, " ", register(rd), ",", hex(some_encoding_function(immUp)))
```

The difference is that `some_encoding_function` is a function which might be equal to a binary encoding function, but no necessarily. This separates both worlds assembly and binary nicely.

</div>

### Conclusion

We concluded in the discussion to keep the status quo and support both `imm` and `immUp`.

### LINKED ISSUE #382: lcb: Signedness and unsignedness for fields in assembly string
This is a follow up for #380. I am integrating the printing of immediate fields into the assembly printer again and found a potential foot gun which I would like to discuss.

Given this format:

  ```
model UtypeInstr (name : Id, opcode : Bin, rhsEx : Ex) : IsaDefs = {
    instruction $name : Utype =                        // upper immediate instructions
      X(rd) := $rhsEx
    encoding $name = {opcode = $opcode}
    assembly $name = (mnemonic, " ", register(rd), ",", hex(imm))
    }
``` 

We are defining in the assembly direct to print `hex(imm)`. This correspondents to the shifted value for `LUI` and `AUIPC`. The generated C++ printing function will look like that:

```
return "ADDI"s
 + " "s
 + getRegisterNameFromXByIndex(MCOperandWrapper(MI->getOperand(0)).unwrapToIntegral()) // rd
 + ","s
 + getRegisterNameFromXByIndex(MCOperandWrapper(MI->getOperand(1)).unwrapToIntegral()) // rs1
 + ","s
 + AsmUtils::formatImm(RV3264Base_ADDI_imm(MI->getOperand(2).getImm()), 10, &MAI);
```

The interesting part is the following expression `RV3264Base_ADDI_imm(MI->getOperand(2).getImm()`. This expression extracts the operand applies the encoding function for the `imm` field.

However, when I run the test suite I get the following immediate printed.

```
(gdb) p MI->getOperand(2).getImm()
$5 = -16
(gdb) p RV3264Base_ADDI_imm(MI->getOperand(2).getImm())
$6 = 4080
```

In other words, the semantic LLVM immediate value is a signed integer. However, all encoding functions will return an unsigned integer because I can't afford to lose a signed bit. This means during printing the default behavior is to print all raw field immediates as unsigned.
I suppose this issue hasn't come up yet because we have explicitly casted it in the previous implementation. However, that might be only correct for RISC-V and not for other architectures.
The question I have is whether that is acceptable / expected behavior and I should update the riscv specification with the necessary casts to make the fields signed?

Proposal:

`assembly $name = (mnemonic, " ", register(rd), ",", hex(imm as Int<32>))`

@AndreasKrall @benedikt-l-huber

kper:
This PR adds support for printing raw fields in assembly. It was necessary to apply an encoding function to get the binary representation of the immediate.

Closes #382 
Closes #380


benedikt-l-huber (approved):
LGTM

PR #384, opened 2025-07-03
asm: Fix immediate decoding in asm parser

benjaminkasper99:
This PR re-implements the immediate decoding in the assembly parser, which was missing after the refactorings of #361.

Note: This implementation only works for architectures where the decoding (= field access) functions only take a single field as input.

### DIFF
--- a/vadl/main/resources/templates/lcb/llvm/lib/Target/AsmParser/AsmParser.cpp
+++ b/vadl/main/resources/templates/lcb/llvm/lib/Target/AsmParser/AsmParser.cpp
@@ -77,28 +77,22 @@ bool [(${namespace})]AsmParser::parse_[(${instruction.name})](MCInst &Inst, Oper
     std::vector<std::string> targets;
     targets =  { [(${instruction.targets})] };
     for(auto target : targets) {
+      bool found = false;
       for(auto index = 1; index < [(${instruction.numOperands})] + 1; index++) {
         [(${namespace})]ParsedOperand& Op = static_cast<[(${namespace})]ParsedOperand&>(*Operands[ index ]);
         StringRef operandName = Op.getTarget();
-        auto isFieldOfOperand = false;
 
         [# th:each="operand : ${instruction.operands}" ]
-        [# th:if="${operand.isImmediate}"]
-        isFieldOfOperand = operandName.equals_insensitive("[(${operand.formatField})]");
-        [/]
-        if(operandName == "[(${operand.name})]" || isFieldOfOperand) {
-          if(target != "[(${operand.name})]") {
-            continue;
-          }
+        if(operandName == "[(${operand.name})]" && target == "[(${operand.targetName})]") {
+          found = true;
+
           if(!Op.isImm() || Op.getImm()->getKind() != MCExpr::ExprKind::Constant) {
               Op.addOperand(Inst);
           } else {
             int64_t opImm64 = dyn_cast<MCConstantExpr>(Op.getImm())->getValue();
 
-            [# th:if="${operand.isImmediate}"]
-            if (isFieldOfOperand) {
-              opImm64 = [(${operand.decodeMethod})](opImm64);
-            }
+            [# th:if="${operand.isFieldOperand}"]
+            opImm64 = [(${operand.decodeMethod})](opImm64);
             [/]
 
             [# th:if="${operand.requiresPredicate}" ]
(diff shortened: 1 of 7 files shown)


vadl/main/resources/templates/lcb/llvm/lib/Target/AsmParser/AsmParser.cpp
```diff
@@ -80,39 +80,49 @@ void [(${namespace})]AsmParser::parse_[(${instruction.name})](MCInst &Inst, Oper
...
           if(!Op.isImm() || Op.getImm()->getKind() != MCExpr::ExprKind::Constant) {
               Op.addOperand(Inst);
           } else {
             int64_t opImm64 = dyn_cast<MCConstantExpr>(Op.getImm())->getValue();
 
+            [# th:if="${operand.isImmediate}"]
+            if (isFieldOfOperand) {
+              opImm64 = [(${operand.decodeMethod})](opImm64);
+            }
+            [/]
```
kper, line 102: major: as discussed in person, I think the code is a bit weird and it is hard to digest.
What do you think when we drop the variable`isFieldOfOperand` and create additional operands just for the fields?
So, you would solve this in `EmitAsmParserCppFilePass#createOperands`. If the operand is a field access function then you append the field access operand to the `result`, but also the first format field. You would need another variable in `TableGenOperand` to indicate that the operand needs to be decoded.

Unfortunately, your and mine approach are flawed because it will probably never work for encoding functions with multiple field access functions. But it's better when we have 1:1 mapping.

vadl/main/vadl/lcb/template/lib/Target/AsmParser/EmitAsmParserCppFilePass.java
```diff
@@ -166,23 +194,33 @@ private List<TableGenOperand> createOperands(TableGenInstruction instruction) {
...
       var casted = (GcbDefaultInstructionOperand) input;
       if (input instanceof TableGenInstructionImmediateOperand immediateOperand) {
+
         var operand = new TableGenOperand(immediateOperand.name(),
             indexOffset,
             true,
-            immediateOperand.immediateOperand().predicateMethod().lower()
+            immediateOperand.immediateOperand().predicateMethod().lower(),
+            true,
+            immediateOperand.formatFields().get(0).simpleName(),
```
kper, line 203: major: Regardless what the final implementation will look like, this is a silent foot gun when there is an encoding function with multiple field access functions. I would prefer it, if we throw a `DiagnosticError` when an operand has multiple format fields to spot the error easier and quicker.

vadl/main/resources/templates/lcb/llvm/lib/Target/AsmParser/AsmParser.cpp
benjaminkasper99: I refactored it in the way you suggested, but I'm not sure it's any better. As you said we still need a flag to know whether to decode, but we also need the original operand name to match it with the expected `target`.

vadl/main/vadl/lcb/template/lib/Target/AsmParser/EmitAsmParserCppFilePass.java
```diff
@@ -151,44 +158,115 @@ private List<ParseInstruction> instructions(PassResults passResults) {
...
       if (input instanceof TableGenInstructionImmediateOperand immediateOperand) {
-        var operand = new TableGenOperand(immediateOperand.name(),
-            indexOffset,
+
+        var formatFields = immediateOperand.formatFields();
+        if (formatFields.size() != 1) {
+          throw Diagnostic.error(
+              "The AsmParser cannot deal with access functions with multiple fields.",
+              input.origin().location()).build();
+        }
```
kper, line 218: hmm I have run the Aarch64 spec with your changes and now the error is thrown which is correct but will block me from continuing working :/
I expected that it will only be thrown when an assembly grammar is defined (which does not exist for Aarch64).
So, my new suggestion is to emit a warning instead and skip the field operand.
There is a something like a class for DeferredDiagnostics to avoid that the program terminates.

`./gradlew run --args="lcb --with-stacktrace vadl/test/resources/testSource/sys/aarch64/aarch64-abi.vadl"`

```
> Task :vadl-cli:run
^[[01m^[[31merror:^[[0m^[[01m The AsmParser cannot deal with access functions with multiple fields.^[[0m
     ^[[36m╭──[vadl/test/resources/testSource/sys/aarch64/aarch64.vadl:822:7]
     │
^[[0m 822 ^[[36m│^[[0m     , rightWSize =  immr as Bits5 - imms as Bits5 - 1
     ^[[36m│^[[0m       ^[[31m^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^[[0m
     ^[[36m│ ^[[0m
```

kper (approved):

PR #389, opened 2025-07-05
iss: Add TUI for cosimulation

Giftzwerg02:
This PR draft currently contains a rough design on how the TUI should look like (the data and process below are currently not connected to the cosimulation process):

TUI layout

The main two screens are horizontally split:

- On the left is a tree-view of the current data / state of the cosimulation run, the tree will be live-updated during the testing process. Below is an input field used to filter the data of the tree, it currently uses a python wrapper library of [jq](https://github.com/jqlang/jq) to filter the JSON-data. Predefined filter-options for common use-cases will probably be useful here. The tree is already lazy-loaded, meaning branches are only added to the TUI if a node is expanded. This should hopefully already make the TUI responsive enough.
- On the right are two gdb windows running as separate processes which are connected to the qemu clients. Again, predefined commands to interact with the clients (maybe even simultaneously) could definitely be useful here.

Note: The TUI uses the python library [textual](https://textual.textualize.io/).


Giftzwerg02:
I am closing this PR-Draft since the broker will be rewritten in rust. 
See new PR: #406

PR #395, opened 2025-07-08
frontend: Add powerpc spec to integration tests

kbschlosser:


flofriday (approved):
Looks good to me, and the tests insure that we won't break the syntax with compiler changes 🎉

PR #396, opened 2025-07-09
lcb: Fixed addition instruction matching with "ADDS"

kper:
The AArch64 spec crashed because it could not find an instruction with register-register addition. The cause for this was that the matchers didn't include the addition with status builtin.


benedikt-l-huber (approved):
LGTM

PR #398, opened 2025-07-10
lcb: Bugfixes Aarch64

kper:
This PR contains the following changes:

- adds the labelling of unconditional jumps in Aarch64 (machine instructions)
- it uses the machine instruction unconditional jump and if it cannot one then looks for a pseudo instruction unconditional jump. This covers both Aarch64 and RISCV.
- improves the heuristic to find the branch instruction's immediate and its bitwidth


benedikt-l-huber (approved):
LGTM

PR #401, opened 2025-07-10
cppcodegen: Support string constants in functions

benjaminkasper99:
This PR adds support for string constants in functions to `CDefaultMixins`.

String functions are needed in the assembler to match instructions whose mnemonics differ from the names defined in the ISA.
In this examle of rv32im, the pseudo instruction is defined as `LGA_32`, but has mnemonic `LGA`:
```
pseudo instruction LGA_32( rd: Index, symbol: Bits<32> ) =
{
   AUIPC { rd = rd, imm = got_pcrel_hi( symbol ) }
   LW { rd = rd, rs1 = rd, imm = pcrel_lo( symbol ) }
}
assembly LGA_32 = ("LGA", " ", register(rd), ",", hex( symbol ))
```

In the assembly grammar this case is handled by expecting `LGA` as input, but then using a parameterless function to assign the actual instruction name to the mnemonic attribute:
```
LGA_32PseudoInstruction @instruction:
   "LGA"
   mnemonic = lga_32_string @operand
   rd = Register @operand ","
   symbol = ImmediateOperand
;

function lga_32_string -> String = "LGA_32"
```

### DIFF
--- a/vadl/main/vadl/cppCodeGen/mixins/CDefaultMixins.java
+++ b/vadl/main/vadl/cppCodeGen/mixins/CDefaultMixins.java
@@ -22,12 +22,14 @@ import static vadl.utils.GraphUtils.getSingleNode;
 
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.annotation.Nullable;
 import vadl.cppCodeGen.CppTypeMap;
 import vadl.cppCodeGen.context.CGenContext;
 import vadl.cppCodeGen.context.CNodeContext;
 import vadl.javaannotations.Handler;
 import vadl.types.DataType;
 import vadl.types.StringType;
+import vadl.types.Type;
 import vadl.utils.Pair;
 import vadl.viam.Function;
 import vadl.viam.Parameter;
@@ -125,13 +127,22 @@ public interface CDefaultMixins {
      */
     default String genFunctionParameters(Parameter[] parameters) {
       var cppArgs = Stream.of(parameters)
-          .map(p -> Pair.of(p.simpleName(), requireNonNull(p.type().asDataType().fittingCppType())))
+          .map(p -> Pair.of(p.simpleName(), requireNonNull(asCppType(p.type()))))
           .toList();
 
       return cppArgs.stream().map(
           s -> CppTypeMap.getCppTypeNameByVadlType(s.right())
               + " " + s.left()
       ).collect(Collectors.joining(", "));
+    }
+
+    @Nullable
+    private Type asCppType(Type originalType) {
+      if (originalType.isDataType()) {
+        return originalType.asDataType().fittingCppType();
+      }
+
+      return originalType;
     }
 
     /**
--- a/vadl/main/vadl/lcb/codegen/assembly/AssemblyParserCodeGenerator.java
+++ b/vadl/main/vadl/lcb/codegen/assembly/AssemblyParserCodeGenerator.java
@@ -337,7 +337,8 @@ public class AssemblyParserCodeGenerator {
     element.parameters().forEach(
         parameter -> {
           ctx.gen(parameter);
-          paramVars.add(varName(parameter) + ".Value");
+          var asStdString = parameter.getAsmType() == StringAsmType.instance() ? ".str()" : "";
+          paramVars.add(varName(parameter) + ".Value" + asStdString);
         }
     );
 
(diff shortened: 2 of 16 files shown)


vadl/test/resources/testSource/lcb/asm/riscv32_functions_in_asm_grammar.vadl
```diff
@@ -0,0 +1,253 @@
...
+      }
+
+  $JLinkInstr (JAL  ; Jtype ; PC     ; opcode = 0b110'1111 ; "") // jump and link
+  $JLinkInstr (JALR ; Itype ; X(rs1) ; opcode = 0b110'0111,      // jump and link register
+                                       funct3 = 0b000 ; ("(", register(rs1), ")"))
+
+  instruction ADDMUL : Rtype =                        // 3 register operand instructions
+      X(rd) := (((X(rs1) as Bits) + (X(rs2) as Bits)) * (X(rs2) as Bits)) as Regs
+  encoding ADDMUL = { opcode = 0b011'0011, funct3 = 0b000, funct7 = 0b000'0001 }
+  assembly ADDMUL = (mnemonic, " ", register(rd), ",", register(rs1), ",", register(rs2))
```
kper, line 131: minor: please remove that

vadl/test/vadl/lcb/AsmFileCheckTest.java
```diff
@@ -25,13 +25,38 @@
...
+
+  @Override
+  protected String getSpikeTarget() {
+    return "rv32im";
+  }
+
+  @Override
+  protected String getAbi() {
+    return "ilp32";
+  }
```
kper, line 61: major: these don't belong here. They are part of an individual architecture. Please make a new sub class for riscv or move them back

vadl/test/vadl/lcb/riscv/SpikeRiscvImageProvider.java
```diff
@@ -33,6 +33,7 @@ public class SpikeRiscvImageProvider {
    * Create an {@link ImageFromDockerfile} or return an already existing image.
    *
    * @param redisCache          is the cache for building LLVM.
+   * @param imageId             is the Id for the image, to allow multiple images per target.
```
kper, line 36: minor: Maybe a more fitting name would be `cacheKey`? Since it is only used for lookup

vadl/test/resources/testSource/lcb/asm/riscv32_functions_in_asm_grammar.vadl
```diff
@@ -0,0 +1,253 @@
...
+  function one -> SInt<64> = 1
+  function firstParam(x: String, y: SInt<64>) -> String = x
+
+  function stringId(x: String) -> String = x
+  function intId(x: SInt<64>) -> SInt<64> = x
+
+  grammar = {
+    AddInstruction @instruction:
+      "ADD"
+      mnemonic = firstParam<stringId<instName>,intId<one>> @operand
```
kper, line 240: question: Isn't it possible to to integrate that into the normal RISC-V spec? While, I understand the need for this test, it is adding multiple minutes to the CI because it has to compile a new LLVM

kper (changes requested):

vadl/test/resources/testSource/lcb/asm/riscv32_functions_in_asm_grammar.vadl
benjaminkasper99: Of course it is possible to do this in the normal Riscv spec. While I do realize that this adds another compilation of LLVM to the CI, I assumed we do not want to clutter the spec with unnecessary definitions just to test some feature. Would you prefer to do this in the normal spec?

vadl/test/vadl/lcb/riscv/SpikeRiscvImageProvider.java
benjaminkasper99: I actually named it `redisKey` first, but then realized that the image name also had to be unique.
https://github.com/OpenVADL/openvadl/pull/401/files#diff-88f641de847969a4c7802c09d9f68aec3d1c3e13a491f0d5e0bce46fcde3a89aR67

vadl/test/vadl/lcb/AsmFileCheckTest.java
benjaminkasper99: You're right, I introduced a new class.

vadl/test/resources/testSource/lcb/asm/riscv32_functions_in_asm_grammar.vadl
benjaminkasper99: Alright, I added it to the normal spec.

PR #406, opened 2025-07-11
iss: Ported cosimulation-broker to rust

Giftzwerg02:
A rewrite of the cosim-broker code to further improve the performance of cosimulation runs. 

While most of the code should be a more or less direct conversion from the Python version of the broker, two other changes are also included in this PR:

1. The traces were originally dumped as json. This has been changed to inserts into a sqlite database. 
2. The project structure was split to separate the cosimulation-broker from the TUI which mainly uses the traces and test-results generated from the broker and is therefore better suited as a standalone application.

In other words, the new version should mostly be identical in terms of features compared to the old version, with the only major changes listed above and some refactoring to further improve performance.


vadl-cosim/.gitignore
```diff
@@ -1,3 +1,22 @@
...
+
+# MSVC Windows builds of rustc generate these, which store debugging information
+*.pdb
+
+# RustRover
+#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+#.idea/
```
Jozott00, line 16: minor: can be removed

vadl-cosim/src/cosim/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+
+                    if !diffs.is_empty() {
+                        return Ok(diffs);
+                    }
+                }
+            }
+        }
+
+        Ok(diffs)
+    }
```
Jozott00, line 169: Please split this up

Jozott00 (changes requested):
Looks good to me in general, we should also think about adding tests to the CI for this. Also some that test error detection. We could do this by modifying the risc-v or aarch64 vadl spec and introduce an error.

Giftzwerg02:
> Looks good to me in general, we should also think about adding tests to the CI for this. Also some that test error detection. We could do this by modifying the risc-v or aarch64 vadl spec and introduce an error. (but this can be done in a subsequent PR)

I wonder if we could use some sort of mutation testing and automatically generate incorrect (but still compileable) vadl-specs from correct specs - and then just check if the tester reports a test-failure. This could automatically cover the tester quite thoroughly.

Jozott00 (approved):

Issue #409, opened 2025-07-12
frontend: Remove `decimal` built-in alias for `sdec`


Jozott00:
Once all VADL specifications were migrated to `sdec`, we must remove the current aliasing of `decimal` to `sdec`.

PR #410, opened 2025-07-12
frontend: Add udec/sdec and remove decimal built-in

### LINKED ISSUE #408: Assembly builtins: signed and unsigned decimal printing.
### Discussed in https://github.com/OpenVADL/openvadl/discussions/393

<div type='discussions-op-text'>

<sup>Originally posted by **AndreasKrall** July  8, 2025</sup>
Currently we only support a signed assembly printing function called `decimal`. We must support both the printing of signed and unsigned values. Currently the only way to support unsigned printing is to cast the printed value to an unsigned which is at least one bit bigger than the current value. I propose to support two different versions of `decimal` for signed and unsigned printing. We need to find suitable names. Suggestions are (please add further suggestions):
```
Sdecimal/Udecimal
decimalS/decimalU
sint/uint
sdec/udec
```
My preference is `sdec/udec`.
`hex` and `bin` will always interpreted as unsigned values.
Currently there are known errors in the RISC-V and AArch64 spec which have to be fixed after the new builtin versions are implemented.</div>

Jozott00:
This PR closes #408.

It adds the `udec` and `sdec` built-ins. The old `decimal` built-in is removed but the name can stil be used as an alias for `sdec`. 

Once all specification were adapted, to used `udec`/`sdec`, we must also remove this alias (#409).

### DIFF
--- a/vadl/main/vadl/ast/AstUtils.java
+++ b/vadl/main/vadl/ast/AstUtils.java
@@ -45,6 +45,10 @@ class AstUtils {
     if (pseudoRewrites.containsKey(name)) {
       var singed = argTypes.stream().anyMatch(t -> t instanceof SIntType);
       name = pseudoRewrites.get(name).get(singed ? 0 : 1);
+    } else if (name.equals("decimal")) {
+      // TODO: Remove this once all decimal calls were replaced by s/udec in the VADL specs (#409)
+      // set decimal to be an alias of sdec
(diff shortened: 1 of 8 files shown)


Jozott00:
@kper I naively replaced all usages of `DECIMAL` in the ASM/LCB generator by `SDEC`. `UDEC` is not yet integrated and will cause a crash when used.

PR #442, opened 2025-07-27
iss: Added before->instruction->after state-transition to diff-context of the cosimulation-report

Giftzwerg02:
This PR more data to the cosimulation-report and also now prints the report in a json-format.

Specifically, the report now additionally contains:
- a "before" state of all the registers before a diff occurred
- the instruction that was executed which caused the diverging state
- the "after" state of all the registers when the diff occurred
for each client.

This feature initially slowed the cosimulator by a large amount - with the issue being that the whole before-state needed to be cloned / pre-computed before every iteration "just in case" a divergence occurred. This resulted in a lot of unnecessary cloning / data-transformations.

The final solution can be seen in the last commit (e5fd5c207bce832fcba7b9e49ebff9a85ee5753f): I introduce a second shared-memory object for each client and switch between them, meaning the "before"-state can now be read from the unused shm-object of the given iteration.

### DIFF
--- a/vadl-cosim/Cargo.lock
+++ b/vadl-cosim/Cargo.lock
@@ -644,9 +644,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.140"
+version = "1.0.141"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "20068b6e96dc6c9bd23e01df8827e6c7e1f2fddd43c21810382803c136b99373"
+checksum = "30b9eff21ebe718216c6ec64e1d9ac57087aad11efc64e32002bce4a0d4c03d3"
 dependencies = [
  "itoa",
  "memchr",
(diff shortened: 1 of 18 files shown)


Jozott00:
I think we should only print the report only if a certain flag is set.

Giftzwerg02:
> I think we should only print the report only if a certain flag is set.

What should happen if the flag is not set? Only set an error exit code?

Giftzwerg02:
> Probably a short version of it, but not as json. So a readable version of the differences, PC, instruction. In this case you can omit information about the rest of the CPU state, just the information that differs between the simulations.

I added `testing.protocol.out.verbosity` to the config:
`"short" `-> human readable result summary
`"full"` -> the full json output

The summary currently looks something like the following:
```
Cosimulation failed!
Failure at pc = 0x40002618 (1073751576)

The following divergences were found:
- "different register data for pc":
	- In "VADL" the value is "0x0C26004000000000"
	- In "UPSTREAM" the value is "0x1826004000000000"

The divergence occurred after the following instructions were executed:
- (pc=1073751576): sxtw x0, w2 (0x407C4093)
```

Jozott00 (approved):

PR #448, opened 2025-08-06
iss: Added TUI to view cosim-traces

Giftzwerg02:
This PR adds an initial implementation / prototype for the cosim-tui which displays the tracing-information that is collected during a cosim-run.

A small demo of the current state can be seen below:

https://github.com/user-attachments/assets/99108084-4c61-44f9-a417-bd683c144b99


Jozott00 (approved):

Jozott00:
Looks nice! What exactly are the shown instructions? are the the last executed ones?

Do I open the TUI after the cosimulation terminated or while it is running?

PR #493, opened 2025-09-01
lcb: Add annotation to overwrite operands

kper:
The problem is that we deduced the instruction's operands based on the behavior so far. That's ok for many instructions, however, there are a few instructions which are an edge case.

For example

```
model ExceptionCallEncAsm (i: InstrWithFunct): IsaDefs = {
    encoding $i.id = {op = $i.opcode}
    assembly $i.id = ($i.mnemo, ' #', decimal(imm16))
    }

model ExceptionCallInstr (i: InstrWithFunct): IsaDefs = {
    instruction $i.id: SuperVisorCallFormat = {
      raise GeneralException (ExceptionClass::$i.funct, 0, PC.next, 0)
      }
    $ExceptionCallEncAsm ($i)
    }

  $ExceptionCallInstr    ((BRK    ; "brk"   ; 0b1101'0100'0010'0000   ; SoftwareBreakpoint ))
```

The instruction uses a field `imm16` in the `assembly` which is never used in the behavior which means that no operand has been created. We fixed that by introducing a new annotation to overwrite the input operands.
With `[add operands : imm16]` a new pass adds the field `imm16` as side effect, which creates an operand.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -77,6 +77,7 @@ with their licence information:
 |        MIT License         | flowbite - tailwind component library                    | https://github.com/themesberg/flowbite                 |
 |     Apache License 2.0     | Error Prone - compile time analysis tool                 | https://github.com/google/error-prone                  |
 |        MIT License         | NullAway - static null checking for java                 | https://github.com/uber/NullAway                       |
+|        MIT License         | Z3 - a theorem prover from Microsoft Research            | https://github.com/Z3Prover/z3                         |
 
 ### Test Libraries
 
--- a/vadl-cli/build.gradle.kts
+++ b/vadl-cli/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     application
+    id("io.github.rascmatt.z3") version "1.0.2"
     id("org.graalvm.buildtools.native") version "0.10.2"
 }
 
--- a/vadl/build.gradle.kts
+++ b/vadl/build.gradle.kts
@@ -19,6 +19,7 @@ import java.util.*
 
 plugins {
     id("vadl.CocoR")
+    id("io.github.rascmatt.z3") version "1.0.2"
 }
 
 
@@ -36,6 +37,7 @@ dependencies {
     implementation("org.apache.commons:commons-lang3:3.18.0")
     implementation("org.apache.commons:commons-text:1.10.0")
 
+    implementation("io.github.rascmatt:z3-bootstrap:1.0.0")
 
     testImplementation(platform("org.junit:junit-bom:5.9.1"))
     testImplementation("org.junit.jupiter:junit-jupiter")
--- a/vadl/main/vadl/configuration/DecoderOptions.java
+++ b/vadl/main/vadl/configuration/DecoderOptions.java
@@ -73,7 +73,10 @@ public class DecoderOptions {
   public enum OptionToSkip {
 
     OPT_CONSTRAINT_SYNTHESIS("Skip constraint synthesis for decoder generation, default: enabled",
-        "constraint-synthesis");
+        "constraint-synthesis"),
+
+    OPT_ENCODING_VERIFICATION("Skip the encoding verification, default: enabled",
+        "encoding-verification");
 
     private final String selector;
     private final String desc;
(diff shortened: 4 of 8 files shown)


vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -240,6 +245,25 @@ public class AnnotationTable {
             def.addAnnotation(new OnlyNegativeNumbersAnnotation());
           }
         }).build();
+
+    annotationOn(InstructionDefinition.class, "overwrite operand", DefinitionRefAnnotation::new)
```
Jozott00, line 249: You are using the `DefinitionRefAnnotation` in a wrong way.
The `DefinitionRefAnnotation` expects exactly one identifier that references some definition node. Also, the resulting value is the `.def` field.

Maybe you should refactor the `DefinitionRefAnnotation` class in a way, such that instead of having only a single resolved `def` field, it has a list `defs` of resolved definitions. With that you also don't have to match names manually, but instead already get a list of definitions (you only have to check that those definitions are `Format.Field`s).

vadl/main/vadl/viam/graph/dependency/DefinedImmediateSideEffectNode.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.annotation.Nullable;
+import vadl.javaannotations.viam.Input;
+import vadl.viam.graph.GraphNodeVisitor;
+import vadl.viam.graph.Node;
+
+/**
+ * Node which indicates that a field or field access function is used. This is required for the LCB
+ * since not all operands of an instruction must be part of a behavior.
+ */
+public class DefinedImmediateSideEffectNode extends SideEffectNode {
```
Jozott00, line 29: Why is this a general VIAM node and not just a LCB node?

vadl/main/vadl/viam/passes/OverwriteInputOperandsPass.java
```diff
@@ -0,0 +1,62 @@
...
+import vadl.viam.Specification;
+import vadl.viam.annotations.DefineOperandAnnotation;
+import vadl.viam.graph.dependency.DefinedImmediateSideEffectNode;
+import vadl.viam.graph.dependency.FieldRefNode;
+
+/**
+ * Reads all the {@link DefineOperandAnnotation} and adds them as
+ * {@link DefinedImmediateSideEffectNode}.
+ */
+public class OverwriteInputOperandsPass extends Pass {
```
Jozott00, line 34: Can this pass be moved to the LCB? As far as I can see, this is not influencing any generic VIAM optimization passes.

vadl/main/vadl/viam/passes/OverwriteInputOperandsPass.java
```diff
@@ -0,0 +1,62 @@
...
+    var machineInstructions = viam.isa().stream()
+        .flatMap(isa -> isa.ownInstructions().stream())
+        .filter(x -> x.hasAnnotation(DefineOperandAnnotation.class));
+
+    machineInstructions.forEach(machineInstruction -> {
+      var annotation = machineInstruction.expectAnnotation(DefineOperandAnnotation.class);
+
+      for (var field : annotation.inputs()) {
+        machineInstruction.behavior().addWithInputs(
+            new DefinedImmediateSideEffectNode(null, new FieldRefNode(field, field.type())));
```
Jozott00, line 56: The side effect is not attached to the `InstrEndNode`, but just floats around, resulting in an invalid graph.

Jozott00 (changes requested):

vadl/main/vadl/viam/passes/OverwriteInputOperandsPass.java
kper: That's ok. I claim that this is for nodes which are unlowerable. So at the moment, I don't care about it. The important thing to me is that the operand exists, so the assembly printing works and the tablegen record exists for the assembler. I don't expect to add it to the instruction selection.

vadl/main/vadl/viam/passes/OverwriteInputOperandsPass.java
Jozott00: Hmm, you could schedule the pass using `addAfterFirst`. So instead of skipping it in all other pass orders, you just add it in the LCB. Also like no other VIAM pass is touching the node right? So there would be architecture test failure.

vadl/main/vadl/viam/graph/dependency/DefinedImmediateSideEffectNode.java
kper: moved to lcb

PR #497, opened 2025-09-02
asm: Rename asm builtins

benjaminkasper99:
This PR renames the assembly builtin functions `laideq` and `laidin` to `LaIdEq` and `LaIdIn`.

@AndreasKrall you expressed concern regarding the possible confusion with lower case `l` and upper case `I`.
Code editor fonts usually distinguish these letters, so I think there shouldn’t be an issue.


AndreasKrall (approved):
Is fine

Issue #512, opened 2025-09-19
frontend: Better alias register slice semantics


Jozott00:
The current implementation of alias register slices as described in #441 was implemented in #445 and #503. 

However, its current behavior is not intuitive and conflicts with the behavior of other slice (writes) in VADL.
Given an alias register like
```
register S: Bits<64>
alias register W = S(31..0)
```
a write to the alias
```
W := 3 as Bits<32>
```
will currently override the whole register `S` with `3` (zero extended).

### Intuitive Behavior

However, intuitively the write should only effect the sub range of the source register given by the slice definition of the alias. So that it is the same as writing
```
S(31..0) := 3 as Bits<32>
```

### Tweaking the Behavior

In many cases the intuitive behavior is not what we want. Instead we want a complete overwrite of the source register. This can be done by either zero extending or sign extending the written value. 

As both options have their use cases, the user has to specify what override mode to choose.
```
register S: Bits<64>

[ overwrite source : zero extending ]
alias register W = S(31..0)

[ overwrite source : sign extending ]
alias register U = S(31..0)
```
So writing 
```
W := 3 as Bits<32>
U := 3 as Bits<32>
```
Is the same as writing
```
S := 3 as Bits<64>
U := 3 as SInt<32> as Bits<64>
```

*Note:* The `[overwrite source :]` annotation can only be used if the slice is a continues slice that starts at `0`.

PR #519, opened 2025-09-29
frontend: New alias register slicing behavior

Jozott00:
This PR changes the semantics of the alias register slicing based on the new `[overwrite source : zero | sign ]` annotation.

Closes: #512


sys/aarch64/aarch64.vadl
```diff
@@ -64,6 +64,7 @@ instruction set architecture AArch64Base = {
   [zero : X(31)]                      // X(31) is the zero register
   alias register  X = S               // general purpose register file with zero register
   [zero : W(31)]                      // X(31) is the zero register
+  [overwrite source: zero]            // S is overwritten on write by zero extending the written value
```
kper, line 67: I thought @AndreasKrall wanted `zero extend on write`

PR #529, opened 2025-10-06
viam: Add dynamic slice node

Jozott00:
The PR adds the `DynSliceNode`, which is required in for loops as the loop index is a dependency expression and not constant. This is the only case where this node is produced.

Here is an example where the index is used to slice a value:
```
register X : Bits<32>
register Y : Bits<32>
// ...

forall i in 0 .. 1 do
      X(i) := Y(i)
```


flofriday (approved):
LGTM ✨

PR #545, opened 2025-10-25
lcb: Add skip-pattern-generation option to the lcb command

benjaminkasper99:
This PR adds `skip-pattern-generation` as an option to the OpenVADL LCB command. Enabling it skips all pattern generation in th  LLVM lowering in the LCB.

~~This PR adds an `--assembler` (shorthand `-a`) option to the OpenVADL LCB command.~~

~~For now this option just disables TableGen pattern generation. In the future it can be extended, for example to disable passes that are not needed for assembler generation.~~

### DIFF
--- a/vadl-cli/main/vadl/cli/LcbCommand.java
+++ b/vadl-cli/main/vadl/cli/LcbCommand.java
@@ -30,7 +30,7 @@ import vadl.pass.PassOrder;
 import vadl.pass.PassOrders;
 
 /**
- * The Command does provide the iss subcommand.
+ * The Command does provide the lcb subcommand.
  */
 @Command(
     name = "lcb",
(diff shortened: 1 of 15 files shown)


vadl-cli/main/vadl/cli/LcbCommand.java
```diff
@@ -44,9 +44,13 @@ public class LcbCommand extends BaseCommand {
...
   String targetName;
 
+  @Option(names = {"-a", "--assembler"}, scope = INHERIT,
+      description = "Only generate the LCB assembler artifacts")
+  boolean onlyAssembler;
+
   @Override
   PassOrder passOrder(GeneralConfiguration configuration) throws IOException {
-    var lcbConfig = new LcbConfiguration(configuration, targetName());
+    var lcbConfig = new LcbConfiguration(configuration, targetName(), onlyAssembler);
```
kper: major: what's the default here?

vadl/main/vadl/configuration/LcbConfiguration.java
```diff
@@ -24,14 +24,31 @@
  */
 public class LcbConfiguration extends GcbConfiguration {
 
+  // Only generate artifacts needed for the LCB assembler
+  private boolean onlyAssembler;
```
kper: same as above

kper (changes requested):

PR #546, opened 2025-10-29
lcb: Fixed calling conventions for multiple register files

kper:
The problem was that the caller and callee saved registers were added to the allocation sequence for all register class. This is of course incorrect because it is allowed to have one calling convention but multiple register files.


benedikt-l-huber (approved):
LGTM :+1:

PR #550, opened 2025-11-01
asm: Add assembler specification for VarRisc

benjaminkasper99:
This PR adds an assembler specification for the VarRisc architecture and tests it with LLVMs FileCheck tool.

Notes:
- This assembler is for the VarRisc version where instructions have suffixes according to their length (e.g., `_L`).
- Also contained in this PR are fixes for various bugs that surfaced while trying to generate an assembler.
- For testing I moved `VarRisc.vadl` to the `testSource` directory and created a symbolic link at its original location in `sys/v-risc`. This mimics the approach currently used with RiscV.


vadl/test/vadl/lcb/riscv/riscv32/asm/AsmParserRiscv32FileCheckTest.java
```diff
@@ -16,41 +16,12 @@
 
 package vadl.lcb.riscv.riscv32.asm;
 
-import vadl.lcb.AsmFileCheckTest;
+import vadl.lcb.riscv.riscv32.AsmRiscv32FileCheckTest;
 
-public class AsmParserRiscv32FileCheckTest extends AsmFileCheckTest {
```
kper: Why did u touch this?

vadl/test/vadl/lcb/riscv/riscv32/asm/AsmCodeEmitterRiscv32FileCheckTest.java
```diff
@@ -16,41 +16,12 @@
 
 package vadl.lcb.riscv.riscv32.asm;
 
-import vadl.lcb.AsmFileCheckTest;
+import vadl.lcb.riscv.riscv32.AsmRiscv32FileCheckTest;
 
-public class AsmCodeEmitterRiscv32FileCheckTest extends AsmFileCheckTest {
```
kper: Why did u touch this?

vadl/test/vadl/lcb/riscv/riscv32/asm/AsmParserRiscv32FileCheckTest.java
benjaminkasper99: While creating the tests for VarRisc I first wanted to extend the previous `AsmFileCheckTest`. But then I noticed that it contained the hard-coded path to `riscv32im.vadl`. Because of this I renamed it to `AsmRiscv32FileCheckTest` and moved it into the Riscv32 subdirectory. Then I also moved the `@override` methods into this class, since now it is clear that this class is only for Riscv32 tests anyways.

vadl/test/vadl/lcb/riscv/riscv32/asm/AsmCodeEmitterRiscv32FileCheckTest.java
benjaminkasper99: see above

PR #559, opened 2025-11-07
typechecker: Improve fault tolerance

flofriday:
Previously the typechecker reported the first error it encountered and exited. Now it can figure out how to recover in many cases. These changes are mostly important for the LSP.

**Changes in the API**
- The typechecker no longer throws a `Diagnostic` but always a `DiagnosticList` even if it just contains a single diagnostic.
- The typechecker will try to recover to a recovery point once a typeerror was found and continue from there. This is so that it can find more typeerrors and report as many as possible to the user.

**Three ways to report errors**
There are now three errors to report errors in the typechecker.
1. `throw error(...)` this was the common approach before and will interrupt the checking process. It is now only used for fatal errors where the type-checker is unrecoverable confused.
2. `addErrorAndStopPartialChecking(error(...))` and `throw new StopPartialCheckingSignal()` this interrupts the checking until a recovery point (the next statement of next definition.
3. `errors.add(error(...))` this is for errors where we can continue directly because the result of the check isn't important for future checks. A good example for this one is the check if a the condition of an if is a bool because if not we can simply report the error and continue as usual.

### DIFF
--- a/vadl/main/vadl/ast/TypeChecker.java
+++ b/vadl/main/vadl/ast/TypeChecker.java
@@ -17,7 +17,6 @@
 package vadl.ast;
 
 import static java.util.Objects.requireNonNull;
-import static vadl.error.Diagnostic.ensure;
 import static vadl.error.Diagnostic.error;
 import static vadl.error.Diagnostic.warning;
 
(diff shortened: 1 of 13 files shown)


vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -85,7 +84,21 @@ enum BranchStrategy {
...
 
+  /**
+   * A custom exception to be thrown to partially stop evaluating to recover to a point where it
+   * makes sense to resume typechecking.
+   */
+  static class StopPartialCheckingSignal extends RuntimeException {
+  }
+
+  /**
+   * Recoverable errrors are stored in this list.
```
Jozott00: Typo `errrors`

flofriday: thanks :)

Jozott00 (approved):

PR #569, opened 2025-11-17
frontend: Add expansion alias lowering

flofriday:
A expansion alias is something like:
register R: Bits<4><4>
alisas register A: Bits<4><2><2>

While compression aliases are easier to implement, their implementation
has to wait since expansions are needed now.

Implementing them required a lot of complex bit manipulation, resulting
in verbose VIAM graphs.

## Example
```vadl
instruction set architecture ARCH = {
  register CR  : Bits<32>
  alias register CRF : Bits<8><4> = CR
}
```

Read Graph:
Screenshot 2025-11-17 at 15-48-02 Specification (tmp) - at VIAM Creation

Write Graph:
Screenshot 2025-11-17 at 15-47-39 Specification (tmp) - at VIAM Creation

<details>
 <summary>Understanding the bit manipulation</summary>
 The bit fideling is somewhat complex, so to better help myself to understand it I wrote some python functions to playaround with values:
 
 ```python
 def mask(msb, lsb):
    return bin(((1<<(msb-lsb+1))-1) << lsb)

# i: index of tensor
# p: part sice of each tensor element
# l: total length of the tensor in bits    
def msblsb(i, p, l):
    left = (l-1)-i*p
    right = left - (p-1)
    return (left, right)    
 ```
</details>

### DIFF
--- a/vadl/main/vadl/ast/TypeChecker.java
+++ b/vadl/main/vadl/ast/TypeChecker.java
@@ -1076,19 +1076,21 @@ public class TypeChecker
       return false;
     }
 
-    var fromIndecies = new ArrayList<>(switch (from) {
-      case TensorType t ->
-          Stream.concat(t.indexDims().stream(), List.of(t.innerType().bitWidth()).stream())
-              .toList();
-      default -> List.of(from.bitWidth());
-    });
+    var fromIndecies = new ArrayList<>(
+        switch (from) {
+          case TensorType t ->
+              Stream.concat(t.indexDims().stream(), List.of(t.innerType().bitWidth()).stream())
+                  .toList();
+          default -> List.of(from.bitWidth());
+        });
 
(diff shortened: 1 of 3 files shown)


vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -1076,19 +1076,21 @@ private boolean canAliasCastType(DataType from, DataType to) {
...
-      default -> List.of(from.bitWidth());
-    };
-
-    var toIndecies = switch (to) {
-      case TensorType t ->
-          Stream.concat(t.indexDims().stream(), List.of(t.innerType().bitWidth()).stream())
-              .toList();
-      default -> List.of(to.bitWidth());
-    };
+    var fromIndecies = new ArrayList<>(
```
Jozott00: Typo: Indices

PR #570, opened 2025-11-17
typechecker: Improve error forall cast required

flofriday:
The error message now suggests _how_ to add a typecast to a forall index, as the syntax isn't trivial.


kbschlosser (approved):
👍

PR #584, opened 2025-11-30
lowering: Add dynamic slicing and tensor reading with missing indices

flofriday:
Previously reading tensors without specifying all required indices resulted in an error.

For example:
```
register X: Bist<4><4><8>

... Y := X(1) // This should work and result in 4 underlying register
              // reads
```


Jozott00 (approved):
lgtm

PR #585, opened 2025-11-30
rtl: Add RTL test base image

rascmatt:
- refactor firtool to only export relevant files
- implement dockerfile for rtl test base

The integration of analysis tools Yosys + OpenSTA is still work in progress, but I'd like get this into master so we can build the docker artifacts.
Usage of the base image in the CI Pipeline is already prepared and tested with a local image build. See PR #583


docker/rtl/test-base/Dockerfile
```diff
@@ -0,0 +1,94 @@
+FROM ghcr.io/openvadl/merged-toolchain@sha256:2ceaf5369996a5f3db0e0bb8e16ae5fe53003d6ac180bd9e60a9ced954f99a9a AS toolchains
+FROM ghcr.io/openvadl/firtool AS firtool
```
Jozott00: This should also include a sha once you stabilize the image.

PR #596, opened 2025-12-07
chore: Add snapshot tests for typechecker

flofriday:
For now this only includes the typechecker because I wanted to get some feedback here.

**Question for the reviewers:**

1) Are you happy with the snapshot layout, I deviated a bit from the issue but I think it's still quite clear what's happending in the tests.

2) Is the updating process clear and simple enough with: `UPDATE_SNAPSHOTS=true ./gradlew test --tests vadl.ast.DiagnosticsTest`?

I think these tests already found some regressions in the diagnostic printing because some diagnostics don't look as good as I remember them, and I'm quite happy we finally have tests that show that. Though this PR doesn't make any attempts to fix them and they will be improved in future releases.

### DIFF
--- a/vadl/main/vadl/error/DiagnosticPrinter.java
+++ b/vadl/main/vadl/error/DiagnosticPrinter.java
@@ -34,6 +34,7 @@ import vadl.utils.SourceLocation;
  */
 public class DiagnosticPrinter {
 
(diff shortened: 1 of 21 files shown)


Jozott00:
Sorry for the late check, I think it's a nice approach!

Issue #618, opened 2025-12-22
rtl: Reenable RiscVInstructionTest


flofriday:
After adding new VIAM nodes in #373, the RTL implementation no longer worked. After talking with @linushdot we decided to temporary disable the test.

PR #593, opened 2025-12-03
rtl: Automated RTL Benchmarking

rascmatt:

PR #622, opened 2025-12-27
rtl: Reenable rv32im instruction tests #618

rascmatt:
Keep the dummy-mia overrides, so we can for now keep the RTL integration tests enabled. I'd like to also continue/implement the automated RTL benchmarking, which is based on the same mechanisms.

Unfortunately this requires a bit of a hack in order to 'override' the MIA within the VIAM, for now I opted to take the last MIA definition element, rather than the first.

### DIFF
--- a/vadl/test/vadl/rtl/RtlLoweringTest.java
+++ b/vadl/test/vadl/rtl/RtlLoweringTest.java
@@ -56,6 +56,7 @@ public class RtlLoweringTest extends AbstractTest {
     var generalConfig =
         new GeneralConfiguration(Path.of("build/test-output"), true);
     var config = new RtlConfiguration(generalConfig);
+    config.setDummyMia(RtlConfiguration.DummyMia.five);
 
     var decoderOptions = new DecoderOptions();
     decoderOptions.setGenerator(DecoderOptions.Generator.REGULAR);


rascmatt:
@flofriday / @linushdot Are any of the actual rv64-mia specs supposed to work already? Or is there a 32 bit version in the works?
(i don't want to pressure anyone, I'd like to genuinely just know)

rascmatt:
@flofriday Thanks! 
I don't think we're using logic elements, at least I don't see any in the `rv64imFive.vadl` spec. Or am I missing something? 

Fyi, I tried the RTL lowering with the `rv64imFive.vadl` spec, and for now only the exception handling seems to cause a validation error. This is fine I guess, since afaik exception handling is also not implemented. Or does the frontend support it? Without these two lines, it also fails during the behaviour lowering. 

Anyways, I don't need it for now, since this PR enables support for the dummy MIA again. So I would say we'll leave it like this for now.

Thank you, pls approve once the pipeline succeeds :)

rascmatt:
@linushdot 
Ok cool, thats good to hear. Should I just wait then? I would need the RTL generation to work reliably for #593, but I guess the PR can wait another week. 
Then again, we just disabled the tests originally (with #618), so I don't think it would be wrong to re-enable them, using the dummy mia until the real ones do work.

linushdot:
Automated testing will need to change as well, because selecting the MiA moves from the dummy command argument to selecting a VADL spec for the MiA. So waiting for the fix of the generator could be easier, but I don't see a problem with either way.

PR #631, opened 2025-12-31
docs: Add ATG reference manual documentation

Jozott00:
This PR adds reference manual documentation to the ATG such that it can be used to generate a Doxygen reference manual.


flofriday (approved):
Read over the changes and the ATG looks much cleaner now with the comments 💪

PR #656, opened 2026-01-06
iss: Added simple check for consecutive memory access calls

Giftzwerg02:
Instructions like `stp` in aarch64 differ in terms of calls to memory between upstream (which only makes one large memory access) and our implementation (which makes two memory accesses, one for each register). This tweak allows handling these cases, i.e. tests will still pass if the accessed memory is the same.


Jozott00 (approved):
LGTM!

Issue #658, opened 2026-01-07
frontend: crash when invoking a macro from an imported ISA


AndreasKrall:
The following test program:
```
import aarch64::AArch64Base

instruction set architecture AArch64SVEandSME extending AArch64Base = {

model Test () : Stat = {
  X(3) := $WXReg(WSize)(4)
}

}
```
results in the following crash
```
                             ___ ___    _   ___ _  _
                            / __| _ \  /_\ / __| || |
                           | (__|   / / _ \\__ \ __ |
                            \___|_|_\/_/ \_\___/_||_|

                       🔥 The OpenVADL compiler crashed 🔥

This shouldn't have happened, please open an issue with the stacktrace below at:
https://github.com/OpenVADL/open-vadl/issues/new


java.lang.IllegalArgumentException: Cannot join source locations that point to different files.
	at vadl.utils.SourceLocation.join(SourceLocation.java:114)
	at vadl.ast.Parser.callOrBinaryExpression(Parser.java:1471)
	at vadl.ast.Parser.term(Parser.java:2772)
	at vadl.ast.Parser.expression(Parser.java:938)
	at vadl.ast.Parser.assignmentOrCallStmt(Parser.java:2648)
	at vadl.ast.Parser.statement(Parser.java:1331)
	at vadl.ast.Parser.macroBody(Parser.java:3700)
	at vadl.ast.Parser.macroDefinition(Parser.java:413)
	at vadl.ast.Parser.isaDefinitionList(Parser.java:862)
	at vadl.ast.Parser.instructionsetDefinition(Parser.java:459)
	at vadl.ast.Parser.topLevelDefinition(Parser.java:354)
	at vadl.ast.Parser.vadl(Parser.java:326)
	at vadl.ast.Parser.Parse(Parser.java:3809)
	at vadl.ast.VadlParser.parse(VadlParser.java:115)
	at vadl.ast.VadlParser.parse(VadlParser.java:62)
	at vadl.cli.BaseCommand.parseToAst(BaseCommand.java:167)
	at vadl.cli.BaseCommand.parseToVIAM(BaseCommand.java:248)
	at vadl.cli.BaseCommand.call(BaseCommand.java:329)
	at vadl.cli.BaseCommand.call(BaseCommand.java:66)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2045)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
[...]

Issue #643, opened 2026-01-03
frontend:  Cannot join source locations that point to different files.


AndreasKrall:
When I have an erroneous specification which imports from a file, like:
```
import aarch64::AArch64Base

instruction set architecture AArch64SVEandSME extending AArch64Base = {
some error
}
```
VADL crashes with the following stack trace:
```
                             ___ ___    _   ___ _  _
                            / __| _ \  /_\ / __| || |
                           | (__|   / / _ \\__ \ __ |
                            \___|_|_\/_/ \_\___/_||_|

                       🔥 The OpenVADL compiler crashed 🔥

This shouldn't have happened, please open an issue with the stacktrace below at:
https://github.com/OpenVADL/open-vadl/issues/new


java.lang.IllegalArgumentException: Cannot join source locations that point to different files.
	at vadl.utils.SourceLocation.join(SourceLocation.java:114)
	at vadl.ast.Parser.callOrBinaryExpression(Parser.java:1472)
	at vadl.ast.Parser.term(Parser.java:2848)
	at vadl.ast.Parser.expression(Parser.java:939)
	at vadl.ast.Parser.argument(Parser.java:3065)
	at vadl.ast.Parser.arguments(Parser.java:3042)
	at vadl.ast.Parser.callOrBinaryExpression(Parser.java:1457)
	at vadl.ast.Parser.term(Parser.java:2848)
	at vadl.ast.Parser.expression(Parser.java:939)
	at vadl.ast.Parser.letStatement(Parser.java:2606)
	at vadl.ast.Parser.statement(Parser.java:1319)
	at vadl.ast.Parser.instructionDefinition(Parser.java:1016)
	at vadl.ast.Parser.isaDefinition(Parser.java:887)
	at vadl.ast.Parser.isaDefinitionList(Parser.java:860)
	at vadl.ast.Parser.macroBody(Parser.java:3781)
	at vadl.ast.Parser.macroDef(Parser.java:414)
	at vadl.ast.Parser.isaDefinitionList(Parser.java:863)
	at vadl.ast.Parser.instructionsetDefinition(Parser.java:460)
	at vadl.ast.Parser.topLevelDefinitions(Parser.java:355)
	at vadl.ast.Parser.vadl(Parser.java:327)
	at vadl.ast.Parser.Parse(Parser.java:3885)
	at vadl.ast.VadlParser.parse(VadlParser.java:115)
	at vadl.ast.VadlParser.parse(VadlParser.java:62)
[...]

PR #661, opened 2026-01-07
utils: Fix SourceLocation joining from model invocation

flofriday:
Closes: #643 
Closes: #658


Jozott00 (dismissed):
I would also add a test where the error is caused in the macro body.

```
model apfelstrudel(): Ex = {
  false + 42
}

constant first = true

constant second = $apfelstrudel
```

This should mark the invocation `$apfelstrudel` and also the expression `false + 42`.

Jozott00:
In this case, it's good from my side, just the wrong locations as mentioned by Andi.

AndreasKrall (approved):
looks fine

Issue #607, opened 2025-12-12
cosim: Crash with attempt to subtract with overflow


Jozott00:
When trying to run the cosim with my spec I get

```
thread 'main' (47182) panicked at cosim-lib/src/ipc/cstructs.rs:559:33:
attempt to subtract with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/panicking.rs:80:14
   2: core::panicking::panic_const::panic_const_sub_overflow
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/panicking.rs:180:17
   3: cosim_lib::ipc::cstructs::BrokerSHMRingBuffer<_>::read_previous
             at /work/code/open-vadl/vadl-cosim/cosim-lib/src/ipc/cstructs.rs:559:33
   4: cosim_lib::ipc::shm::SharedMemory<cosim_lib::ipc::cstructs::BrokerSHMRingBuffer<_>>::read_buffer_prev
             at /work/code/open-vadl/vadl-cosim/cosim-lib/src/ipc/shm.rs:31:20
   5: cosim_lib::cosim::Broker::check_clients_are_initially_synchronized::{{closure}}
             at /work/code/open-vadl/vadl-cosim/cosim-lib/src/cosim/mod.rs:411:27
   6: core::iter::adapters::map::map_fold::{{closure}}
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/iter/adapters/map.rs:88:28
   7: <core::slice::iter::Iter<T> as core::iter::traits::iterator::Iterator>::fold
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/slice/iter/macros.rs:255:27
   8: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
             at /rustc/ded5c06cf21d2b93bffd5d884aa6e96934ee4234/library/core/src/iter/adapters/map.rs:128:19
   9: core::iter::traits::iterator::Iterator::for_each
[...]

PR #666, opened 2026-01-09
iss: Parallelize ppc64/main.py testing script

Giftzwerg02:
After some further testing, I believe that the reason for the seemingly random test-failures when running the tests in parallel, was due to `asyncio` spawning too many instances at once - causing and out-of-memory issue.

Simply limiting the number of concurrent cosim-spawns not only solves this issue but also seems to get the best performance when setting this number to the number of cores of the machine.

This PR does exactly that, also swapping `asyncio` with the `ProcessPoolExecutor` which gave better results w.r.t. performance.

### DIFF
--- a/vadl-cosim/cosim-lib/src/ipc/cstructs.rs
+++ b/vadl-cosim/cosim-lib/src/ipc/cstructs.rs
@@ -557,7 +557,7 @@ impl<const SIZE: usize> BrokerSHMRingBuffer<SIZE> {
     // Once end_read is called, the reference is no longer valid, therefore the value needs to
     // already be used or cloned once `end_read` is called.
     pub const fn read_previous(&self) -> &BrokerSHMData {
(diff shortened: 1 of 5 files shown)


vadl-cosim/cosim-lib/src/ipc/cstructs.rs
```diff
@@ -557,7 +557,7 @@ impl<const SIZE: usize> BrokerSHMRingBuffer<SIZE> {
     // Once end_read is called, the reference is no longer valid, therefore the value needs to
     // already be used or cloned once `end_read` is called.
     pub const fn read_previous(&self) -> &BrokerSHMData {
-        let idx = self.ring_idx(self.read_idx - 1);
+        let idx = self.ring_idx(self.read_idx.wrapping_sub(1));
```
Jozott00: Does this close #607 ?

PR #669, opened 2026-01-11
setup: Add jlink build for the lsp

flofriday:
I'm really unsure about this implementation, I think it works but this is mostly inspired from some obscure websites and my confidence isn't that high here.

@Jozott00 can you read over it (and verify that it at least doesn't mess with anything else) and try it out on your machine with the JAVA_HOME not set to Java25?

### DIFF
--- a/vadl/build.gradle.kts
+++ b/vadl/build.gradle.kts
@@ -41,6 +41,7 @@ dependencies {
 
     implementation("io.github.rascmatt:z3-bootstrap:1.0.0")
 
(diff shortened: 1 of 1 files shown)


Jozott00:
While it is running without crashes (although I did not connect to it using a plugin), the logger is missing, so nothing is printed. 

So we should also bundle a logging provider.

Jozott00:
I added the logback provider module in 7d5bdf7.

PR #678, opened 2026-01-14
setup: Merge LSP application into CLI

### LINKED ISSUE #674: lsp: Merge CLIs and binaries
In yesterdays meeting we discussed the merge of lsp into the main open-vadl binary.

The CLI will look something along the lines of:
```
openvadl lsp [--port <int>]
```

## Reasons
Most of the following reasons only apply because the compiler and language server share the same frontend implementation and we don't have any plans to diverge them.

### Editor Plugins
Currently the strategy is that the plugins bundle the language server but that strategy doesn't seem to pay of as the language server is ~50MB and we would need to package native binaries (either GraalVM native builds or JLink customs JRE) for each of the 6 supported target architecture-os pairs. This will result in 300MB bundle size where only a sixth is ever used.

Another strategy often used is that the plugins download the language server for the right architecture, reducing the strain on bandwidth and the plugin always has the latest version without needing to wait for approval of a IDE/Editor marketplace. @Jozott00 invested a lot of time the last couple of days publishing packages and building for different architectures but that strategies still involves a lot of setup.

The core idea though is that users will only have use for the vadl-lsp and plugins if they also develop in vadl which makes it reasonable to assume that they already have a vadl toolchain on their machine. But the other direction is also a safe assumption that _most_ users that have a vadl toolchain will use it to develop specs and therefore will need an lsp.

For editor plugins this simplifies the strategy greatly, as they only need to find the `open-vadl` binary in the  PATH (or maybe somewhere else that is configurable in the plugin) and execute it.

Also there won't be a potentially slow and potentially failing download of an LS.

### User experience
For the user experience this is also a big advantage as the LS will always match compiler. So there will never be the case where the compiler throws a different error than the compiler.

### Developer comfort
For us developers this is also great because new features implemented already work in the LS before they are even pushed to GitHub as we just need create a new build of the compiler.

##  Notes on the implementation
For now I would leave the lsp implementation in it's submodule, because we are also rethinking the module setup at the moment.

Mostly this is just to remove the current entrypoint and add a subcommand to the CLI and adjust the setup.

### tsgo
The new go-based typescript compiler also bundles a lsp as a subcommand with the `--lsp` flag:
```
tsgo --lsp --help
Usage of lsp:
  -pipe string
        use named pipe for communication
  -pprofDir string
        Generate pprof CPU/memory profiles to the given directory.
  -socket string
        use socket for communication
  -stdio
        use stdio for communication
```

### gleam
The gleam compiler does something quite similar with the `lsp` subcommand (this version would fit our cli better):
```
Run the language server, to be used by editors

Usage: gleam lsp

Options:
  -h, --help  Print help
```

I also talked with Giacomo Cavalieri from the Gleam core team and he only had great things to say about this approach.

Jozott00:
This PR changes the setup according to #674, so instead of producing one openvadl CLI and one LSP executable, we just produce a single CLI executable that integrates the LSP as a sub command `openvadl lsp`.

Closes #674

### DIFF
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -31,8 +31,10 @@ jobs:
         env:
           isCI: true
       - name: Test Report
-        uses: dorny/test-reporter@v1
+        uses: dorny/test-reporter@v2
         if: success() || failure()
+        env:
+          NODE_OPTIONS: --max-old-space-size=8192
         with:
           name: report ${{ inputs.test-task }}
           path: '**/build/test-results/**/TEST-*.xml'


flofriday (approved):
LGTM

flofriday (approved):

PR #679, opened 2026-01-17
lcb: Setup builtins and intrinsics

kper:
In the initial implementation, there was no difference between builtin and intrinsic attributes. This was a bit awkward since both different. This PR generates a separate PR to generate the builtin attributes.

## Summary

- lcb: Differentiate between builtin and intrinsic


kper:
Current dependencies on/for this PR:

* master:
  * **PR #679** 👈
    * **PR #680**

This comment was autogenerated by [stax](https://github.com/cesarferreira/stax)

PR #680, opened 2026-01-17
lcb: Initialize mapping file between builtin and intrinsic

kper:
This PR creates a new file which will be responsible for the mapping between builtins and intrinsics. This PR does not have any logic, except the generation of the file.

## Summary

- lcb: Setup mapping file between builtin and intrinsic

PR #682, opened 2026-01-18
lcb: Initialize mapping file between builtin and intrinsic

kper:
This PR creates a new file which will be responsible for the mapping between builtins and intrinsics. This PR does not have any logic, except the generation of the file.

Restored https://github.com/OpenVADL/openvadl/pull/680


kper:
merged in https://github.com/OpenVADL/openvadl/pull/681

Issue #648, opened 2026-01-04
question: compute log2 of a constant


AndreasKrall:
I need the log2 of a constant to determine the necessary bit vector size. I tried:
```
  constant VL        = 128              // vector length, number of bits in a vector register
  constant ElementsB = VL >> 3          // number of byte   sized elements in a vector register
  constant Log2ElemB = VADL::cob(ElementsB-1)
```
but this does not work as I got the following error:
```
error: Type Mismatch
     ╭──[test.vadl:3:24]
     │
   3 │   constant Log2ElemB = VADL::cob(ElementsB-1)
     │                        ^^^^^^^^^^^^^^^^^^^^^^
     │
     Expected [class vadl.types.BitsType] but got `[Const<15>]`
```
Do you have an idea how to compute log2 for a constant? As a workaround I just use the biggest number.
And in the future it will not be necessary too, as the size of the index type of a forall will be done automatically.

PR #696, opened 2026-02-02
Chore: Refactor typechecker builtins

flofriday:
This is just a refactor that is supposed to unblock me. It generalizes the checking and const evaluation of built ins whether they come from a binary expression, unary expression or from a function call. 

All the tests are still passing and and will unblock to reuse that logic for many more cases in the typechecker.

Closes: #648

### DIFF
--- a/vadl/main/vadl/ast/TypeChecker.java
+++ b/vadl/main/vadl/ast/TypeChecker.java
@@ -606,8 +606,14 @@ public class TypeChecker
 
   }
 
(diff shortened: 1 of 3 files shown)


AndreasKrall:
@flofriday There are at least 4 errors in the constant evaluator at the end of the test report with null pointer exceptions.

flofriday:
Yes just saw it, I also expanded the FrontendIntegration Tests now that next time they really run all the available files :)

PR #704, opened 2026-02-07
typechecker: Improve exception thrown for wrong nodes

flofriday:
Some nodes simply should never reach the typechecker


flofriday:
This was already improved in https://github.com/OpenVADL/openvadl/pull/915

PR #717, opened 2026-02-12
iss: Fix LoopUnrollerPass to support aarch64/sve.vadl

Jozott00:
This PR addresses the issue of crashes when generating the `aarch64/sve.vadl` specification by fixing the `LoopUnrollerPass`.

Instructions that utilize `fold` expressions remain commented out because they are not yet supported by the ISS. This issue will be resolved in an upcoming PR.

Additionally, this PR modifies the dump behavior of the ISS.
Previously, the ISS would dump the VIAM in case of a crash. However, this became problematic due to the increasing size of the spec, so I disabled it.
To re-enable the behavior, you can use the `—-dump=on-failure` flag.


AndreasKrall (approved):
As far as I can understand it it looks fine to me, but PPC lbz test fails, where I can not see the connection.

PR #723, opened 2026-02-16
lsp: Add default communication via stdout/stdin

jaraonthe-dot-net:
Note that this cannot be tested via `./gradlew run --args="lsp"`, as gradle prints information to stdout. Use e.g. jlink build instead.

### DIFF
--- a/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
+++ b/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
@@ -39,6 +39,7 @@ import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseErrorCode;
 import org.eclipse.lsp4j.services.TextDocumentService;
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import vadl.ast.Ast;
 import vadl.ast.LspTokenizer;
 import vadl.ast.ModelRemover;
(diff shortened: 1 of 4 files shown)


vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -23,6 +23,7 @@
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
```
flofriday: **chore:** In general we try to avoid Nonnull as the nullability checker will assume non-null by default and only needs hints when a value is nullable.

flofriday (commented):
As discussed in the meeting today, there might be a cleaner option by simply configuring all loggers to use stderr, since we only ever use them for debug purposes.

With that change, the LSP (and all other packages used by it) can keep creating loggers as they have before. 

I've also looked over the whole codebase and removed possible interferring usages of stdout in #724. Which means there shouldn't be any problems. Even though redirecting System.out is a really interesting solution I don't think we will need it here. :/

jaraonthe-dot-net:
@flofriday Should changing logging to stderr be done in this PR, or somewhere else, e.g. #724?

jaraonthe-dot-net:
> The logging of the lsp server is very verbose, so I wonder if this could be an issue.

I just realized that our logback config file sets an `INFO` log level by default, and most of the lsp server logs are `DEBUG`. I.e. the vast amount of log output of the lsp server is thrown away anyway no matter if a NOPLogger is used.

PR #726, opened 2026-02-16
cosim: Fix build issue with disabled test

flofriday:
I think #725 missed to import the Disabled annotation which causes the build issues.


kper:
ah thanks
It was highlighted in the IDE

PR #732, opened 2026-02-18
setup: Fix checkstyle-suppressions for windows

der1steSpieler:
@linushdot Can you verify that it also works on your windows machine.


flofriday (approved):
LGTM
But couldn't test it on Windows for a lack of physical computers.

linushdot:
> @linushdot Can you verify that it also works on your windows machine.

What should I check? I don't really know what the bug addressed here is.

PR #747, opened 2026-02-24
setup: Update `prepareMergedJarsDir` task to include all projects' jars

### LINKED ISSUE #712: setup: Jlink build instructions missing
I fixed a bug in the compiler and wanted to rebuild the jlink image, but when I execute it still has the behaivor of the old compiler.

Here is an example after I did:
```bash
./gradlew :vadl-cli:installDist           
./gradlew :vadl-cli:jlink 
```

Executing the normal JVM build:

Image

Executing the JLink build:

Image

I must be missing something here because this doesn't seem right. (They files even get regenerated if I delete `vadl-cli/build/image/bin/openvadl`)

Jozott00:
Fixes #712

### DIFF
--- a/vadl-cli/build.gradle.kts
+++ b/vadl-cli/build.gradle.kts
@@ -102,10 +102,15 @@ jlink {
     }
 }
 
-// The plugin only merges dependency jars, not the main application jar.
-// Manually copy it into mergedjars to include it in the merged module.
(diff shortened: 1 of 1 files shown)


flofriday (approved):
Looks good to me 🎉 :)

PR #748, opened 2026-02-24
ume: Initial QEMU user-mode setup

arcane-quill:
check ume definitions in ATG and Definitions.java

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,9 @@ bin/
 ### Python ###
 .venv
 
+### Java ###
+*.log
+
 ### OTHER ###
 .swp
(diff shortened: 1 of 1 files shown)


vadl/main/vadl/ast/vadl.ATG
```diff
@@ -157,6 +157,7 @@ TOKENS
   ARCHITECTURE = "architecture".
   ARGUMENT     = "argument".
   ASSEMBLY     = "assembly".
+  AS_ID        = "AsId".
```
flofriday: **praise:** Nice catch, thanks ✨

PR #754, opened 2026-02-26
ume: Add hardcoded RV64UME QEMU files

### LINKED ISSUE #756: ume: Emit hardcoded linux-user target with RV64CSR compatibility
The first step to add a full UME generation backend is to emit a working QEMU linux-user mode target this is completely hardcoded, but compatible with a VADL generated QEMU system target.

Jozott00:
This PR introduces an initial hardcoded implementation for the RV64UME linux-user target, which is defined in `sys/risc-v/rv64ume.vadl`. This temporary specification is being used by @arcane-quill to implement the UME generation step-by-step.

The files included in this PR are essentially a simplified version of the RISC-V linux-user implementation. They are only generated when the `openvadl iss` command is executed with the `rv64ume.vadl` specification. 

Note that, due to QEMU’s configuration for linux-user, warnings are treated as errors. To disable this, the configuration command must include the `—-disable-werror` flag. Additionally, the compilation failed if the build directory was previously used to build a qemu-system target. Therefore, it’s a good idea to clear the build directory before configuring and building the linux-user target.

Closes #756

### DIFF
--- a/vadl/main/resources/templates/iss/configs/targets/gen-arch-linux-user.mak
+++ b/vadl/main/resources/templates/iss/configs/targets/gen-arch-linux-user.mak
@@ -1,0 +1,7 @@
+TARGET_ARCH=rv64ume
+TARGET_XML_FILES= gdb-xml/rv64ume-cpu.xml
+# TODO: Find out what can be removed (probably most of it)
+TARGET_BASE_ARCH=rv64ume
+TARGET_ABI_DIR=rv64ume
+TARGET_SYSTBL_ABI=common,64,rv64ume,rlimit,memfd_secret
+TARGET_SYSTBL=syscall.tbl
--- a/vadl/main/resources/templates/iss/linux-user/elfload.c
+++ b/vadl/main/resources/templates/iss/linux-user/elfload.c
@@ -1,0 +1,4500 @@
+/* This is the Linux kernel elf-loading code, ported into user space */
+#include "qemu/osdep.h"
+#include <sys/param.h>
+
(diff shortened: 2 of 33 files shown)


vadl/main/resources/templates/iss/linux-user/elfload.c
arcane-quill: elfload.c and syscall_defs.h were both already present in the qemu implementation. are these the only two files i need to carry over from the linux-user folder and do they need to be interpolated aswell?

vadl/main/resources/templates/iss/configs/targets/gen-arch-linux-user.mak
arcane-quill: what exactly do the xml files do?

vadl/main/resources/templates/iss/linux-user/elfload.c
Jozott00: Yeah, so those files include some target specific code that must be adjusted.

Currently it interpolates this:
https://github.com/OpenVADL/openvadl/blob/347a99fcfeb3aaa6b0213dc8b0689d1afa182e33/vadl/main/resources/templates/iss/linux-user/elfload.c#L1845-L1859

And in `syscall_defs.h` it adds some conditional includes like this (currently just duplicating what RISC-V does, however this must also be generated)
https://github.com/OpenVADL/openvadl/blob/347a99fcfeb3aaa6b0213dc8b0689d1afa182e33/vadl/main/resources/templates/iss/linux-user/syscall_defs.h#L73-L80

PR #774, opened 2026-03-01
frontend: Migrate to virtual file system

flofriday:
Some tooling needs a custom abstraction layer over files. For example the LSP needs to inject unsaved edits into files for the compiler.
Now all reads in the frontend should be determined through VFS abstraction.

### DIFF
--- a/vadl/main/vadl/ast/VadlParser.java
+++ b/vadl/main/vadl/ast/VadlParser.java
@@ -18,10 +18,9 @@ package vadl.ast;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.net.URI;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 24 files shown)


vadl/main/vadl/ast/VadlParser.java
```diff
@@ -40,42 +43,48 @@ public class VadlParser {
...
+  /**
+   * Parses the VADL source program at the specified path into an AST.
+   */
+  public static Ast parse(Path path, VirtualFileSystem fileSystem) throws IOException {
+    return parse(path, fileSystem, Collections.emptyMap());
   }
 
   /**
    * Parses the VADL source program at the specified path into an AST.
    * Works just like {@link VadlParser#parse(String, Map, URI)},
```
jaraonthe-dot-net: The referenced method no longer exists (URI).

vadl/main/vadl/utils/SourceLocation.java
```diff
@@ -208,46 +206,23 @@ public enum IDEDetectionMode {
...
     }
 
     String printablePath;
 
-    if (mode == IDEDetectionMode.ABSOLUTE || (mode == IDEDetectionMode.AUTO && (isIntelliJIDE()
-        || this.uri.getScheme().equals("memory")))) {
+    if (mode == IDEDetectionMode.ABSOLUTE || (mode == IDEDetectionMode.AUTO && isIntelliJIDE())) {
       // IntelliJ integrated terminal needs special treatment
-      printablePath = this.uri.toString();
+      printablePath = "file://" + fileSystem.toAbsolutePath(requireNonNull(path));
```
Jozott00: thought: We might wanna test if IDEs will also recognize non-URI paths (without `file://`)

flofriday: Great idea, maybe that workaround isn't needed anynmore 🤔

vadl/main/vadl/utils/VirtualFileSystem.java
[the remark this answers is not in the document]
Jozott00: I mean you could add a static `.instance()` method, but I think it is also fine the way it is.

Regarding passing around the file system, I also can't come up with any better approach

PR #775, opened 2026-03-01
iss: Add big-endian

BadGraphixD:
- Add [bigEndian] annotation for memory definition
- TCG now generates little-endian instructions per default and big-endian instructions when the annotation is used
- Memory functions in helper functions now explicitly use endianness given by memory definition
- Memory region init code now uses endianness given by memory definition
- gen-arch-softmmu.mak and gen-arch-linux-user.mak now contain explicit target endianness settings
- ppc64.vadl spec now uses big-endian
- ppc64 cosim is configured to read be memory and to use the be binary
- cosim memory checks are disabled for ppc64

### DIFF
--- a/vadl-test/resources/cosim_configs/ppc64_config.toml
+++ b/vadl-test/resources/cosim_configs/ppc64_config.toml
@@ -177,7 +177,7 @@ srr1 = "srr1"
 layer = "insn"
 
 # Whether reads/writes from/to memory should also be compared
-with_memory_checks = true
+with_memory_checks = false
 
 # "lockstep": All clients are run and compared one *execution-step* at a time.
 # 			  This means that the test will exit on the first divergence (or at the end if no diffs where found)


rascmatt (commented):
We currently assume hardcoded little-endian byte order for the decoder right now, this would have to be adapted to use the memory annotation: See 

https://github.com/OpenVADL/openvadl/blob/dc2b67fbdbd539c3eaf4bc4e966325bcc36f5147/vadl/main/vadl/vdt/passes/VdtInputPreparationPass.java#L91-L98

We might have to sync this with the load function used in the ISS, see 

https://github.com/OpenVADL/openvadl/blob/dc2b67fbdbd539c3eaf4bc4e966325bcc36f5147/vadl/main/resources/templates/iss/target/gen-arch/translate.c#L61-L65

I would suggest to leave the hardcoded big-endian order for the RTL target, and we'll address this once we disccuss endianness in the hardware generation.

BadGraphixD:
> We currently assume hardcoded little-endian byte order for the decoder right now, this would have to be adapted to use the memory annotation: See
> 
> https://github.com/OpenVADL/openvadl/blob/dc2b67fbdbd539c3eaf4bc4e966325bcc36f5147/vadl/main/vadl/vdt/passes/VdtInputPreparationPass.java#L91-L98
> 
> We might have to sync this with the load function used in the ISS, see
> 
> https://github.com/OpenVADL/openvadl/blob/dc2b67fbdbd539c3eaf4bc4e966325bcc36f5147/vadl/main/resources/templates/iss/target/gen-arch/translate.c#L61-L65
> 
> I would suggest to leave the hardcoded big-endian order for the RTL target, and we'll address this once we disccuss endianness in the hardware generation.

@rascmatt I am not entirely sure why these need to be changed. I have tried to change them and have encountered some problems.

As it is now, next_insn already loads and converts the instruction to the correct byte order (I am also not sure why it is then always swapped, so with little-endian systems, the byte order of the instruction as it is received by the decoder is big-endian). This already works.

With the byte order swapped in next_insn (meaning using false) and using big-endian in VdtInputPreparationPass, it almost behaves correctly: the correct instruction is selected, but the operands and immediate are extracted incorrectly (as if still little-endian). Example (ppc64):

`lis 16, -7255`

is translated to the following TCG:

`add_i64 X16,$0x0,$0xffffffffe3a90000 (no changes to both files)`
`add_i64 X15,X3,$0x3e0000 (changes to both files)`

PR #780, opened 2026-03-04
diagnostic: Fix order of snippets in DiagnosticPrinter

benjaminkasper99:
This `Diagnostic` I introduced in #758 references two snippets in the VADL source code when thrown (more precisely: the snippet of `instruction.assembly()` and the snippet where `conflictingRule` is defined):
https://github.com/OpenVADL/openvadl/blob/a112795c178c7600f80b6953828a8e7218610c99/vadl/main/vadl/lcb/passes/asm/AsmGrammarRuleGenerationPass.java#L201-L211

When testing this warning I noticed that for both snippets the same source location was printed in the console:
warning_wrong

I think the bug is in the `DiagnosticPrinter`, where the snippet of the `primaryLocation` is added _after_ all `secondaryLocations`, when it should be added _first_. Applying the suggested change prints the warning with correct source locations:
warning_right

Finally, regarding the affected snapshot tests:

- As this PR changes the order of printed snippets, tests containing multiple snippets must be adjusted to fit the printed order.
- The test `invalidImportAndRedeclareModel` actually captures the bug this PR tries to fix. It includes an error with snippets from two different VADL files, but currently expects only one location to be printed. With this PR the test is fixed and now expects the two different locations to be printed.

### DIFF
--- a/vadl/main/vadl/error/DiagnosticPrinter.java
+++ b/vadl/main/vadl/error/DiagnosticPrinter.java
@@ -129,9 +129,10 @@ public class DiagnosticPrinter {
             forceUnixPaths)));
     builder.append("     │\n");
 
-    // TODO: Sort them accordig to their line number in the future
-    var allSnippets = new ArrayList<>(diagnostic.multiLocation.secondaryLocations());
+    // TODO: Sort them according to their line number in the future
+    var allSnippets = new ArrayList<Diagnostic.LabeledLocation>();
     allSnippets.add(diagnostic.multiLocation.primaryLocation());
+    allSnippets.addAll(diagnostic.multiLocation.secondaryLocations());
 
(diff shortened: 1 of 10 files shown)


flofriday:
I agree that we should definitely fix this issue and the example you showed is really confusing, I'm sorry I pushed that off for so long.

But if we are fixing that problem we might as well do it properly.
Here are some of my thoughts.
- The first location after the tilte of the error/warning should be the root of the problem (what we call the primaryLocation)
- In my mind if there are secondary locations it's often because of a conflict, like redeclaration and therefore I thought it would make sense to first layout all the "correct" definitions and then lastly show you your conflicting declaration.
- I think it could make sense diving into other compilers and investigating how they handle multilocation diagnostics.

PR #784, opened 2026-03-05
frontend: Further imporve location retention

flofriday:


flofriday:
Deprecated in favor of #872

PR #792, opened 2026-03-07
typechecker: Fix crash on division by zero

flofriday:
The constant evaluator previously didn't have any checks for division by zero.


AndreasKrall (changes requested):
Please remove the double entry of UDIVS in line 159/160.

AndreasKrall (approved):
looks fine now

PR #800, opened 2026-03-09
chore: Set logging level for tc

kper:


Jozott00:
This isn't working for tests, as tests don't use the vadl-cli. This is already set in the `logback-test.xml`. Otherwise, you wouldn't see it in IntelliJ.

kper:
The github summary has a limitation of 1MB and 65000 chars. Therefore, attaching the logs will not work. One possible solution would be to upload the report to gh-pages.

Issue #787, opened 2026-03-05
typechecker: Error when passing numeric constants as arguments to AsmBuiltins


benjaminkasper99:
The typechecker throws an unexpected error when passing numeric constants as arguments to AsmBuiltin functions.

An unexpected error occurs when `LaIdEq` is called as shown in `Rule1` in the example below. The function `LaIdEq` takes two parameters: the first is an unsigned integer and the second is a string. Based on this signature, the call in `Rule1` should be valid, but it results in the error noted in the comment.

Interestingly, a manual cast as shown in `Rule2`, or passing numeric constants to user-defined functions as in `Rule3`, works as expected.

```java
instruction set architecture ISA = {}
application binary interface ABI for ISA = {}

assembly description AD for ABI = {
  grammar  = {
    Rule1 : ?( LaIdEq(1, "a") ) "a" | "a" ; // Type Mismatch: Expected [class vadl.types.UIntType, class vadl.types.StringType] but got `[Const<1>, String]`

    Rule2 : ?( LaIdEq(1 as UInt<1>, "a") ) "a" | "a" ; // works as expected

    Rule3 : ?( func(1) ) "a" | "a" ; // works as expected
  }

  function func(x: UInt<32>) -> Bool = false
}
```

Once this is fixed the usages of `LaIdEq` in the following specifications can be simplified:
- sys/risc-v/rv32im.vadl
- sys/risc-v/rv32imAsmGen.vadl
- sys/v-risc/ABI_no_suffix.vadl

Edit: There are also two tests in `AsmLL1CheckerTest` which can be re-enabled once this is fixed.

PR #810, opened 2026-03-12
typechecker: Improve const casting for builtins

flofriday:
Closes: #787


benjaminkasper99 (changes requested):
Constants work as expected now, but there is one issue I described below.

PR #815, opened 2026-03-14
asm: Remove unnecessary casts in assembly grammars

benjaminkasper99:
Remove VADL type casts from assembly grammars, as they are no longer necessary following the fixes in #810.


AndreasKrall (approved):
Looks fine to me

PR #832, opened 2026-03-22
iss: Add TB-state saved registers

BadGraphixD:
- Added the `[ execution state ]` annotation (the name may change)
- Only simple registers can be annotated (no register files or tensors)
- If the annotated register has a format type, the format fields can be referenced in the annotation
- Reads from annotated registers are translated to reads from the DisasContext instead of TCG instructions (currently "ctx->tb_state.reg")
- During TB state creation, annotated registers are stored in the TB state flags
- During DisasContext initialisation, the register values are decoded from the TB state flags and stored in ctx->tb_state. Parts of the reigsters that are not stored in the flags are set to zeros
- Heavily modified the TB chaining and jumping behaviour in TcgOpLoweringPass to account for changes to the tb state
  - Changes to the TB state cause the translation block to end
  - Not only non-static pc changes, but also non-static TB changes in general cause a tcg lookup
- Register reads that access only bits that are in the TB state are now converted to static read nodes. Reads of bits that are not in the TB state remain normal register reads
- The number of bits that can be annotated as saved in the TB state is now capped at 32 and an error is thrown if the limit is exceeded
- Added annotation to MSR register (the "sf" and "pr" bits) in ppc64 spec


BadGraphixD:
Found a bug in TcgOpLoweringPass, related to TcgSetIsJmp: when there is a conditional non-static write to a register saved in the tb-state, then it generates DISAS_CHAIN, which is wrong, because the next tb-state is not known at TCG gen time.

Jozott00 (commented):
Nice job!

Currently, writes to registers that are partially part of the CPU state are conservatively handled, such that the TB state is consider invalid even if the TB state register part stays the same. Later (in an other PR) we should optimize this. However, it will probably require some change on how the frontend constructs the VIAM, as currently analyzing this property is harder than necessary.

vadl/main/resources/templates/iss/target/gen-arch/translate.c
```diff
@@ -242,6 +257,11 @@ static void [(${gen_arch_lower})]_tr_tb_stop(DisasContextBase *db, CPUState *cpu
     			// jump to subsequent instruction
     			gen_goto_tb(ctx, 0, db->pc_next);
     			break;
+    		case DISAS_EXIT:
+    			// force tb exit
+          gen_update_pc(ctx, db->pc_next);
+          tcg_gen_exit_tb(NULL, 0);
+          break;
     		case DISAS_NORETURN:
```
Jozott00: **minor:** Please, align the indent with the other code. 

If the other code's indent is wrong, you may adjust the other code to yours.

vadl/main/vadl/iss/passes/tcgLowering/nodes/TcgSetIsJmp.java
```diff
@@ -41,7 +41,8 @@ public class TcgSetIsJmp extends TcgNode {
   public enum Type {
```
Jozott00, line 43: **minor:** write docs for the new `EXIT` type and clarify the difference to `NORETURN` and `CHAIN`.

Jozott00 (approved):
Great! Just two minor comments

Jozott00 (approved):

PR #834, opened 2026-03-22
ume: Add interpolation to rv64ume and custom template tags

### LINKED ISSUE #757: ume: Define VIAM definitions for the UME VADL specification
The VIAM `UserModeEmulation` definition (and others) should model all aspects of the user mode emulation VADL specification in a well consumable (and potentially extensible) way. 

In a first step, it is not required to add all UME feature configurations to the definitions, but only those, required to infer all values necessary to generate the UME implementation that was hardcoded in #756.

### LINKED ISSUE #759: ume: Add pass to add a hardcoded `UserModeEmulation` VIAM definition
Create a temporary pass that constructs a `UserModeEmulation` VIAM definition object (and all required sub-objects), only if the `rv64ume.vadl` specification is getting generated.

This object must then be added into the VIAM specification. The hardcoded definition must hold all the necessary information as if it were derived directly from the `rv64ume.vadl` specification.

### LINKED ISSUE #761: ume: Parameterize QEMU templates and render them using the VIAM
Until now, the emitted QEMU linux-user target comes from hardcoded C templates that interpolate no variables. This was done in #754 .

Now that we have a `UserModeEmulation` VIAM definition (#757) and passes that extract/compute all information necessary (#760), we can replace the hardcoded parts of the C templates by variable interpolation with variables that come from the VIAM and passes.

When this issue is solved, the C templates are not RV64UME specific anymore and contain no hardcoded assumptions about the generated linux-user target specification. Although the VIAM specification/definition itself is still hardcoded (#759).

arcane-quill:
PR for issue #757 (and #761, #759) 
interpolated hardcoded rv64ume files from linux-user 
created UmeHardcodedRiscvDefinitionPass and UmeTemplateRenderingPass to add custom template tags (e.g. registers, exceptions..) with hardcoded values that later get filled in by the vadl file 
added VIAM definition UserModeEmulation


vadl/main/vadl/ast/AstVisitor.java
flofriday, line 1: **chore:** Let's try to revert the changes in the frontend, since we don't want to add something here.

arcane-quill: since its just some empty spaces that were changed in the files, will they be removed from the PR section if i just edit it out?

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -50,45 +52,48 @@ void cpu_loop(CPURV64UMEState *env)
         case EXCP_ATOMIC:
             cpu_exec_step_atomic(cs);
             break;
-        case RV64UME_EXCP_EXC:
-            cause = env->arg_exc_cause;
+        case [(${gen_arch_upper})]_EXCP_EXC:
+            cause = env->[(${config.excCauseVar})];
```
Jozott00: Not all exceptions are called `EXC`. This is specification dependent, and therefore must not be hardcoded here. 

Also not all specifications have this cause variable. This is also specific to how we defined the VADL RISC-V specification.

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -50,45 +52,48 @@ void cpu_loop(CPURV64UMEState *env)
...
+                                      env->[(${register_tensors[0].name_lower})][ [(${config.sysReg})] ],
+                                                       [(${config.args.size() > 0 ? 'env->' + register_tensors[0].name_lower + '[' + config.args.get(0) + ']' : '0'})],
+                                                       [(${config.args.size() > 1 ? 'env->' + register_tensors[0].name_lower + '[' + config.args.get(1) + ']' : '0'})],
+                                                       [(${config.args.size() > 2 ? 'env->' + register_tensors[0].name_lower + '[' + config.args.get(2) + ']' : '0'})],
+                                                       [(${config.args.size() > 3 ? 'env->' + register_tensors[0].name_lower + '[' + config.args.get(3) + ']' : '0'})],
+                                                       [(${config.args.size() > 4 ? 'env->' + register_tensors[0].name_lower + '[' + config.args.get(4) + ']' : '0'})],
+                                                       [(${config.args.size() > 5 ? 'env->' + register_tensors[0].name_lower + '[' + config.args.get(5) + ']' : '0'})],
+                                                       0, 0);
```
Jozott00: The number of argument registers as well as the access (`register_tensors[0]`) is only valid for this RISC-V specification and not a generally valid approach.

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -50,45 +52,48 @@ void cpu_loop(CPURV64UMEState *env)
...
+                [# th:if="${config.hasIcacheFlush}"]
                 }
+                [/]
                 if (ret == -QEMU_ERESTARTSYS) {
-                    env->pc -= 4;
+                    env->[(${pc_reg.name_lower})] -= [(${config.insn_width_bytes})];
                 } else if (ret != -QEMU_ESIGRETURN) {
-                    env->x[RV64UME_REG_A0] = ret;
-                }
+                    env->[(${register_tensors[0].name_lower})][ [(${config.retReg})] ] = ret;                }
```
Jozott00: Same as for the argument registers

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -109,8 +114,8 @@ void target_cpu_copy_regs(CPUArchState *env, struct target_pt_regs *regs)
     TaskState *ts = get_task_state(cpu);
     struct image_info *info = ts->info;
 
-    env->pc = regs->sepc;
-    env->x[RV64UME_REG_SP] = regs->sp;
+    env->[(${pc_reg.name_lower})] = regs->[(${config.ptRegPc})];
```
Jozott00: What does `ptRegPc` stand for?

vadl/main/resources/templates/iss/linux-user/gen-arch/signal.c
```diff
@@ -9,13 +9,13 @@
...
- * Minimal sigcontext matching rv64ume state:
+ * Minimal sigcontext matching [(${gen_arch_lower})] state:
  * - pc
  * - x1..x31 (x0 is always zero and omitted)
  */
 struct target_sigcontext {
-    abi_long pc;
-    abi_long gpr[31];
+    abi_long [(${pc_reg.name_lower})];
+    abi_long gpr[ [(${register_tensors[0].size - 1})] ];
```
Jozott00: Also highly RISC-V spec specific

.idea/misc.xml
Jozott00, line 1: Please revert this file

vadl/main/resources/templates/iss/linux-user/gen-arch/signal.c
```diff
@@ -91,12 +91,12 @@ void setup_rt_frame(int sig, struct target_sigaction *ka,
...
-    env->x[RV64UME_REG_A0] = sig;
-    env->x[RV64UME_REG_A1] = frame_addr + offsetof(struct target_rt_sigframe, info);
-    env->x[RV64UME_REG_A2] = frame_addr + offsetof(struct target_rt_sigframe, uc);
-    env->x[RV64UME_REG_RA] = default_rt_sigreturn;
+    env->[(${pc_reg.name_lower})] = ka->_sa_handler;
+    env->[(${register_tensors[0].name_lower})][ [(${config.spReg})] ] = frame_addr;
+    env->[(${register_tensors[0].name_lower})][ [(${config.args[0]})] ] = sig;
+    env->[(${register_tensors[0].name_lower})][ [(${config.args[1]})] ] = frame_addr + offsetof(struct target_rt_sigframe, info);
+    env->[(${register_tensors[0].name_lower})][ [(${config.args[2]})] ] = frame_addr + offsetof(struct target_rt_sigframe, uc);
+    env->[(${register_tensors[0].name_lower})][ [(${config.raReg})] ] = default_rt_sigreturn;
```
Jozott00: All of this register handling is only valid for the RISC-V spec

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
```diff
@@ -0,0 +1,65 @@
...
+    return PassName.of("Rendering UME template: "
+        + templateFilename);
+  }
+
+  @Override
+  protected Map<String, Object> createVariables(PassResults passResults,
+                                                Specification specification) {
+    var vars = super.createVariables(passResults, specification);
+
+    UserModeEmulation ume = UserModeEmulation.createDefault();
```
Jozott00: It would be better to have a dedicated `UmeHarcodedRiscvDefinitionPass` instead of producing the default in the rendering pass.

vadl/main/vadl/pass/PassOrders.java
```diff
@@ -697,19 +698,24 @@ private static void addUserModeEmitPasses(PassOrder order, IssConfiguration conf
...
           .add(issDefault("/configs/targets/gen-arch-linux-user.mak", config))
 
           .add(issDefault("/linux-user/meson.build", config))
           .add(issDefault("/linux-user/elfload.c", config))
           .add(issDefault("/linux-user/syscall_defs.h", config))
 
           .add(issDefault("/linux-user/gen-arch/meson.build", config))
-          .add(issDefault("/linux-user/gen-arch/cpu_loop.c", config))
-          .add(issDefault("/linux-user/gen-arch/signal.c", config))
+          //.add(issDefault("/linux-user/gen-arch/cpu_loop.c", config))
```
Jozott00: Please remove the comments of the old implementation.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,222 @@
...
+ * This class defines architecture-specific parameters required for Linux-user emulation,
+ * including system call register mappings, signal trampoline instructions, stack
+ * alignment requirements, and exception IDs.
+ * </p>
+ */
+public class UserModeEmulation extends Definition {
+  /**
+   * Returns a map representation of this UserModeEmulation for template rendering.
+   */
+  public Map<String, Object> asMap() {
```
Jozott00: This should not be part of the VIAM definition, as it is tightly coupled with the generator specific template rendering. Please take a look at how this is handled for the `Instruction` and `RegisterTensor` definitions in the ISS.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,222 @@
...
+  private final String breakpointExcName;
+  private final String illegalInstrExcName;
+  private final String ptRegPc;
+  private final String ptRegSp;
+  private final String excCauseVar;
+  private final boolean hasIcacheFlush;
+  private final int insnWidthBytes;
+  private final int stackAlignMask;
+  private final int sigtrampLoadSyscallInstr;
+  private final int sigtrampTrapInstr;
```
Jozott00: `UserModeEmulation` is currently modeled as a bag of QEMU/RISC-V rendering parameters rather than as a VIAM definition. Unlike Processor, or Instruction, it stores primitive backend details (int register numbers, String field names, raw opcodes) instead of references to semantic VIAM objects. That makes it non-generic and hard to reuse for architectures with alias registers, multiple register files, or different exception/register conventions. This should be redesigned around VIAM-level references, with QEMU-specific encodings derived during lowering or kept in an ISS-specific extension. Please take a close look at other VIAM definition classes.

Also the implemented `InstructionSetArchitecture` and `Abi` is not stored in the class.

Jozott00 (changes requested):
Unfortunately the current state is too RISC-V specific and needs an overhaul. Please take a closer look at other VIAM definitions and how they are handled in the code emission.

arcane-quill:
> Unfortunately the current state is too RISC-V specific and needs an overhaul. Please take a closer look at other VIAM definitions and how they are handled in the code emission.

Thanks for the extensive review, I'm working on it now

arcane-quill:
hi @Jozott00 I resolved the stuff that was missing, hopefully we can get this merged now 

and do u know why passorders keeps telling me there is a merge conflict despite none of it showing up locally, even after rebasing?

flofriday:
> and do u know why passorders keeps telling me there is a merge conflict despite none of it showing up locally, even after rebasing?

Mostly guesses here but I assume you are not syncing your local branches with the remote (GitHub).

For example
```bash
git rebase master
```

Isn't good enough to rebase with the newest changes from GitHub.
This is mostly because `master` on your machine isn't updated and is stuck on an old commit.

**The long way** is to do:
```bash
git checkout master
git pull # (download from the remote, and also update the current master branch to the newest version from the remote)
git checkout <your-branch>
git rebase master
```

But you can even shorten this a bit with
```bash
git fetch # Downloads the branches from the remote but does not update the local branches only the branches prefixed with origin will be updated, so master is still stuck on an old commit but origin/master is synced with the remote
git rebase origin/master
```

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+ * This class defines architecture-specific parameters required for Linux-user emulation,
+ * including system call register mappings, signal trampoline instructions, stack
+ * alignment requirements, and exception IDs.
+ * </p>
+ */
+public class UserModeEmulation extends Definition {
+
+  private final Identifier excCauseVar;
+  private final RegisterTensor mainRegisterFile;
+  private final RegisterUtils.Register sysReg;
```
Jozott00: major: Using `vadl.lcb.templateUtils.RegisterUtils.Register` is the wrong abstraction

First, it is part of the LCB and should not be referenced by the `vadl.viam` package, second, it is a rendering utility, not a VIAM construct. It does not contain the original `RegisterTensor` and has exactly one index.

Instead, add a new class in the `vadl.viam` package:
```
public record RegisterLocation(
    RegisterResource resource,
    List<Constant.Value> indices,
    SourceLocation location
) implements WithLocation {
}
```
And use it instead. This will model those register references/locations much better and remains generic.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+  private final RegisterUtils.Register tpReg;
+  private final List<RegisterTensor> signalStateTensors;
+  private final List<RegisterUtils.Register> args;
+  private final Map<String, Integer> excIds;
+  private final Instruction syscallInstr;
+  private final ExceptionDef syscallException;
+  private final ExceptionDef breakpointExcName;
+  private final ExceptionDef illegalInstrExcName;
+  private final Identifier initialPc;
+  private final Identifier initialSp;
```
Jozott00: question: What is the `initialPC` and `initialSp`? Can you give me an example, cause I am confused how its value can be represented as `Identifier`.

From what I understand, this can be removed, as it can be synthesized later by the generator.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+    this.syscallException = syscallException;
+    this.mainRegisterFile = mainRegisterFile;
+    this.signalStateTensors = signalStateTensors;
+    if (args == null || args.isEmpty()) {
+      throw new IllegalArgumentException("args must not be null/empty");
+    }
+
+    if (excIds == null || excIds.isEmpty()) {
+      throw new IllegalArgumentException("excIds must not be null/empty");
+    }
```
Jozott00: minor: We use nullaway, so null checking of arguments that are not annoated with `@Nullable` isn't necessary.

Jozott00: question: Is there a reason why you require the args to be non-empty? From the generator side, it shouldn't matter if there is any argument.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+  private final RegisterUtils.Register retReg;
+  private final RegisterUtils.Register spReg;
+  private final RegisterUtils.Register raReg;
+  private final RegisterUtils.Register tpReg;
+  private final List<RegisterTensor> signalStateTensors;
+  private final List<RegisterUtils.Register> args;
+  private final Map<String, Integer> excIds;
+  private final Instruction syscallInstr;
+  private final ExceptionDef syscallException;
+  private final ExceptionDef breakpointExcName;
```
Jozott00: minor: Should be `breakpointExc` instead of `breakpointExcName`

(same below)

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+  private final ExceptionDef illegalInstrExcName;
```
Jozott00: minor: This should be annotated with `@Nullable` as I guess it is optional in general.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+    this.sigtrampLoadSyscallInstr = sigtrampLoadSyscallInstr;
+    this.sigtrampTrapInstr = sigtrampTrapInstr;
+  }
+
+  /**
+   * Creates a default {@link UserModeEmulation} configuration,
+   * pre-configured for the RISC-V architecture.
+   * * @return a standard RISC-V user-mode emulation setup.
+   */
+  public static UserModeEmulation createDummySolution() {
```
Jozott00: major: The `UserModeEmulation` class should not contain dummy values.

Move this into the `UmeHardcodedRiscvDefinitionPass` which is the pass that is specifically created for this purpose.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+        excIds,
+        mockSyscallInsn,
+        mockBreakpointExc,
+        mockIllegalExc,
+        ptRegsPcField, ptRegsSpField, riscvCauseVar,
+        true, 4, 0xf,
+        0x08b00893, 0x00000073);
+  }
+
+  public int getSigtrampLoadSyscallInstr() {
```
Jozott00: minor: By our conventions, getters do not start with `get` but just the property name.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+ * Represents the configuration for QEMU user-mode emulation.
+ * <p>
+ * This class defines architecture-specific parameters required for Linux-user emulation,
+ * including system call register mappings, signal trampoline instructions, stack
+ * alignment requirements, and exception IDs.
+ * </p>
+ */
+public class UserModeEmulation extends Definition {
+
+  private final Identifier excCauseVar;
```
Jozott00: question: What does this identifier represent and why is it an identifier?

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+  private final List<RegisterUtils.Register> args;
+  private final Map<String, Integer> excIds;
+  private final Instruction syscallInstr;
+  private final ExceptionDef syscallException;
+  private final ExceptionDef breakpointExcName;
+  private final ExceptionDef illegalInstrExcName;
+  private final Identifier initialPc;
+  private final Identifier initialSp;
+  private final boolean hasIcacheFlush;
+  private final int insnWidthBytes;
```
Jozott00: major: This should not be part of the VIAM definition because it can be directly inferred from the `syscallInstr` property

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+  private final List<RegisterTensor> signalStateTensors;
+  private final List<RegisterUtils.Register> args;
+  private final Map<String, Integer> excIds;
+  private final Instruction syscallInstr;
+  private final ExceptionDef syscallException;
+  private final ExceptionDef breakpointExcName;
+  private final ExceptionDef illegalInstrExcName;
+  private final Identifier initialPc;
+  private final Identifier initialSp;
+  private final boolean hasIcacheFlush;
```
Jozott00: major: Remove the entire `icache-flush` logic for now, including the special cases in code generation. It should work fine even without them in place. 

This is currently out of scope, and this boolean property will probably not be enough to control its logic anyway.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+
+/**
+ * Represents the configuration for QEMU user-mode emulation.
+ * <p>
+ * This class defines architecture-specific parameters required for Linux-user emulation,
+ * including system call register mappings, signal trampoline instructions, stack
+ * alignment requirements, and exception IDs.
+ * </p>
+ */
+public class UserModeEmulation extends Definition {
```
Jozott00, line 29: major: This should have a reference to the `InstructionSetArchitecture` and `Abi`

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+  private final Map<String, Integer> excIds;
+  private final Instruction syscallInstr;
+  private final ExceptionDef syscallException;
+  private final ExceptionDef breakpointExcName;
+  private final ExceptionDef illegalInstrExcName;
+  private final Identifier initialPc;
+  private final Identifier initialSp;
+  private final boolean hasIcacheFlush;
+  private final int insnWidthBytes;
+  private final int stackAlignMask;
```
Jozott00: major: The stack alignment must not be defined as mask, but as bytes. 

Additionally, this should most likely not be defined in the ume definition, as it can be extracted from the ABI.

For now, you can omit it entirely, as it is only used for signal handling, which we should not support for now.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+  private final int sigtrampLoadSyscallInstr;
+  private final int sigtrampTrapInstr;
```
Jozott00: major: Remove those from the VIAM entirely as we omit signal handling for now.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+public class UserModeEmulation extends Definition {
+
+  private final Identifier excCauseVar;
+  private final RegisterTensor mainRegisterFile;
+  private final RegisterUtils.Register sysReg;
+  private final RegisterUtils.Register retReg;
+  private final RegisterUtils.Register spReg;
+  private final RegisterUtils.Register raReg;
+  private final RegisterUtils.Register tpReg;
+  private final List<RegisterTensor> signalStateTensors;
```
Jozott00: major: Remove this property, as we omit signal handling for now

vadl/main/vadl/viam/UserModeEmulation.java
Jozott00, line 29: major: The current approach is not sufficient for proper signal handling, so this first step should ignore signal handling entirely. Minimal programs like writing to stdout and exiting will still work without it. Later, when the UME is in a good shape, we will add support for it.

So there are some properties that should be removed for now (see comments below).

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+ */
+public class UserModeEmulation extends Definition {
+
+  private final Identifier excCauseVar;
+  private final RegisterTensor mainRegisterFile;
+  private final RegisterUtils.Register sysReg;
+  private final RegisterUtils.Register retReg;
+  private final RegisterUtils.Register spReg;
+  private final RegisterUtils.Register raReg;
+  private final RegisterUtils.Register tpReg;
```
Jozott00: major: Remove them, as they are only required for signal handling

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+ * alignment requirements, and exception IDs.
+ * </p>
+ */
+public class UserModeEmulation extends Definition {
+
+  private final Identifier excCauseVar;
+  private final RegisterTensor mainRegisterFile;
+  private final RegisterUtils.Register sysReg;
+  private final RegisterUtils.Register retReg;
+  private final RegisterUtils.Register spReg;
```
Jozott00: major: This can be removed, as it should come from the ABI definition

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+ * <p>
+ * This class defines architecture-specific parameters required for Linux-user emulation,
+ * including system call register mappings, signal trampoline instructions, stack
+ * alignment requirements, and exception IDs.
+ * </p>
+ */
+public class UserModeEmulation extends Definition {
+
+  private final Identifier excCauseVar;
+  private final RegisterTensor mainRegisterFile;
```
Jozott00: major: Remove this and instead use a proper `RegisterLocation` struct (as described in an other comment).

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,333 @@
...
+  private final RegisterUtils.Register sysReg;
+  private final RegisterUtils.Register retReg;
+  private final RegisterUtils.Register spReg;
+  private final RegisterUtils.Register raReg;
+  private final RegisterUtils.Register tpReg;
+  private final List<RegisterTensor> signalStateTensors;
+  private final List<RegisterUtils.Register> args;
+  private final Map<String, Integer> excIds;
```
Jozott00: major: This should not be part of the VIAM. It is something that the generator may purely synthesize during generation.

vadl/main/vadl/viam/UserModeEmulation.java
Jozott00: major: You currently create all definitions from scratch. However, instead you should derive most of them from the `Specification` passed to the `UmeHardcodedRiscvDefinitionPass` if the specification is the `rv64ume.vadl` on.e 

Some examples: syscall instruction, all exceptions, all registers

vadl/main/vadl/iss/passes/UmeHardcodedRiscvDefinitionPass.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Override
+  public PassName getName() {
+    return PassName.of("UME Hardcoded RISC-V Definition");
+  }
+
+  @CheckForNull
+  @Override
+  public Object execute(PassResults passResults, Specification viam) throws IOException {
```
Jozott00, line 61: major: This should check whether we are currently running the `rv64ume.vadl` specification. If some other specification is processed, se just return.

vadl/main/vadl/iss/passes/UmeHardcodedRiscvDefinitionPass.java
```diff
@@ -0,0 +1,47 @@
...
+  public UmeHardcodedRiscvDefinitionPass(IssConfiguration configuration) {
+    super(configuration);
+  }
+
+  @Override
+  public PassName getName() {
+    return PassName.of("UME Hardcoded RISC-V Definition");
+  }
+
+  @CheckForNull
```
Jozott00: minor: Use `@Nullable` instead

vadl/main/resources/templates/iss/linux-user/gen-arch/target_syscall.h
```diff
@@ -2,11 +2,11 @@
...
- * Reference: linux/arch/rv64ume/include/uapi/asm/ptrace.h
+ * Reference: linux/arch/[(${gen_arch_lower})]/include/uapi/asm/ptrace.h
  */
 
-#ifndef LINUX_USER_RV64UME_TARGET_SYSCALL_H
-#define LINUX_USER_RV64UME_TARGET_SYSCALL_H
+#ifndef LINUX_USER_[(${gen_arch_upper})]_TARGET_SYSCALL_H
+#define LINUX_USER_[(${gen_arch_upper})]_TARGET_SYSCALL_H
 
 struct target_pt_regs {
```
Jozott00, line 11: minor: Reduce this to just `pc` and `sp` for now. 

Later, we can expand it.

vadl/main/resources/templates/iss/linux-user/gen-arch/signal.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Emulation of Linux signals for rv64ume user-mode.
+ * Emulation of Linux signals for [(${gen_arch_lower})] user-mode.
  */
```
Jozott00: major: I think for now, signals are too overwhelming. 

Let's turn it into a bare minimum implementation that just ensures compilation succeeds. 

Something like this (although not tested).
```c
#include "qemu/osdep.h"
#include "qemu.h"
#include "user-internals.h"
#include "signal-common.h"

void setup_rt_frame(int sig, struct target_sigaction *ka,
                    target_siginfo_t *info,
                    target_sigset_t *set, CPU[(${gen_arch_upper})]State *env)
{
    (void)ka;
    (void)info;
    (void)set;
    (void)env;

    if (sig == TARGET_SIGSEGV) {
        force_sig(TARGET_SIGSEGV);
    } else {
        force_sig(sig);
    }
}

long do_rt_sigreturn(CPU[(${gen_arch_upper})]State *env)
{
    (void)env;

    force_sig(TARGET_SIGSEGV);
    return 0;
}

void setup_sigtramp(abi_ulong sigtramp_page)
{
    (void)sigtramp_page;
}
```


Also test if it compiles when you are removing the whole content. In that case, do that instead.

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -50,45 +51,47 @@ void cpu_loop(CPURV64UMEState *env)
...
-                                     env->x[RV64UME_REG_A5],
-                                     0, 0);
+                                      env->[(${config.mainRegisterFile})][ [(${config.sysReg})] ],
+                                      [# th:each="arg : ${config.args}"]
+                                      env->[(${config.mainRegisterFile})][ [(${arg})] ],
+                                      [/]
+                                      0, 0);
+                [# th:if="${config.hasIcacheFlush}"]
                 }
+                [/]
```
Jozott00: major: Just assume there is no icache-flush

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -98,6 +101,43 @@ void cpu_loop(CPURV64UMEState *env)
                      trapnr);
             exit(EXIT_FAILURE);
         }
+        [/]
+        [# th:if="${#strings.isEmpty(config.excCauseVar)}"]
```
Jozott00: major: This check is unnecessary, as exceptions in the ISS are always rendered as cause variables in the state.

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -109,8 +149,8 @@ void target_cpu_copy_regs(CPUArchState *env, struct target_pt_regs *regs)
     TaskState *ts = get_task_state(cpu);
     struct image_info *info = ts->info;
 
-    env->pc = regs->sepc;
-    env->x[RV64UME_REG_SP] = regs->sp;
+    env->[(${pc_reg.name_lower})] = regs->[(${config.initialPc})];
+    env->[(${config.mainRegisterFile})][ [(${config.spReg})] ] = regs->[(${config.initalSp})];
```
Jozott00: major: As we just hardcode the `target_pt_regs` anyway, just use `regs->pc` and `regs->sp` instead.

vadl/main/resources/templates/iss/linux-user/gen-arch/syscall.tbl
Jozott00, line 1: minor: This whole table is currently hardcoded. However, I think we should do the dynamic interpolation in a separate PR.

vadl/main/vadl/pass/PassOrders.java
```diff
@@ -707,19 +709,22 @@ private static void addUserModeEmitPasses(PassOrder order, IssConfiguration conf
     var inputPath = config.inputPath();
     if (inputPath != null && inputPath.getFileName().endsWith("rv64ume.vadl")) {
       order
+          .add(new UmeHardcodedRiscvDefinitionPass(config))
+          .add(new UmeTemplateRenderingPass(config, "cpu_loop.c"))
```
Jozott00: major: The `cpu_loop.c` is now emitted twice

Jozott00 (changes requested):
I think there are still some important things to change. Mostly, what and how the VIAM definition stores specification details. 

I also think, that we should reduce the scope for the PR a bit, as signals are just a bit too much to start with. Therefore, let's completely remove all logic that is related to signals.

The UME generation requires quite a bit of information from the ABI definition and could share some constructs (like alias registers and special register references). Therefore, I will do a little ABI refactoring, so you can use `RegisterRef` instead of defining you own `RegisterLocation` as mentioned in a comment below.

EDIT: Also note, that this PR won't fully close #761 as there is still RISC-V specific stuff in the template files.

vadl/main/vadl/viam/UserModeEmulation.java
arcane-quill: because I tried to fill the variables with data without having to touch the abi, since then I'd have to add it to the vadl file, which isn't part of this PR, no? I've been trying to find a way around it but no luck. Do I just add the abi section to the rv64ume.vadl file now? I already did all the other requested changes

Jozott00: Yes, add/import the ABI to the spec. However, the `system instruction` and exceptions could already be determined from the ISA without the ABI. Also for most registers, the ABI is optional, as you can use the ISA registers instead of the ABI alias registers (for arguments and so on).

sys/risc-v/rv64ume.vadl
```diff
@@ -4,6 +4,42 @@ import rv64csr::{ RV64IMZicsr }
...
+    caller saved = [ a{0..7} ]
+    callee saved = [ ra, fp,
+                     X(9),  X(18), X(19), X(20), X(21),
+                     X(22), X(23), X(24), X(25), X(26), X(27) ]
+
+    special absolute address load instruction = LA
+    special local address load instruction   = LLA
+
+    special return instruction = RET
+    special call instruction = CALL
```
Jozott00, line 36: Fine for now, later this should be minimized to a minimal set of definitions that is required for the UME (https://github.com/OpenVADL/openvadl/issues/968).

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,106 @@
...
+ * including system call register mappings, signal trampoline instructions, stack
+ * alignment requirements, and exception IDs.
+ * </p>
+ */
+public class UserModeEmulation extends Definition {
+
+  private final InstructionSetArchitecture isa;
+  private final Abi abi;
+
+  private final List<RegisterUtils.Register> args;
```
Jozott00: major: You are still using the LCB template `RegisterUtils` instead of the `RegisterRef` as added in https://github.com/OpenVADL/openvadl/pull/948.

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
```diff
@@ -0,0 +1,101 @@
...
+        .orElseThrow(() -> new IllegalStateException("No UserModeEmulation defined"));
+
+    Abi abi = ume.abi();
+    RegisterTensor mainRegFile = (RegisterTensor) abi.stackPointer().registerFile();
+
+    Map<String, Integer> excIds = Map.of(
+        "ILLEGAL_INSTR", 2,
+        "ECALL", 11,
+        "BREAKPOINT", 3
+    );
```
Jozott00: major: This is still hardcoded. However, it should be obtained from the VIAM instead.

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
```diff
@@ -0,0 +1,101 @@
...
+  @Override
+  protected Map<String, Object> createVariables(PassResults passResults,
+                                                Specification specification) {
+    var vars = super.createVariables(passResults, specification);
+
+    UserModeEmulation ume = specification.userModeEmulation()
+        .orElseThrow(() -> new IllegalStateException("No UserModeEmulation defined"));
+
+    Abi abi = ume.abi();
+    RegisterTensor mainRegFile = (RegisterTensor) abi.stackPointer().registerFile();
```
Jozott00: major: This is not the right abstraction. In theory each syscall related register could originate from a different register file (or no "register file" at all). So you can't just assume that there is one `RegisterTensor` that includes all registers that are used for the syscall.

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
```diff
@@ -0,0 +1,101 @@
...
+
+    Map<String, Integer> excIds = Map.of(
+        "ILLEGAL_INSTR", 2,
+        "ECALL", 11,
+        "BREAKPOINT", 3
+    );
+
+    vars.put("config", Map.ofEntries(
+        Map.entry("sysReg", abi.stackPointer().addr()),
```
Jozott00: major: You assign the SP as sysReg, which I assume to be wrong.

vadl/main/vadl/viam/UserModeEmulation.java
```diff
@@ -0,0 +1,106 @@
...
+  private final Instruction syscallInstr;
+  private final ExceptionDef syscallException;
```
Jozott00: major: I think with the discussion summary https://github.com/OpenVADL/openvadl/pull/950#issuecomment-4486982124, the UME definition shouldn't have an `ExceptionDef` anymore.

The exception used in QEMU under the hood is a synthetic one.

vadl/main/vadl/iss/passes/UmeHardcodedRiscvDefinitionPass.java
```diff
@@ -0,0 +1,178 @@
...
+        Type.bits(1),
+        1
+    );
+
+    List<RegisterTensor.Dimension> dimensions = List.of(regDim, dummyDim);
+    RegisterTensor mainFile = new RegisterTensor(
+        new Identifier(new String[]{"x"},
+            SourceLocation.INVALID_SOURCE_LOCATION),
+        dimensions
+    );
```
Jozott00: major: This should come from the passed ISA, and shouldn't be created here.

vadl/main/vadl/iss/passes/UmeHardcodedRiscvDefinitionPass.java
```diff
@@ -0,0 +1,178 @@
...
+        ExceptionDef.Kind.DECLARED
+    );
+
+    ExceptionDef mockIllegalExc = new ExceptionDef(
+        new Identifier(new String[]{"ILLEGAL_INSTR"},
+            SourceLocation.INVALID_SOURCE_LOCATION),
+        emptyParams,
+        emptyGraph,
+        ExceptionDef.Kind.DECLARED
+    );
```
Jozott00: major: Either obtain them from the ISA or remove them (if they are not part of the current VADL spec)

vadl/main/vadl/iss/passes/UmeHardcodedRiscvDefinitionPass.java
```diff
@@ -0,0 +1,178 @@
...
+    Encoding emptyEncoding = new Encoding(
+        new Identifier(new String[]{"dummy_encoding"},
+            SourceLocation.INVALID_SOURCE_LOCATION),
+        dummyFormat, new Encoding.Field[0]);
+
+    Instruction mockSyscallInsn = new Instruction(
+        new Identifier(new String[]{"ECALL"},
+            SourceLocation.INVALID_SOURCE_LOCATION),
+        emptyGraph, emptyAssembly, emptyEncoding
+    );
```
Jozott00, line 123: major: Should also be obtained from the ISA and not created here.

Jozott00 (changes requested):
Your branch is currently conflicting with `origin/master`. This is because we moved the ISS PassOrder construction to `IssPassOrder`. 

Additionally there are still some issues from the last review that were not addressed yet. 

Also please test the generated QEMU, I think at the PR's current state, the generated QEMU wouldn't work.

vadl/main/vadl/viam/UserModeEmulation.java
arcane-quill: Yea I'm having issues here because my local master branch won't properly update (to pull in the RegisterRef), so I think I'd need to rebase, but I'm behind quite a few commits and I'm worried I'll mess up the history

Jozott00: You should be able to merge `origin/master` into your feature branch. 
I will squash this PR during merge, so the branch-local commit history doesn't really matter.

arcane-quill: Okay, I'll do a merge of master into this branch instead of a rebase then

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
arcane-quill: should I remove this, since we agreed on the synthetic exception anyway?

Jozott00: Yes, you can remove it. You can also skip illegal_instr and breakpoint handling for now.

arcane-quill:
I'm retrieving the args from the ABI now, is it fine like this?

Jozott00:
> I'm retrieving the args from the ABI now, is it fine like this?

You mean the syscall argument registers? Those must come from the UME definition and can't be taken from the ABI, as in general, the ABI argument registers and the syscall argument registers are not the same.

arcane-quill:
> > I'm retrieving the args from the ABI now, is it fine like this?
> 
> You mean the syscall argument registers? Those must come from the UME definition and can't be taken from the ABI, as in general, the ABI argument registers and the syscall argument registers are not the same.

right but since I only have the ABI right now, I was thinking it might be a bit cleaner than using the hardcoded version, but I can also revert it back to the way it was to avoid confusion

Jozott00:
I think for now you should create the register references the hardcoded way. However, don't create the `X` register tensor itself.

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
arcane-quill: @Jozott00 can I add pc to the struct? else it will always give me this error " ‘struct target_pt_regs’ has no member named ‘pc’; did you mean ‘sepc’?" for regs->pc

arcane-quill:
@Jozott00 I'm trying to get the cpuloop to work using the synthetic syscall exception. Is what I implemented in my latest commit the right approach?

Jozott00:
Yes, looks good to me 👍

sys/risc-v/rv64csr.vadl
```diff
@@ -141,14 +143,14 @@ instruction set architecture RV64IZicsr extending RV3264Base = {
...
+  model ECallInstr (name : Id, imm : Lit, excStat : Stat) : IsaDefs = {
+    instruction $name : Itype = $excStat                    // ecall / ebreak instructions
     encoding $name = {opcode = 0b111'0011, funct3 = 0b000, rd = 0b0'0000, rs1 = 0b0'0000, imm = $imm}
     assembly $name = (mnemonic)
     }
 
-  $ECallInstr (ECALL ; 0; M_ECALL)        // environment (sys) call
-  $ECallInstr (EBREAK; 1; BREAKPOINT)     // environment (sys) break
+  $ECallInstr (ECALL ; 0; raise Syscall)        // environment (sys) call
+  $ECallInstr (EBREAK; 1; raise Exc(ExcCode::BREAKPOINT))     // environment (sys) break
```
Jozott00: Please revert these chagnes, as they the previous implementation is the right one for full-system emulation.

Could you quickly explain why you changed the VADL definition here? The UME doesn't use the exceptions defined in the VADL spec anyway, no?

arcane-quill: the current rv64ume vadl file imports this vadl file. "0" is usually caught by the cpuloop in the "_EXC_EXCP" case that we used before so when I looked at the exceptions in this vadl file and replaced it with the syscall, it generated the helpers correctly and the exception was caught. since i had to add the raise to the behavior, this was the only thing that made sense, especially since we dont have exceptions defined in the ume vadl yet

Jozott00: The syscall used by UME is synthetic. Therefore the user should not require to provide any exception that is called "syscall" in a specific way to make the syscall exception mechanism work. This means, the entire mechanism is hidden from the user and this should also work if no exception is defined at all (as the user just provides the instruction that triggers the syscall).

Jozott00: I just thought about it: maybe we should let RV64UME extend the RV64IM for now, as Zicsr is currently very full system mode specific. So for simplicity, you could just use the RV64IM which includes no exceptions at all.

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -50,55 +43,31 @@ void cpu_loop(CPURV64UMEState *env)
...
-                          cause);
-                exit(EXIT_FAILURE);
+        case [(${gen_arch_upper})]_EXCP_UME_SYSCALL:
+            env->[(${pc_info.accessor})] += [(${config.insn_width_bytes})];
+                ret = do_syscall(env,
+                                  env->[(${config.sysRegFile})][ [(${config.sysReg})] ],
+                                  [# th:each="arg : ${config.args}"]
+                                  env->[(${arg.file})][ [(${arg.index})] ],
+                                  [/]
+                                  0, 0);
```
Jozott00: minor: Move this to the correct indention level

vadl/main/resources/templates/iss/linux-user/gen-arch/cpu_loop.c
```diff
@@ -109,8 +78,8 @@ void target_cpu_copy_regs(CPUArchState *env, struct target_pt_regs *regs)
     TaskState *ts = get_task_state(cpu);
     struct image_info *info = ts->info;
 
-    env->pc = regs->sepc;
-    env->x[RV64UME_REG_SP] = regs->sp;
+    env->[(${pc_info.accessor})] = regs->sepc;
+    env->[(${config.spRegFile})][ [(${config.spReg})] ] = regs->[(${config.spRegName})];
```
Jozott00, line 82: Fine for now, however, the SP could also be something else than a register file entry. Please create an issue, so we don't forget to fix this.

vadl/main/resources/templates/iss/linux-user/gen-arch/signal.c
```diff
@@ -1,171 +1,33 @@
...
-    return;
-
-badframe:
-    unlock_user_struct(frame, frame_addr, 1);
     if (sig == TARGET_SIGSEGV) {
-        ka->_sa_handler = TARGET_SIG_DFL;
+        force_sig(TARGET_SIGSEGV);
+    } else {
+        force_sig(sig);
     }
```
Jozott00: I think this can be reduced to `force_sig(sig)`

vadl/main/resources/templates/iss/linux-user/gen-arch/syscall.tbl
```diff
@@ -302,8 +302,8 @@
 
 244	or1k	or1k_atomic			sys_or1k_atomic
 
-258	rv64ume	rv64ume_hwprobe			sys_rv64ume_hwprobe
-259	rv64ume	rv64ume_flush_icache		sys_rv64ume_flush_icache
+258	[(${gen_arch_lower})]	[(${gen_arch_lower})]_hwprobe			sys_[(${gen_arch_lower})]_hwprobe
+259	[(${gen_arch_lower})]	[(${gen_arch_lower})]_flush_icache		sys_[(${gen_arch_lower})]_flush_icache
```
Jozott00, line 307: Fine for now, however, this table must be generated from the specification. Currently it hardcodes the two "special" riscv syscalls. Please create an issue for this.

vadl/main/resources/templates/iss/linux-user/gen-arch/target_cpu.h
```diff
@@ -1,40 +1,38 @@
...
-    RV64UME_REG_A5 = 15,
-    RV64UME_REG_A7 = 17,
+    [(${gen_arch_upper})]_REG_RA  = 1,
+    [(${gen_arch_upper})]_REG_SP  = [(${config.spReg})],
+    [(${gen_arch_upper})]_REG_TP  = 4,
+
+    [# th:each="arg, stat : ${config.args}"]
+    [(${gen_arch_upper})]_REG_ARG[(${stat.index})] = [(${arg.index})],
+    [/]
 };
```
Jozott00, line 13: The return address and thread pointer registers are both hardcoded.

Jozott00, line 13: Here you assume a register-file number as well. However, this doesn't have to be the case. It's fine for now, but should be fixed in a later PR (same issue as before).

vadl/main/resources/templates/iss/linux-user/meson.build
```diff
@@ -56,7 +56,7 @@ subdir('x86_64')
 subdir('xtensa')
 # VADL generated target
 [# th:if="${gen_arch_upper == 'RV64UME'}"]
```
Jozott00: The condition isn't valid anymore. It can be removed or adapted to some other condition that checks if a UME target was emitted.

vadl/main/vadl/iss/passes/tcg/IssHardcodedTcgAddOnPass.java
```diff
@@ -34,6 +45,27 @@
  */
 public class IssHardcodedTcgAddOnPass extends AbstractIssPass {
 
+  private static Graph createDummySyscallGraph() {
+    var g = new Graph("synthetic-ume-syscall");
+    g.setSourceLocation(SourceLocation.INVALID_SOURCE_LOCATION);
```
Jozott00: Remove the set of invalid source location. This is already the default value. (same for the other cases).

vadl/main/vadl/iss/passes/tcg/IssHardcodedTcgAddOnPass.java
```diff
@@ -43,17 +75,32 @@ public PassName getName() {
     return PassName.of("ISS Hardcoded TCG Add-Ons");
   }
 
-  List<Consumer<Instruction>> instrAddOns = List.of(
+  Consumer<Instruction> injectSyscallRaise = instr -> {
+
+    if (!"ECALL".equalsIgnoreCase(instr.simpleName())) {
+      return;
+    }
```
Jozott00: This shouldn't be hardcoded to "ECALL". The correct instruction can be retrieved from the UME definition.

vadl/main/vadl/iss/passes/tcg/IssHardcodedTcgAddOnPass.java
```diff
@@ -34,6 +45,27 @@
  */
 public class IssHardcodedTcgAddOnPass extends AbstractIssPass {
```
Jozott00: This class was originally used for stuff that was not supported by the frontend and therefore generated here. As there is no such stuff anymore, lets make it UME syntactic syscall exception specific. 

Rename the class (and file) and update the documentation to what the class actually does.

vadl/main/vadl/iss/passes/tcg/IssHardcodedTcgAddOnPass.java
```diff
@@ -34,6 +45,27 @@
...
 
+  private static Graph createDummySyscallGraph() {
```
Jozott00: Lets move this private method to the bottom of the class.

vadl/main/vadl/iss/passes/UmeHardcodedRiscvDefinitionPass.java
```diff
@@ -0,0 +1,145 @@
...
+
+    Format dummyFormat = new Format(new Identifier(new String[]{"dummy_format"},
+        SourceLocation.INVALID_SOURCE_LOCATION), mockType);
+
+    Encoding emptyEncoding = new Encoding(
+        new Identifier(new String[]{"dummy_encoding"},
+            SourceLocation.INVALID_SOURCE_LOCATION),
+        dummyFormat, new Encoding.Field[0]);
+
+    Instruction mockSyscallInsn = new Instruction(
```
Jozott00: ECALL is a real instruction that is already part of the passed RV64UME specification. So this shouldn't be syntactically created, but instead retrieved from the specification. Therefore you also don' tneed the function, assembly, format and encoding.

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
```diff
@@ -0,0 +1,93 @@
...
+
+    Abi abi = ume.abi();
+
+    vars.put("config", Map.ofEntries(
+        Map.entry("sysReg", ume.getSyscallNumber().singleIndex()),
+        Map.entry("sysRegFile", ume.getSyscallNumber().resource().simpleName().toLowerCase()),
+        Map.entry("retReg", ume.getSyscallReturn().singleIndex()),
+        Map.entry("retRegFile", ume.getSyscallReturn().resource().simpleName().toLowerCase()),
+        Map.entry("spReg", abi.stackPointer().addr()),
+        Map.entry("spRegFile",  abi.stackPointer().registerFile().simpleName().toLowerCase()),
```
Jozott00, line 74: Corresponding to previous comments: this assumes that these registers are part of a register file. This should be fixed together with the other instances of this issue.

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
```diff
@@ -0,0 +1,93 @@
...
+        Map.entry("spRegName",  abi.aliases()
```
Jozott00, line 79: Please add a comment that explains that this is just trying to find the most idiomatic name for the SP register. So that if the user defines a custom alias, it uses its name instead of the default "sp".

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
```diff
@@ -0,0 +1,93 @@
...
+                Pair.of(abi.stackPointer().registerFile(), abi.stackPointer().addr()),
+                List.of(new Abi.RegisterAlias("sp"))
+            )
+            .getFirst().value()),
+        Map.entry("args", ume.args().stream()
+            .map(ref -> Map.of(
+                "index", ref.singleIndex(),
+                "file",  ref.resource().simpleName().toLowerCase()
+            ))
+            .toList()),
```
Jozott00, line 90: Same register file problem as before (fine for now).

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
```diff
@@ -0,0 +1,93 @@
...
+        Map.entry("syscallInstr", ume.syscallInstr().simpleName()),
```
Jozott00, line 91: What is the syscallInstr name used for in the template?

vadl/main/vadl/pass/order/IssPassOrder.java
```diff
@@ -215,37 +216,37 @@ private static void addIssTemplatePasses(PassOrder order, IssConfiguration confi
...
-          .add(IssDefaultRenderingPass.issDefault("/linux-user/gen-arch/target_proc.h", config))
-          .add(IssDefaultRenderingPass.issDefault("/linux-user/gen-arch/target_resource.h",
-              config))
-          .add(IssDefaultRenderingPass.issDefault("/linux-user/gen-arch/target_signal.h", config))
-          .add(IssDefaultRenderingPass.issDefault("/linux-user/gen-arch/target_structs.h",
-              config))
-          .add(IssDefaultRenderingPass.issDefault("/linux-user/gen-arch/target_syscall.h",
-              config))
-          .add(IssDefaultRenderingPass.issDefault("/linux-user/gen-arch/termbits.h", config));
+          .add(new UmeHardcodedRiscvDefinitionPass(config))
```
Jozott00: This must be placed at the very start of the ISS pass order. Currently only the emit passes have the hardcoded UME definition available, while e.g. the syscall exception injection pass does not. Move it to the `create` method. Once you implemented the frontend part, it will be removed anyway.

vadl/main/vadl/iss/passes/tcg/UmeSyntheticSyscallExceptionPass.java
```diff
@@ -43,17 +75,32 @@ public PassName getName() {
...
   @Override
   public @Nullable Object execute(PassResults passResults, Specification viam)
       throws IOException {
 
-    normalTcgInstrs(viam)
-        .forEach(i ->
-            instrAddOns.forEach(f -> f.accept(i)));
+    var isa = viam.isa().orElseThrow();
+    var excInfo = isa.expectExtension(ExceptionInfo.class);
+    excInfo.addException(UME_SYSCALL_EXC);
```
Jozott00, line 94: Only add the syscall exception if we are actually generating UME. Just add a check at the top of the execute method to return early.

vadl/main/resources/templates/iss/linux-user/gen-arch/target_elf.h
```diff
@@ -5,8 +5,8 @@
  * later version. See the COPYING file in the top-level directory.
  */
 
-#ifndef RV64UME_TARGET_ELF_H
-#define RV64UME_TARGET_ELF_H
+#ifndef [(${gen_arch_upper})]_TARGET_ELF_H
+#define [(${gen_arch_upper})]_TARGET_ELF_H
 static inline const char *cpu_get_model(uint32_t eflags)
 {
     return "max";
```
Jozott00: This causes a failure of the invocation, as `max` is not a valid CPU model (the only one is `rv64ume-cpu` in case of rv64ume)

Jozott00 (changes requested):
Looks better now. There are still many cases where you assume the register to be part of a register file with a certain rendered structure. For now we can leave them as is. 

A couple of issues to fix are minor or something that still assumes RISC-V.

Have you tested the generated user mode QEMU with some user-space binary?

vadl/main/vadl/iss/passes/tcg/IssHardcodedTcgAddOnPass.java
arcane-quill: I couldn't find the old IssHardcodedTcgAddOnPass mentioned anywhere in the documentation, do I describe the now renamed class in the ume section of the tutorial.md? If yes, do I add the commit to this PR? I did change the comment that's above the class name in the file already

vadl/main/vadl/iss/passes/UmeTemplateRenderingPass.java
arcane-quill: Its not yet used in the template, I added it because its already hardcoded incase I need to use it later

Jozott00 (approved):

Jozott00:
@arcane-quill Please update the branch to the lastest commit on master (merging master into this branch should be fine, as we will squash the PR anyways).

PR #838, opened 2026-03-22
frontend: Add internal api for potentially broken ASTs

flofriday:
The LSP might want to access the AST even if the frontend produces errors.
In that case the AST might not be in any good state and the LSP has to dread lightly but might still benefit from it.

@jaraonthe-dot-net can you confirm that this API would help you with the LSP development?

### DIFF
--- a/vadl-cli/main/vadl/cli/BaseCommand.java
+++ b/vadl-cli/main/vadl/cli/BaseCommand.java
@@ -178,17 +178,11 @@ public abstract class BaseCommand implements Callable<Integer> {
    * @return the AST.
    */
   private Ast parseToAst() {
-    try {
-      Ast ast = VadlParser.parse(input, new DiskVirtualFileSystem(),
-          Objects.requireNonNullElseGet(modelOverrides, Map::of));
-      new Ungrouper().ungroup(ast);
-      new ModelRemover().removeModels(ast);
-      return ast;
-    } catch (IOException e) {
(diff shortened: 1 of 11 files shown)


jaraonthe-dot-net:
> @jaraonthe-dot-net can you confirm that this API would help you with the LSP development?

Getting a partial AST even if there are Diagnostics is certainly helpful, but this current implementation is not giving me anything new to work with:
- Publish Diagnostics: Only cares about Diagnostics, not AST, and it still only receives Diagnostics from the first stage that fails, nothing subsequent (which is probably impossible to support anyway).
- Goto Definition: Currently only uses the first (the `VadlParser`) stage, thus it is not using `Frontend` methods at all - if at some point further stages (the `TypeChecker`?) add more data that is useful for Goto Definition, then maybe your approach is useful here. Currently, any Diagnostic produced by `VadlParser` (like simple syntax errors) hinder Goto Definition.
- Syntax Highlighting currently uses the `Scanner` directly. It is not clear where we'll go from here. If at some point in the future the AST is relevant here, we face a similar situation as for Goto Definition above: Do we need any further stages besides `VadlParser`?

Do you have any plans on `VadlParser` returning a `PotentiallyBrokenAst`? Even if the Parser simply stops at the first Diagnostic (e.g. a syntax error), a partial AST up to that point may still be useful for Goto Definition.

flofriday:
This is closed in favor of #867

PR #842, opened 2026-03-23
lsp: Add tests

### LINKED ISSUE #821: lsp: Add unit tests
As the groundwork of the language server is done, all the classes (if feasible) in `vadl.lsp` should be covered with unit tests. This should protect us against regressions going forward.

jaraonthe-dot-net:
Closes #821


vadl-lsp/main/vadl/lsp/LspSnapshotFileSystem.java
```diff
@@ -53,10 +54,21 @@ class LspSnapshotFileSystem implements VirtualFileSystem {
...
+  }
+
+  /**
+   * Creates a VirtualFileSystem backed with the given documents.
+   *
+   * @param documents Maps URI string to corresponding document. This map is copied.
+   */
+  LspSnapshotFileSystem(Map<String, Document> documents) {
+    this(documents, new DiskVirtualFileSystem());
   }
```
flofriday: **nitpick:** Maybe it makes sense to always force the caller to provide a VFS?

vadl-lsp/test/vadl/lsp/DocumentTest.java
```diff
@@ -0,0 +1,267 @@
...
+  @Test
+  void textLinesConstructor() {
+    Document document = new Document(TEST_URI, 13, TEST_LINES);
+
+    assertThat(document.uri).isEqualTo(TEST_URI);
+    assertThat(document.version).isEqualTo(13);
+    assertThat(document.textLines).isEqualTo(TEST_LINES);
+    assertThat(document.getText()).isEqualTo(TEST_TEXT);
+  }
+
```
flofriday, line 72: **note:** I don't think this tests anything meaningful. Obviously we can leave it, it doesn't hurt either except for adding noise.

But in the future I think we can skip that.

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/mainTest-aliasRegister-input.vadl
```diff
@@ -0,0 +1,9 @@
+// 7:22 position
+instruction set architecture TEST = {
+  register X : Bits<5> -> Bits<32>
+}
+application binary interface ABI for TEST = {
+  alias register hello = X(0)
+
+  return address = hello
+}
```
flofriday: **issue:** I think there is an off-by one error here because the actual line of the goto is 8.

I totally see how that can happen but I think they all are much easier to read in editor if the numbers line up.

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/aliasRegister.snapshot
flofriday, line 1: **praise:** The snapshots are a bit verbose but they are quite easy to read, great work!

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/noDefinition-input.vadl
```diff
@@ -0,0 +1,2 @@
+// 1:3 position
+constant hrumph = 42
```
flofriday, line 2: **thought:** Maybe it would be nice to also have a comment in the files what's happening in them.

I know the file name explains them somewhat but still, a simple _Clicking a keyword simply doensn't have a goto location_ would already help by reading them.

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/mainTest-noDefinition-input.vadl
```diff
@@ -0,0 +1,2 @@
+// 1:3 position
```
flofriday: **thought:** In the frontend tests we often have some All Caps labels for test related comments

Maybe something like that:
```
// GOTO POSITION 1:3
```

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/aliasRegister-input.vadl
flofriday, line 1: **nitpick:** I think we can clean up the file names.

`mainTest-aliasRegister-input.vald` -> `aliasRegister.vadl`
`mainTest-aliasRegister.dat` -> `aliasRegister.dat`

Or maybe instead of a generic `.dat` we could use `.snap` like rusts insta library or `.snapshot`, but dat is also fine.

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.PublishDiagnosticsTest/twoFiles.snapshot
```diff
@@ -0,0 +1,56 @@
...
+    version = 0
+  ]]
+
+- ... which looks like this: 
+  Error: Errors in imported file: 
+  mainTest-twoFiles-otherFile.vadl
+  ================
+    <RANGE></>// error in imported file
+    import "mainTest-twoFiles-otherFile"::TEST
+
```
flofriday, line 51: **chore:** This error reporting doesn't look right, fixing this might be too much for this PR, but let's create an issue to investigate this.

vadl-lsp/test/vadl/lsp/LspSnapshotFileSystemTest.java
flofriday, line 1: **thought:** Again I'm not convinced that such simple unit tests provide a lot of value but again we can leave it.

vadl-lsp/test/vadl/lsp/integration/GotoDefinitionTest.java
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * <p>Uses snapshot input files to parameterize test cases.
+ */
+public class GotoDefinitionTest extends IntegrationTest {
+  private static final String INPUT_NAME = "input.vadl";
+  private static final Pattern POSITION_PATTERN =
+      Pattern.compile("^//\\s*(\\d+):(\\d+)", Pattern.MULTILINE);
+
+  @ParameterizedTest
+  @ValueSource(strings = {"noDefinition", "aliasRegister", "isa", "twoFiles"})
```
flofriday, line 48: **thought:** Maybe this list can be dynamically created by inspecting the files on disk. There might be some extra work to avoid files that are just dependencies and not actual inputs but then adding new tests is only creating one new file.

flofriday (commented):
Overall, quite good, just some adjustments.
But the snapshots are quite readable. 🎉

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/aliasRegister-input.vadl
jaraonthe-dot-net: - I like `.snapshot`; I'm using that now
- I thought about using subdirectories for tests that have test cases (e.g. `mainTest/aliasRegister.snapshot`), but IntelliJ displays resource directory paths like Java namespaces (i.e. `.` as separator), thus it may be confusing what is the test class and what the test method. So I'll leave the file names as is.

vadl-lsp/test/vadl/lsp/integration/GotoDefinitionTest.java
jaraonthe-dot-net: This is a good idea, but I'd like to keep it out-of-scope for now. It can always be revisited if manually maintaining this list becomes too annoying.

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.PublishDiagnosticsTest/twoFiles.snapshot
jaraonthe-dot-net: It behaves exactly as designed. LSP is asked to provide diagnostics for `...-input.vadl`, not for the erroneous `...-otherFile.vadl`. I.e. there are no errors to report for the requested file itself, but the user should still be aware that there are errors in included files (which may lead to errors in the current file not being reported as the parser stopped before). That's why this special error is placed right at the beginning of the document.

We could try to put the error at the location of the import statement (is that the place where the user would expect it?), but I'm not sure if the Parser and Ast provide the required information. If you would like to have this behavior, feel free to create an issue.

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/aliasRegister-input.vadl
flofriday: I'm fine with the .snapshot, but with the structure of the GotoDefinitionTest class I don't think you will ever have any method other than the `mainTest`. So I don't get why we would need any sub-directory nor prefix.

This structure with a single parameterized test method is also what we have in many other snapshot tests so if we both cannot think of a second method that will be added soon here, I would strongly be in favor of reducing the clutter in the naming.

PS: As a side-note I think the subdirectories would have worked, that's also how we structed the tests in the frontend and it's always good to stick to some conventions within a project.

jaraonthe-dot-net: Ok, I've removed the test method from those snapshot names that don't need it.

flofriday (approved):

PR #867, opened 2026-04-02
frontend: Add internal broken parser api

flofriday:
This API is mostly for the LSP, @jaraonthe-dot-net can you verify that this would be helpful for you?

There is also a commit in here that cleans up the parser API and removes a couple of barely used functions.

This PR is the successor to #838


jaraonthe-dot-net (changes requested):
If I'm not mistaken, this should be useful for Goto Definition, as locations _before_ an error can be serviced. I should try this out in a follow-up issue/PR.

So yes, this PR is useful and you may continue.

jaraonthe-dot-net (approved):

PR #870, opened 2026-04-09
sys/ppc64: Improve XER flag handling

kbschlosser:
- Improves XER flag accesses: now split into 6 registers
- Fixes cosim issue: sliced registers now compare the first occurrence of a register name (PowerPC QEMU outputs XER twice for some reason; only the first occurrence is correct). Full register comparisons already behave this way.


AndreasKrall (approved):
Looks fine to me.

Issue #660, opened 2026-01-07
tensors: support min and max in forall fold


AndreasKrall:
Currently `min` and `max` are not supported as fold operators. SVE needs these operators. They should be added.

AndreasKrall:
You cannot find it in the documentation as we have nothing there. We should discuss how to solve it. What I need is at least signed and unsigned min and max. We have the choice to add the four builtins `smin`, `smax`, `umin` and `umax` to our VADL builtin collection and we can write these names without `VADL::` additionally to the operators allowed after the fold keyword. So we would allow not only the binary operators `+`, `*`, `&`, `|` and `^`, but also these builtins added to our builtin collection. But it could be that in the future we could need other fold operations like saturated addition. So the more general solution would be to allow the usage of all binary VADL builtins and we add these 4 builtins to our builtin collection. Another more general solution could be allow the calling of any function. The disadvantage is that we can only define functions for operands with a fixed bit width and the implementation overhead is higher. In the not too distant future we also have to support the floating point operations. What are your opinions?
```
forall i in 0..3 fold smin       with Z(rs1)(i)  // minimum what I need
forall i in 0..3 fold VADL::smin with Z(rs1)(i)  // I do not like the namespace requirement
forall i in 0..3 fold usatadd    with Z(rs1)(i)  // unsigned saturated add
forall i in 0..3 fold umull      with Z(rs1)(i)  // unsigned multiplication
```

flofriday:
Yes, I agree, but this will be some work because a bit in the parser has to be rewritten.

flofriday:
Closed with: #875

PR #875, opened 2026-04-11
viam: Add min max builtin functions

flofriday:
This PR also fixes a issue in the dumps, because identifierpaths previously didn't print the value they were holding.

Related issue: #660

### DIFF
--- a/vadl/main/vadl/types/BuiltInTable.java
+++ b/vadl/main/vadl/types/BuiltInTable.java
@@ -511,6 +511,47 @@ public class BuiltInTable {
           .returnsFirstBitWidthAndStatus(UIntType.class)
           .build();
 
+  /**
+   * {@code function smin ( a : SInt<N>, b : SInt<N> ) -> SInt<N>}
+   */
+  public static final BuiltIn SMIN =
+      func("VADL::smin", null, Type.relation(SIntType.class, SIntType.class, SIntType.class))
+          .compute((Constant.Value a, Constant.Value b) -> a.min(b, true))
+          .takesAllWithSameBitWidths()
+          .returnsFirstBitWidth(SIntType.class)
+          .build();
+
+  /**
+   * {@code function umin ( a : UInt<N>, b : UInt<N> ) -> UInt<N>}
+   */
+  public static final BuiltIn UMIN =
+      func("VADL::umin", null, Type.relation(UIntType.class, UIntType.class, UIntType.class))
+          .compute((Constant.Value a, Constant.Value b) -> a.min(b, false))
+          .takesAllWithSameBitWidths()
+          .returnsFirstBitWidth(SIntType.class)
+          .build();
+
+
+  /**
+   * {@code function smax ( a : SInt<N>, b : SInt<N> ) -> SInt<N>}
+   */
+  public static final BuiltIn SMAX =
+      func("VADL::smax", null, Type.relation(SIntType.class, SIntType.class, SIntType.class))
+          .compute((Constant.Value a, Constant.Value b) -> a.max(b, true))
+          .takesAllWithSameBitWidths()
+          .returnsFirstBitWidth(SIntType.class)
+          .build();
+
+  /**
+   * {@code function umax( a : UInt<N>, b : UInt<N> ) -> UInt<N>}
+   */
+  public static final BuiltIn UMAX =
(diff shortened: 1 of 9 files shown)


vadl/main/vadl/viam/Constant.java
```diff
@@ -511,6 +511,28 @@ public Constant.Value modulo(Constant.Value other, boolean signed) {
...
+    /**
+     * Return the min value. If both are equal, the first will be returned.
+     */
+    public Constant.Value min(Constant.Value other, boolean signed) {
+      ensure(type().isTrivialCastTo(other.type()),
+          "Min must be of same type, but other was %s",
+          other.type());
+
+      return this.leq(other, signed).intValue() == 0 ? this : other;
+    }
```
flofriday, line 523: **question:** Is this _shortcut_ in the implementation here valid, or should I've reimplemented it?

Jozott00: Yeah should be fine. However, I think using `.bool()` instead of `.intValue() == 0` would be cleaner (and more readable). Note that the `.bool()` is the inverse of the current comparison. 

However, taking a closer look, the comparison looks wrong to me. 
`this.leq(other, ...).intValue()` returns 1 if `this <= other` and 0 otherwise. So it should be `this.leq(other, signed).intValue() != 0 ? this : other;` no?

```suggestion
    public Constant.Value min(Constant.Value other, boolean signed) {
      ensure(type().isTrivialCastTo(other.type()),
          "Min must be of same type, but other was %s",
          other.type());

      return this.leq(other, signed).bool() ? this : other;
    }
```

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -511,6 +511,47 @@ public class BuiltInTable {
           .returnsFirstBitWidthAndStatus(UIntType.class)
           .build();
 
+  /**
+   * {@code function smin ( a : SInt<N>, b : SInt<N> ) -> SInt<N>}
+   */
+  public static final BuiltIn SMIN =
+      func("VADL::smin", null, Type.relation(SIntType.class, SIntType.class, SIntType.class))
+          .compute((Constant.Value a, Constant.Value b) -> a.min(b, false))
```
Jozott00: Wrong sign flag
```suggestion
          .compute((Constant.Value a, Constant.Value b) -> a.min(b, true))
```

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -511,6 +511,47 @@ public class BuiltInTable {
...
+          .returnsFirstBitWidth(SIntType.class)
+          .build();
+
+
+  /**
+   * {@code function smax ( a : SInt<N>, b : SInt<N> ) -> SInt<N>}
+   */
+  public static final BuiltIn SMAX =
+      func("VADL::smax", null, Type.relation(SIntType.class, SIntType.class, SIntType.class))
+          .compute((Constant.Value a, Constant.Value b) -> a.max(b, false))
```
Jozott00: Wrong sign flag
```suggestion
          .compute((Constant.Value a, Constant.Value b) -> a.max(b, true))
```

vadl/main/vadl/viam/Constant.java
```diff
@@ -511,6 +511,28 @@ public Constant.Value modulo(Constant.Value other, boolean signed) {
...
+
+    /**
+     * Return the max value. If both are equal, the first will be returned.
+     */
+    public Constant.Value max(Constant.Value other, boolean signed) {
+      ensure(type().isTrivialCastTo(other.type()),
+          "Max must be of same type, but other was %s",
+          other.type());
+
+      return this.geq(other, signed).intValue() == 0 ? this : other;
```
Jozott00: Has the same issue as above

```suggestion
      return this.geq(other, signed).bool() ? this : other;
```

Jozott00 (changes requested):
I think there are some logic bugs.
Please add tests to the `ConstantTests.java` for the added built-ins.

vadl/main/vadl/viam/Constant.java
flofriday: You are absolutely right they are wrong.

Jozott00 (approved):

PR #876, opened 2026-04-12
sys/ppc64: Optimize addi and addis instructions

kbschlosser:
Optimize addi and addis instructions, which currently run 2.4x slower than their QEMU counterparts.


AndreasKrall (approved):
looks fine to me

PR #877, opened 2026-04-13
docs: Add co-simulation section to documentation

Giftzwerg02:
Adds a simple user-documentation for the co-simulator, explaining how the co-simulator needs to be configured.


AndreasKrall (approved):
I approve it to make it available. But in many sentences there is a dot missing at the end. I do not know what cargo is and if it has to be installed.

Jozott00 (approved):
LGTM

PR #878, opened 2026-04-13
frontend: Add operation definition

flofriday:
The definition previously was already parsable but not included in the typechecker, lowering or VIAM.

Operations can be defined in the ISA or MIA and either explicitly or via the annotation.

### DIFF
--- a/vadl/main/vadl/ast/AnnotationTable.java
+++ b/vadl/main/vadl/ast/AnnotationTable.java
@@ -185,6 +185,17 @@ public class AnnotationTable {
         })
         .build();
 
+    annotationOn(InstructionDefinition.class, "operation",
+        () -> new IdentifersAnnotation(OperationDefinition.class))
+        .check((def, annotation, lowering) -> {
+          // Add the definition to the operation defined
+          annotation.identifiers.forEach(identifier -> {
+            var operation = (OperationDefinition) requireNonNull(identifier.target());
+            operation.instructions.add(def);
+          });
+        })
+        .build();
+
     annotationOn(AliasDefinition.class, "overwrite source",
         () -> new EnumAnnotation(List.of("zero", "sign")))
         .check((def, annotation, lowering) -> {
@@ -1265,6 +1276,125 @@ class EnumAnnotation extends Annotation {
     return "[ " + name + " : " + options + " ]";
   }
 }
+
+/**
+ * A simple annotation that stores a single Identifier.
+ *
+ * <p>Examples for such annotations:
+ * <pre>
+ * [ relatedInstr : ADDI ]
+ * </pre>
+ */
+class IdentiferAnnotation extends Annotation {
+  @LazyInit
+  Identifier identifier;
+
+  @Nullable
+  final Class<? extends Definition> targetClass;
+
+  public IdentiferAnnotation() {
+    super();
+    targetClass = null;
+  }
+
+  public IdentiferAnnotation(Class<? extends Definition> targetClass) {
+    super();
+    this.targetClass = targetClass;
+  }
+
+  @Override
+  void resolveName(AnnotationDefinition definition, SymbolTable.SymbolResolver resolver) {
+    verifyValuesCnt(definition, 1);
+    var firstValue = definition.values.getFirst();
+
+    if (!(firstValue instanceof Identifier identifier)) {
+      throw error("Invalid Annotation Argument", firstValue)
+          .locationDescription(firstValue, "Expected an identifier but got %s",
+              firstValue.getClass().getSimpleName())
+          .build();
+    }
+    this.identifier = identifier;
+
+    if (targetClass != null) {
+      definition.symbolTable().requireAs(identifier, targetClass);
+    } else {
+      definition.symbolTable().requireAs(identifier, Node.class);
+    }
+  }
+
+  @Override
+  void typeCheck(AnnotationDefinition definition, TypeChecker typeChecker) {
+    // Only typecheck expressions
+    if (targetClass != null && Expr.class.isAssignableFrom(targetClass)) {
+      typeChecker.check(identifier);
+    }
+  }
+
+  @Override
+  public String usageString() {
+    return "[ " + name + " : <Id> ]";
+  }
+}
+
+/**
+ * A simple annotation that stores a one or more Identifiers.
+ *
+ * <p>Examples for such annotations:
+ * <pre>
+ * [ relatedInstr : ADDI ADDU ]
+ * </pre>
+ */
+class IdentifersAnnotation extends Annotation {
+  List<Identifier> identifiers = new ArrayList<>();
+
+  @Nullable
+  final Class<? extends Definition> targetClass;
+
+  public IdentifersAnnotation() {
+    super();
+    targetClass = null;
+  }
+
+  public IdentifersAnnotation(Class<? extends Definition> targetClass) {
+    super();
+    this.targetClass = targetClass;
+  }
+
+  @Override
+  void resolveName(AnnotationDefinition definition, SymbolTable.SymbolResolver resolver) {
+    verifyValuesNonEmpty(definition);
+
+    for (var value : definition.values) {
+      if (!(value instanceof Identifier identifier)) {
+        throw error("Invalid Annotation Argument", value)
+            .locationDescription(value, "Expected an identifier but got %s",
+                value.getClass().getSimpleName())
+            .build();
+      }
+      identifiers.add(identifier);
+
+      if (targetClass != null) {
+        definition.symbolTable().requireAs(identifier, targetClass);
+      } else {
+        definition.symbolTable().requireAs(identifier, Node.class);
+      }
+    }
+  }
+
+  @Override
+  void typeCheck(AnnotationDefinition definition, TypeChecker typeChecker) {
+    // Only typecheck expressions
+    if (targetClass != null && Expr.class.isAssignableFrom(targetClass)) {
+      identifiers.forEach(typeChecker::check);
+    }
+  }
+
+  @Override
+  public String usageString() {
+    return "[ " + name + " : <Id>... ]";
+  }
+}
+
 
 /**
  * A annotation that holds a single expression. Used for more complex annotations.
--- a/vadl/main/vadl/ast/AnnotationTable.java
+++ b/vadl/main/vadl/ast/AnnotationTable.java
@@ -301,12 +301,14 @@ public class AnnotationTable {
           }
         }).build();
 
-    annotationOn(InstructionDefinition.class, "add operands", DefinitionRefAnnotation::new)
+    annotationOn(InstructionDefinition.class, "add operands",
+        () -> new IdentifersAnnotation(Definition.class))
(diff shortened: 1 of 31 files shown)


AndreasKrall:
In the parser also the operation annotation for instructions is missing. Here is an example from the TI64C spec:
```
[operation : match: Id ( $reg = A => D1; _=> D2)]
instruction ...
```
This results into one of the two the expanded operations:
```
[operation : D1]
[operation : D2]
instruction ...
```
The annotation syntax currently supports full macro expansion for expressions, so everything should work fine.
The EBNF syntax for the operation annotation is
```
"[" "operation" ":" identifier { "," identifier } "]"
```
As identifier is a subtype of expression everything should be fine.
So this has to be added to the frontend too.
Typechecking for the annotation should check that the identifiers are names of operations.

The typechecker for the operation definition should check the identifiers in the set are either instruction names or operation names.

So it has to be checked that for an operation name used in an operation annotation there is an operation definition (usually with an empty set).

flofriday:
@AndreasKrall I think the parser is fine, we just need to add the new annotation to the annotation table. ~~But I would prefer to leave that for a later PR and currently focus on getting at least the explicit non-annotation approach implemented so that Linus has _something_ to work with.~~ 

**Update:** Nevermind, I implemented the annotation approach as well.

AndreasKrall (approved):
Looks fine to me. As far as I understood your code, you only type check operations inside the ISA. Operation definitions also can occur in the MiA. In the parser it is already supported. Please check, if the type checker and VIAM generation also works inside the MiA.

vadl/main/vadl/dump/entities/DefinitionEntity.java
```diff
@@ -133,6 +134,7 @@ private int rank() {
       is(Instruction.class),
       is(PseudoInstruction.class),
       is(Relocation.class),
+      is(Operation.class),
```
flofriday, line 137: **question:** @Jozott00 can you explain if this makes sense or where else to put it?

vadl/main/vadl/dump/entitySuppliers/ViamEntitySupplier.java
```diff
@@ -145,6 +147,18 @@ public void visit(Encoding.Field encodingField) {
...
+  public void visit(Operation operation) {
+    // Operations should be displayed in the TOC if they are at ISA level
+    // or at MicroArchitecture level.
+    var entity = entityOf(operation);
+    Objects.requireNonNull(entity.parent());
+    if (!(entity.parent().origin() instanceof InstructionSetArchitecture)
+        && !(entity.parent().origin() instanceof MicroArchitecture)) {
+      replaceAsSubEntityOfParent(operation);
+    }
+  }
```
flofriday, line 160: **question:** @Jozott00 I _think_ this is right from the pattern but I don't really get why and when we are replacing Entities here.

AndreasKrall:
I think that the simple cases are working. I saw some code which I do not find anymore, where you wrote, that you do not save operation names in the instruction list of an operation as you already replace the operation name with the instructions it stands for.

But such a replacement algorithm is a little bit tricky and I couldn't find it in your code. I will present a more complicated operation definition, operation names start with the letter o followed by a number, instructions start with the letter i followed by a number:
```
operation o1 = {i1, o2}      // o1 = {i1, i2, i3, i4}
operation o2 = {i2, o3}      // o2 = {i1, i2, i3, i4}
operation o3 = {i3, o1, o4}  // o3 = {i1, i2, i3, i4}
operation o4 = {i4}          // o4 = {i4}
```
So recursion is allowed here and has a well defined semantics. You have to build the closure sets and have to iterate through the sets until no element is added to any operation. Currently we only support the union of operations.

We plan in the far future to add set difference and set intersection. We have to think about how this can be added with no problems with the semantics.

So we should discuss if the operation name resolving is already done in the frontend or if it should be done in a later pass at the VIAM level. In general I would prefer a resolution as early as possible.

I have the feeling, if we allow recursion, that we cannot add difference and intersection. I have to think more about it or find some proofs in the literature.

vadl/main/vadl/dump/entities/DefinitionEntity.java
Jozott00: Yes looks good to me

vadl/main/vadl/dump/entitySuppliers/ViamEntitySupplier.java
Jozott00: Yes, should be right. Essentially, everything that is in `entities` is considered to be part of the TOC (the side panel). However, in some cases we don't want it to be there, i.e. it is a more nested definition e.g. format fields, so we actually remove it from `entities` using `replaceAsSubEntityOfParent()`.

If those operations can be defined only in top-level, ISA or MicroArchitecture, you can actually omit the visitor implementation entirely, as it won't ever reach the `replaceAsSubEntityOfParent()`.

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -1266,6 +1277,125 @@ public String usageString() {
...
 
+/**
+ * A simple annotation that stores a single Identifier.
+ *
+ * <p>Examples for such annotations:
+ * <pre>
+ * [ relatedInstr : ADDI ]
+ * </pre>
+ */
+class IdentiferAnnotation extends Annotation {
```
Jozott00: **chore:** The identifier must reference something right? If yes, this should probably be mentioned in the comment.

Jozott00: **major:** Instead of creating this annotation class, could we use (and maybe extend) the `DefinitionRefAnnotation` instead? On a first look, it seems like they are very similar.

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -1266,6 +1277,125 @@ public String usageString() {
...
+
+/**
+ * A simple annotation that stores a one or more Identifiers.
+ *
+ * <p>Examples for such annotations:
+ * <pre>
+ * [ relatedInstr : ADDI ADDU ]
+ * </pre>
+ */
+class IdentifersAnnotation extends Annotation {
```
Jozott00, line 1243: **major:** I think we should not have both, a single identifier and multi identifier annotation, but instead just a multi identifier annotation, where the user may define the expected number of identifiers during construction of the annotation class.

vadl/test/resources/frontend-snapshots/annotation/invalidOperationAnnotationInvalidTarget.vadl
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+//  Reported Diagnostics:
+//
+//  error: Unknown OperationDefinition: "BranchOp"
+//       ╭──[test/resources/frontend-snapshots/annotation/invalidOperationAnnotationInvalidTarget.vadl:11:15]
+//       │
+//    11 │   [operation: BranchOp]
+//       │               ^^^^^^^^ No operationdefinition with this name exists.
```
Jozott00, line 28: **minor:** `operationdefinition` should `operation definition`

Jozott00 (approved):
LGTM in general, however, I think we can reduce the number of required annotation classes.

vadl/main/vadl/ast/AnnotationTable.java
Jozott00: So maybe we should collapse `IdentifierAnnotation, IdentifiersAnnotation and DefinitionRefAnnotation` into a single annotation class.

AndreasKrall:
> I think that the simple cases are working. I saw some code which I do not find anymore, where you wrote, that you do not save operation names in the instruction list of an operation as you already replace the operation name with the instructions it stands for.
> 
> But such a replacement algorithm is a little bit tricky and I couldn't find it in your code. I will present a more complicated operation definition, operation names start with the letter o followed by a number, instructions start with the letter i followed by a number:
> 
> ```
> operation o1 = {i1, o2}      // o1 = {i1, i2, i3, i4}
> operation o2 = {i2, o3}      // o2 = {i1, i2, i3, i4}
> operation o3 = {i3, o1, o4}  // o3 = {i1, i2, i3, i4}
> operation o4 = {i4}          // o4 = {i4}
> ```
> 
> So recursion is allowed here and has a well defined semantics. You have to build the closure sets and have to iterate through the sets until no element is added to any operation. Currently we only support the union of operations.
> 
> We plan in the far future to add set difference and set intersection. We have to think about how this can be added with no problems with the semantics.
> 
> So we should discuss if the operation name resolving is already done in the frontend or if it should be done in a later pass at the VIAM level. In general I would prefer a resolution as early as possible.
> 
> I have the feeling, if we allow recursion, that we cannot add difference and intersection. I have to think more about it or find some proofs in the literature.

Over the night I came to the conclusion that we should forbid recursion as in all other definitions and just add the recursion check as in all other definitions. It is not needed and a user can easier understand the set contents. Set difference is not monotonic.

Nevertheless Gemini says it should be possible to support recursion union and intersection, but not set difference. It says:
Fixed-Point Semantics: A recursive set definition often takes the form 𝑆 = ℱ (𝑆), where ℱ is an operation involving unions, intersections, and differences. Such equations are solved by finding a fixed point, typically the smallest set satisfying the definition.
Knaster-Tarski Theorem: If an operator 𝐹(𝑋) is monotonic (if 𝐴 ⊆ 𝐵, then 𝐹(𝐴) ⊆ 𝐹(𝐵)), it is guaranteed to have least and greatest fixed points.
Monotonicity Issues: While union and intersection are monotonic, set difference (or negation in set-theoretic types) is not necessarily monotonic. However, in many applications (like dataflow analysis or domain theory), recursive systems can be restructured or the definition restricted (e.g., only positive occurrences of the defined set) to ensure monotonicity.

flofriday:
> LGTM in general, however, I think we can reduce the number of required annotation classes.

I agree, there are way to many classes there, I don't know how, when this happened, I'll eventually look into that.

vadl/test/resources/frontend-snapshots/annotation/invalidOperationAnnotationInvalidTarget.vadl
flofriday: I agree, I _assume_ this will be a larger refactoring, since I didn't hardcode this name so I've created a sperate issue for that: #879

vadl/main/vadl/ast/AnnotationTable.java
flofriday: Same as the other comment, will answer detailed there

vadl/main/vadl/ast/AnnotationTable.java
flofriday: I've merged all three into `IdentifiersAnnotation`. You can recreate the DefinitionAnnotation by passing `Definition.class` to the constructor. But you can also pass something more specific. 

And if you only want to allow a single one there is a handy static method  `IdentifiersAnnotation.single()` to do that too. Don't know if it will be used but it wasn't too much hassle to add it.

Jozott00 (approved):

PR #880, opened 2026-04-15
lcb: make xaarch64 compilable / fix immediates/intrinsics

Tbswss:
The goal of this PR is to make the generated compiler from the xaarch64 spec compilable. Following changes were made:

- immediates have a minimum llvm-type with a bitwidth of the smallest register of the architecture
- the generated return-type of intrinsics is not hardcoded anymore
- redundant "zext" functions are no longer written in patterns


Tbswss:
The failing iss test case doesn't seem to be related to this merge request, as other pull requests fail with the same error.

kper (approved):

IhrWerbepartner:
This should in theory also remove the need for the `upcast access to` annotation in the RISCV spec.

PR #884, opened 2026-04-17
diagnostic: Improve diagnostic rendering

flofriday:
The file URI is no longer enclosed by brackets because recent IntelliJ versions could no longer open the link. (The last bracket was now parsed as part of the URI).

Previously in some cases there wasn't a newline between two errors, making them not that readable.

The "xy lines omitted" rendering for large error messages was wrongly indented.

Some coloring improvements, to better indicate which parts are important.
Screenshot 2026-04-17 at 12 01 09

Custom highlighting for text within backticks.
Screenshot 2026-04-17 at 12 23 39

### DIFF
--- a/vadl/main/vadl/error/DiagnosticPrinter.java
+++ b/vadl/main/vadl/error/DiagnosticPrinter.java
@@ -87,7 +87,10 @@ public class DiagnosticPrinter {
   private void toString(Diagnostic diagnostic, StringBuilder builder) {
     printHeader(diagnostic, builder);
     printMultiSourcePreview(diagnostic, builder);
-    builder.append(indentBy(messageBlock(diagnostic.messages), "     "));
+    if (!diagnostic.messages.isEmpty()) {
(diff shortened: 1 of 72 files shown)


vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -1779,8 +1779,20 @@ public Void visit(ImportDefinition importDefinition) {
   }
 
   @Override
-  public Void visit(ProcessDefinition processDefinition) {
-    throwUnimplemented(processDefinition);
+  public Void visit(ProcessDefinition definition) {
```
rascmatt: Did you intend to implement the checking of the process definition here? If you want to keep it in the commit it should be fine

rascmatt (commented):
Otherwise LGTM

vadl/main/vadl/ast/TypeChecker.java
flofriday: great catch, no I didn't 😅

rascmatt (approved):
Nice, lgtm

PR #885, opened 2026-04-17
cli: Add timing csv dump

flofriday:
We already have stdout printing, but for benchmarking this is much easier for computers to parse.

### DIFF
--- a/vadl-cli/main/vadl/cli/BaseCommand.java
+++ b/vadl-cli/main/vadl/cli/BaseCommand.java
@@ -100,6 +100,12 @@ public abstract class BaseCommand implements Callable<Integer> {
       description = "Print timings of the phases of the compiler")
   boolean showTimings;
 
+  @Option(names = "--timings-csv",
+      scope = INHERIT,
+      hidden = true,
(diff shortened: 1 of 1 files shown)


vadl-cli/main/vadl/cli/BaseCommand.java
```diff
@@ -100,6 +100,11 @@ public abstract class BaseCommand implements Callable<Integer> {
       description = "Print timings of the phases of the compiler")
   boolean showTimings;
 
+  @Option(names = "--timings-csv",
+      scope = INHERIT,
+      description = "Write pass timings to <output>/timings.csv")
+  boolean writeTimingsCsv;
```
Jozott00, line 107: **thought:** Maybe we should hide this option from `--help`

Jozott00 (approved):

vadl-cli/main/vadl/cli/BaseCommand.java
flofriday: Yes, I agree, we can still re-enable it if we ever find that it's useful for more people :)

Jozott00 (approved):

PR #888, opened 2026-04-20
iss: Fix >64 bit CPU reset and register dump

Jozott00:
This PR fixes a bug where the generated `reset` and register content dumping code was broken for registers >64 bit. 

This fixes compilation of the `aarch64/vprocessor.vadl` with `VL > 512`.
However, without any optimizations in place, the generation (i.e. the generated helpers) explode, which causes unusable linking times.


AndreasKrall (approved):
looks fine to me, but I didn't check the link times.

PR #889, opened 2026-04-20
lcb: remove upcast access to annotation

IhrWerbepartner:
With the correct constant node type generation merged in #880, there is no need for this annotation anymore.


AndreasKrall (approved):
looks fine to me.

PR #891, opened 2026-04-20
spec: Add hexagon spec

flofriday:
This adds a minimal version of the hexagon spec.
Many annotations, and definitions aren't implemented because they don't work.

**For the reviewer:** The original commit of this PR contains the hexagon spec with minimal changes from the original VADL implementation and the second then adapts it to compile with the new frontend. It might make sense to review only look over the second commit.

### DIFF
--- a/sys/hexagon/hexagon.vadl
+++ b/sys/hexagon/hexagon.vadl
@@ -81,7 +81,9 @@ instruction set architecture QDSP6 =
   alias register UTIMERHI        = C(31)   // Qtimer register (high)
 
   [next]
-  group counter PC : Address
+  // TODO: Reenable group counter
+  // group counter PC : Address
+  program counter PC : Address
 
   // Valid 64-byte pairs:
   // R1:0, R3:2, ... R29:28, R31:30
@@ -126,14 +128,17 @@ instruction set architecture QDSP6 =
       , 2 => P.P2(0) = 1
       , _ => P.P3(0) = 1
       }
-  process Pset ( reg: PredicateIndex, val: Byte ) = {
-    match (reg) with
-      { 0 => P.P0 := val
-      , 1 => P.P1 := val
-      , 2 => P.P2 := val
-      , 3 => P.P3 := val
-      }
-  }
+
+  // TODO: Reenable proces
+  // process Pset ( reg: PredicateIndex, val: Byte ) = {
+  // match (reg) with
+  //     { 0 => P.P0 := val
+  //     , 1 => P.P1 := val
+  //     , 2 => P.P2 := val
+  //     , 3 => P.P3 := val
+  //     }
+  // }
+  function Pset(reg: PredicateIndex, val: Byte) -> Word = 0
   function Pget ( reg: PredicateIndex ) -> Byte =
     match (reg) with
       { 0 => P.P0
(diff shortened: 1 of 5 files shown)


flofriday:
@rascmatt I'm sure there are some other errors in the encoding of this spec because `run --args="check sys/hexagon/hexagon.vadl"` crashes _somewhere_ in the VDT verification pass, while it tries to create a diagnostic.

rascmatt:
@flofriday 
Thank you for pointing it out, there was a bug in the decoding logic which tries to find the overlapping instructions from the counterexample returned by the SMT solver. 

I fixed the error, so it correctly constructs the diagnostic, however I'm still looking into the overlapping encoding. Concretely, the instructions

`JumpReg` (`BranchRegFormat`) and `CallImmPredPos` (`CallImmPredFormat`) overlap (from the point of the encoding declaration alone), so e.g. the encoding `0x882a8052` matches both definitions.

However, from a VLIW standpoint they could still be unique, depending on where in the VLIW bundle slots they are allowed to appear. So I'll have to analyze that.

flofriday:
Awesome thanks for fixing that :)
Is there somewhere where we should add some tests to save ourselfs against regression?

rascmatt:
@flofriday 

So I had a closer look at the Hexagon VADL Spec, and compared it with the encoding definition in the [Hexagon Reference Manual](https://docs.qualcomm.com/doc/80-N2040-33/80-N2040-33_REV_E_Qualcomm_Hexagon_V60_V61_ProgrammerS_Reference_Manual.pdf):

First of all, according to the `group VLIW` specification (and the manual) the `J` and `JR` type instructions can appear in the same VLIW slot, so this should not a problem with the fact that the decoder does not respect the VLIW stuff yet.

It seems to me like the VADL spec for the `CallImmPredPos` and `CallImmPredNeg` does not specify the correct opcode, which should be `0b11010`. I changed it as follows:

https://github.com/OpenVADL/openvadl/blob/9a9d0b7692eadb659b2e3e710901da7b5c1989a5/sys/hexagon/hexagon.vadl#L727-L730

(also I checked the encoding of the `JumpReg` instruction, which looks fine to me)

@AndreasKrall Could you please cross check this change?

Now the semantic encoding verification succeeds, but I doubt we have a way to run any regression tests (e.g. running the old simulator) to test the behavior of those instructions specifically. Or did you have any other regression tests in mind (@flofriday)?

AndreasKrall:
> @flofriday
> 
> So I had a closer look at the Hexagon VADL Spec, and compared it with the encoding definition in the [Hexagon Reference Manual](https://docs.qualcomm.com/doc/80-N2040-33/80-N2040-33_REV_E_Qualcomm_Hexagon_V60_V61_ProgrammerS_Reference_Manual.pdf):
> 
> First of all, according to the `group VLIW` specification (and the manual) the `J` and `JR` type instructions can appear in the same VLIW slot, so this should not a problem with the fact that the decoder does not respect the VLIW stuff yet.
> 
> It seems to me like the VADL spec for the `CallImmPredPos` and `CallImmPredNeg` does not specify the correct opcode, which should be `0b11010`. I changed it as follows:
> 
> https://github.com/OpenVADL/openvadl/blob/9a9d0b7692eadb659b2e3e710901da7b5c1989a5/sys/hexagon/hexagon.vadl#L727-L730
> 
> (also I checked the encoding of the `JumpReg` instruction, which looks fine to me)
> 
> @AndreasKrall Could you please cross check this change?
> 
> Now the semantic encoding verification succeeds, but I doubt we have a way to run any regression tests (e.g. running the old simulator) to test the behavior of those instructions specifically. Or did you have any other regression tests in mind (@flofriday)?

Yes, `0b11010` is correct. There is a QEMU hexagon. But I don't think that it is easy to just compare the decoders and I doubt that our cosimulator can handle VLIW. But @Giftzwerg02 plans to do project/master thesis.

AndreasKrall (approved):
Looks fine to me.

rascmatt (approved):
LGTM

PR #892, opened 2026-04-20
frontend: Add current and nextnext program counter subcalls

flofriday:
These subcalls were long overdue and are now really needed to get hexagon running.

### DIFF
--- a/vadl/main/vadl/ast/BehaviorLowering.java
+++ b/vadl/main/vadl/ast/BehaviorLowering.java
@@ -1327,8 +1327,12 @@ class BehaviorLowering implements StatementVisitor<SubgraphContext>, ExprVisitor
           int offset = 0;
           for (var subcall : expr.subCalls) {
             var subcallName = subcall.identifier().name;
-            if (subcallName.equals("next")) {
+            if (subcallName.equals("current")) {
+              offset = 0;
+            } else if (subcallName.equals("next")) {
               offset += instrWidthInByte;
+            } else if (subcallName.equals("nextnext")) {
+              offset += instrWidthInByte * 2;
             } else {
               throw new IllegalStateException("unknown subcall: " + subcallName);
             }
--- a/vadl/main/vadl/ast/TypeChecker.java
+++ b/vadl/main/vadl/ast/TypeChecker.java
@@ -3511,7 +3511,7 @@ public class TypeChecker
           addErrorAndStopChecking(error("Invalid counter sub-call", expr)
               .locationDescription(expr, "Cannot do sub call and slice on counter.").build());
         }
-        var validSubCalls = List.of("next");
+        var validSubCalls = List.of("current", "next", "nextnext");
         if (expr.subCalls.size() != 1) {
           addErrorAndStopChecking(error("Invalid counter sub-call", expr)
               .locationDescription(expr, "Only a single sub-call expected.").build());
--- a/vadl/test/resources/frontend-snapshots/lowering/readProgramCounterStates.vadl
+++ b/vadl/test/resources/frontend-snapshots/lowering/readProgramCounterStates.vadl
@@ -1,0 +1,170 @@
+// INCLUDE-VIAM-DUMP
+
+// This test verifies that the subcalls of a program counter can be read
+
+instruction set architecture TEST = {
+
+  register X: Bits<5><8>
+  program counter PC: Bits<8>
+
+  format Frmt : Bits<8> = { a: Bits<8> }
+
+  model ReadPC(name: Id, readExpr: Ex): IsaDefs = {
+    instruction $name : Frmt = { X(0) := $readExpr}
+    assembly $name = AsStr($name)
+    encoding $name = { a = 1 }
+  }
+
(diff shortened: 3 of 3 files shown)


vadl/main/vadl/ast/BehaviorLowering.java
```diff
@@ -1327,8 +1327,12 @@ private ExpressionNode visitSubCall(CallIndexExpr expr, ExpressionNode exprBefor
           int offset = 0;
           for (var subcall : expr.subCalls) {
             var subcallName = subcall.identifier().name;
-            if (subcallName.equals("next")) {
+            if (subcallName.equals("current")) {
+              offset = 0;
+            } else if (subcallName.equals("next")) {
               offset += instrWidthInByte;
+            } else if (subcallName.equals("nextnext")) {
+              offset += instrWidthInByte * 2;
```
rascmatt, line 1335: This currently assumes fixed instruction widths, right? 

Given we know the format of the current instruction, the `next` offset should be easily computable, even for dynamic insn widths. However for `nextnext` I'm not sure how we can determine that statically? I suppose one solution to that could be to disallow `nextnext` if the insn width is not constant.

Also, is the offset really fixed to width in bytes? Or should it depend on the range of the `memory` definition? E.g. the following defines byte-addressable memory, but in theory that could be something else.
```vadl
memory MEM : Addr  -> Byte
```
But I suppose that using differently-sized memory steps (particularly non-byte divisible) is going to break much more than just that. 

For the Hexagon Spec the latter should be fine as is as it uses Byte-addressable memomry. 

However I don't think we should hardcode the instruction width: While Hexagon doesn't use `nextnext`, it does define two different format sizes, `Word` (32-bit) and `Double` (64-bit).

flofriday: I agree. 👍
Unfortunately this isn't quite as simple so I created https://github.com/OpenVADL/openvadl/issues/895 to track that.

Jozott00: > This currently assumes fixed instruction widths, right?

I think, this problem might also occure for `.next` if the `program counter` definition is annotated with `[ next ]` or `[ next next ]`. In that case, the actually returned address would be `next next` and `next next next` respectively lol. 

I guess we should disallow all combinations that result in an instruction address prediction that requires more than the current instruction width iff not all instructions have the same width.

Jozott00 (changes requested):
Looks good, but we should add checks for the disallowed combinations of `.next`, `.nextnext`, `[ next ]` and `[ next next ]`, as well as respective diagnostic tests.

vadl/main/vadl/ast/BehaviorLowering.java
AndreasKrall: @Jozott00 Your assumption about the semantics is wrong. There is nothing like a triple next. The builtin member methods are always relative to the current instruction. The size of the instructions in a nextnext must be the same. @flofriday, @Jozott00 and @rascmatt please reread 2.6.1 Program Counter declarations in the language tutorial and if it can be misunderstood please give ideas how to improve the presentation.

AndreasKrall:
> Looks good, but we should add checks for the disallowed combinations of `.next`, `.nextnext`, `[ next ]` and `[ next next ]`, as well as respective diagnostic tests.

There are no invalid combinations, see tutorial 2.6.1

vadl/main/vadl/ast/BehaviorLowering.java
AndreasKrall: > This currently assumes fixed instruction widths, right?
> 
> Given we know the format of the current instruction, the `next` offset should be easily computable, even for dynamic insn widths. However for `nextnext` I'm not sure how we can determine that statically? I suppose one solution to that could be to disallow `nextnext` if the insn width is not constant.
> 
> Also, is the offset really fixed to width in bytes? Or should it depend on the range of the `memory` definition? E.g. the following defines byte-addressable memory, but in theory that could be something else.
> 
> ```
> memory MEM : Addr  -> Byte
> ```
> 
> But I suppose that using differently-sized memory steps (particularly non-byte divisible) is going to break much more than just that.
> 
> For the Hexagon Spec the latter should be fine as is as it uses Byte-addressable memomry.
> 
> However I don't think we should hardcode the instruction width: While Hexagon doesn't use `nextnext`, it does define two different format sizes, `Word` (32-bit) and `Double` (64-bit).

An instruction length always must be a multiple of the instruction memory element size which is not necessarily a byte. We can define multiple memories, e.g.:
```
using Addr = Bits<24>
using Data = Bits<12>
using Inst = Bits<24>

[data]
memory DMem : Addr -> Data
[instruction]
memory IMem : Addr -> Inst
```
Where we could have 24 bit and 48 bit instructions for example. Such nice things are available in DSP architectures.

As far as I know the `[data]` and `[instruction]` annotation are currently not implemented, we only support Neumann architectures.

AndreasKrall (approved):
Is fine as a short term implementation. Determining the length of the current instruction will be covered in #895.

PR #893, opened 2026-04-20
CLI: Add spec stats

flofriday:
We mostly only need this for publications where we _approximate_ the size of specs based on the instruction count, lines of code etc.

**Note for the reviewer:** This PR also includes a commit that removes the unused `IOException` from the VFS. We must have missed that previously, but I wanted to do that now because otherwise the new code would have also had to include that exception.


Jozott00 (approved):
LGTM

flofriday:
❤️‍🩹 Just some formatting fixes...

Jozott00 (approved):

PR #897, opened 2026-04-21
spec: Add PC subcalls to the hexagon spec

flofriday:
This is only now possible since #892 was merged.


AndreasKrall (approved):
As interim test it is fine. But the .current implementation of a group counter is more complex than of a program counter and for a group counter there will be no nextnext.

PR #898, opened 2026-04-21
iss: Optimize assignments of forall tensor expressions 

Jozott00:
For instructions that compute a vector result as tensor expression and assign the expression cross-lane to the vector register like in AArch64 SVE:
```
Z(zdn) := forall i : IndexE in 0..NElems-1
                  tensor if pred(i)
                    then VADL::add(left(i), right(i))
                    else left(i)
```
we can transform it into an equivalent `forall do` statement:
```
forall i : IndexE in 0..NElems-1 do
   Z(zdn)(i) := if pred(i)
       then VADL::add(left(i), right(i))
       else left(i)
```

This allows the application of optimization as for normal `forall do` statements. Specifically, the forall unroll prevention that was merged in #896, combined with this transformation, allows the rendering of tensor expressions as `for` loops in C.

To achieve this, this PR also enhances the analysis performed by the `SafeResourceReadPass`. It now considers constant read/write chunk windows and recognizes whether two distinct register regions are accessed independently during different iterations.

### DIFF
--- a/vadl/main/vadl/iss/codegen/IssProcGen.java
+++ b/vadl/main/vadl/iss/codegen/IssProcGen.java
@@ -113,13 +113,9 @@ abstract class IssProcGen implements CDefaultMixins.All,
         "Helper read preload expects <=64-bit reads, got %d-bit on %s", width, read);
     var valueType = CppTypeMap.nextFittingUInt(read.type().asDataType());
     var name = readRegVariable(read);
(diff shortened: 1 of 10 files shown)


AndreasKrall (approved):
Looks fine to me. It is good to have this as an option, but I believe that the expression form is better for QEMU vector instructions. But you can choose the variant you need depending on heuristics. The ISA invocation has many parameters, a comment to each parameter would help to understand it.

PR #900, opened 2026-04-21
iss: Add bi-endian annotation for memory definition

BadGraphixD:
- Added the annotation [ littleEndian : Condition ]
  - It cannot be applied together with [ bigEndian ]
- VIAM memory operations now have a field specifying whether to read/write in reverse byte order
- Iss load/store nodes now have an endianness
- The ApplyMemoryEndiannessPass checks the endianness of the memory definition of reads/writes and changes the memory operations/inserts logic accordingly. It also verifies that the bi-endian condition only reads tb-saved register bits
- Instruction loading uses the bi-endian condition of the memory definition
- ppc64.vadl spec is now bi-endian instead of big-endian


vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -218,12 +220,25 @@ public class AnnotationTable {
         })
         .build();
 
-    annotationOn(MemoryDefinition.class, "bigEndian", EnableAnnotation::new)
-        .applyViam((def, annotation, lowering) -> {
-          var viamDef = (Memory) def;
-          viamDef.addAnnotation(new BigEndianAnnotation());
+    groupOn(MemoryDefinition.class)
+        .add("bigEndian", EnableAnnotation::new)
+        .add("littleEndian", ExprAnnotation::new)
```
Jozott00: **minor:** Lets change the name to `big endian` and `little endian`

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -218,12 +220,25 @@ public class AnnotationTable {
...
+        .applyViam(ctx -> {
+          var memDef = ctx.viamDef(Memory.class);
+          ctx.get("bigEndian").ifPresent(ann -> memDef.addAnnotation(new BigEndianAnnotation()));
+          ctx.get("littleEndian", ExprAnnotation.class).ifPresent(ann -> {
+            var graph = new BehaviorLowering(ctx.lowering).getFunctionGraph(ann.expr,
+                memDef.simpleName() + " Little Endian Condition");
+            graph.setParentDefinition(memDef);
+            memDef.setLittleEndianCondition(graph);
+          });
+        }).build();
```
Jozott00, line 248: **question:** Do I see it right, that right now, only `little endian` is able to set a condition?

vadl/main/vadl/viam/Memory.java
```diff
@@ -84,6 +90,22 @@ public ConcreteRelationType relationType() {
     return Type.concreteRelation(addressType, resultType);
   }
 
+  @Nullable
+  public Graph littleEndianCondition() {
```
Jozott00: **minor:** Add documentation to clarify that the graph is a function graph (without params)

vadl/main/vadl/iss/passes/IssRegisterAccessLoweringPass.java
```diff
@@ -97,6 +102,10 @@ public PassName getName() {
   public @Nullable Object execute(PassResults passResults, Specification viam) throws IOException {
     tcgInstrs(viam).forEach(i ->
         new IssStaticRegisterAccessConverter(i.behavior()).run());
+    viam.isa().get().ownMemories().stream()
+        .map(Memory::littleEndianCondition)
+        .filter(Objects::nonNull)
+        .forEach(c -> new IssStaticRegisterAccessConverter(c).run());
     ViamUtils.findAllBehaviors(viam).forEach(behavior ->
```
Jozott00, line 109: **major:** This shouldn't be necessary as `Memory` now implements `DefProp.WithOptionalBehavior`

vadl/main/vadl/viam/passes/ApplyMemoryEndiannessPass.java
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  private void checkMemoryEndiannessConditions(Specification viam) {
+    var errors = new ArrayList<Diagnostic>();
+    viam.isa().ifPresent(isa -> isa.ownMemories().forEach(mem -> {
+      var condition = mem.littleEndianCondition();
+      if (condition == null) {
+        return;
+      }
+      condition.getNodes(WriteResourceNode.class).findAny()
```
Jozott00: **minor:** This shouldn't be possible because first, the used annotation only allows expressions and second, the behavior lowering returns a function graph, which can't have any side effect by construction

vadl/main/vadl/viam/passes/ApplyMemoryEndiannessPass.java
```diff
@@ -0,0 +1,259 @@
...
+                          Memory memory) {
+    var reg = read.regTensor();
+    var resetValue = resetVector.get(reg);
+    if (resetValue == null) {
+      throw error(String.format("""
+          Value of register %s is unknown
+          """, reg.simpleName()),
+          read
+      ).description("""
+          The memory written here is be-endian. What endianness is used depends on the \
```
Jozott00: **minor: ** Typo
```suggestion
          The memory written here is bi-endian. What endianness is used depends on the \
```

vadl/main/vadl/viam/passes/ApplyMemoryEndiannessPass.java
```diff
@@ -0,0 +1,259 @@
...
+          value of register %s, which is not known at this point. Only registers initialized \
+          to a constant value in the default branch of the cpu reset procedure can be \
```
Jozott00: **minor:**
```suggestion
          to a constant value in the default branch of the processor reset procedure can be \
```

vadl/main/vadl/viam/passes/ApplyMemoryEndiannessPass.java
```diff
@@ -0,0 +1,259 @@
...
+          statically inferred.
+          """, reg.simpleName()
+      ).locationHelp(
+          reset,
+          "Cpu reset procedure"
```
Jozott00: **minor:**

```suggestion
          "Processor reset procedure"
```

vadl/main/vadl/viam/passes/ApplyMemoryEndiannessPass.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+    var condition = getConditionCopy(write.memory());
+    if (condition == null) {
+      // leave write as is since the memory is not bi-endian
+      return;
+    }
+    var next = GraphUtils.getSingleUsage(write, AbstractEndNode.class);
+    next.removeSideEffect(write);
+    var prev = requireNonNull(next.predecessor());
+    prev.setNext(null);
```
Jozott00: **minor:** Not important, but there is also the `.unlinkNext()` method

vadl/main/vadl/viam/graph/dependency/ReadMemNode.java
```diff
@@ -39,6 +39,12 @@ public class ReadMemNode extends ReadResourceNode {
   @DataValue
   protected int words;
 
+  /**
+   * Whether to read with big-endian byte order.
+   */
+  @DataValue
+  protected boolean reverseBytes;
```
Jozott00: **major:** I don't like that it is called `reverseBytes`, as this always assumes the `Memory` being `little endian`. 

I think we should add a VIAM endianess enum and instead define an `overwrittenEndianness` (or similiar) that is nullable. Then we can just determine the endianness explicitly, which makes it easier to reason about.

vadl/main/vadl/viam/graph/dependency/ReadMemNode.java
```diff
@@ -69,6 +93,12 @@ public int readBitWidth() {
     return memory.wordSize() * words;
   }
 
+  /**
+   * Whether to read with big-endian byte order.
+   */
+  public boolean reverseBytes() {
```
Jozott00: **major:** Same as above

vadl/main/vadl/viam/Memory.java
```diff
@@ -18,21 +18,27 @@
...
+import vadl.viam.graph.Graph;
 
 /**
  * The Memory class represents a VADL memory definition.
  *
  * <p>It always takes an address type and result type. The result type
  * specifies the word size of the memory.</p>
  */
-public class Memory extends Resource {
+public class Memory extends Resource implements DefProp.WithOptionalBehaviour {
```
Jozott00, line 33: **major:** Now with this PR in place, I think it might be cleaner to add an explicit `endianness` property, which is an enum (the same as mentioned in an other comment). By default it is `littleEndian`, and if the big endian annotation is set, the annotation handler applies the change to the `Memory` definition. 

So we don't need the `BigEndianAnnotatoin` in the VIAM anymore. What do you think?

Jozott00 (changes requested):
Very nice!
Just little things to discuss.

vadl/main/vadl/ast/AnnotationTable.java
BadGraphixD: Yes, basically directly copied from here: https://openvadl.github.io/openvadl/tutorial.html#memory_declaration

Should both be able to have a condition?

vadl/main/vadl/iss/passes/IssRegisterAccessLoweringPass.java
BadGraphixD: Here I am using two different converters:
- IssStaticRegisterAccessConverter
  - only used for tcg instructions and the memory condition
- IssRegisterAccessLowering
  - used for all behaviours

vadl/main/vadl/ast/AnnotationTable.java
Jozott00: Yes I think so, I guess the tutorial were just example usages. 
@AndreasKrall should it be possible to define both `little endian` and `big endian` with a condition?

vadl/main/vadl/iss/passes/IssRegisterAccessLoweringPass.java
Jozott00: Ahh yes true, my fault

vadl/main/vadl/ast/AnnotationTable.java
BadGraphixD: Both have an optional condition now.

vadl/main/vadl/iss/template/target/EmitIssTranslateCPass.java
```diff
@@ -136,6 +141,20 @@ private String genRegInitCode(Specification specification) {
     return sb.toString();
   }
 
+  private String renderLittleEndianCondition(Specification specification) {
```
Jozott00: **minor:** Name should be `renderBiEndianCondition` now

Jozott00 (approved):
Nice!

vadl/main/vadl/ast/AnnotationTable.java
AndreasKrall: Yes both should have a condition, it makes the usage easier. Of course it could be enough just to have little endian as default and as only annotation have big endian with a condition.

AndreasKrall (approved):
looks fine to me

PR #901, opened 2026-04-21
diagnostic: Deflate similar diagnostics into one

### LINKED ISSUE #890: frontend: Error deflating
If a faulty macro is invoked multiple times, the corresponding error is emitted for each invocation. 

Example:
```vadl
model badboi(name: Id): Defs = {
  constant $name: Bits<32> = "Not a bits"
}

$badboi(apfel)
$badboi(strudel)
$badboi(meister)
```

Produces:
```
     ╭── file:///Users/flo/src/open-vadl/example.vadl:2:30
     │
   2 │   constant $name: Bits<32> = "Not a bits"
     │                              ^^^^^^^^^^^^ Expected Bits<32> but got String.
     ⋮
   5 │ $badboi(apfel)
     │ -------------- from this model invocation
     │

error: Type Mismatch
     ╭── file:///Users/flo/src/open-vadl/example.vadl:2:30
     │
   2 │   constant $name: Bits<32> = "Not a bits"
     │                              ^^^^^^^^^^^^ Expected Bits<32> but got String.
     ⋮
   6 │ $badboi(strudel)
     │ ---------------- from this model invocation
     │

error: Type Mismatch
     ╭── file:///Users/flo/src/open-vadl/example.vadl:2:30
     │
   2 │   constant $name: Bits<32> = "Not a bits"
     │                              ^^^^^^^^^^^^ Expected Bits<32> but got String.
     ⋮
   7 │ $badboi(meister)
     │ ---------------- from this model invocation
     │
```

While we _somteimes_ need to to do this as the only the invocation can cause a typing issue, we should be able to detect when the only difference is in the location this macro got invoced in and compress them somehow into one diagnostic.
I don't yet have an idea how this should get printed, maybe we can look into how other compilers solve that.

### LINKED ISSUE #894: frontend: Compact macro traces
Long macro chains can get quite unreadable with the current diagnostic printing. Maybe we can take some inspiration from stack traces from other languages to improve their readability.

flofriday:
If two diagnostics only differ in the location they got expanded from, then they will now be collapsed into a single diagnostic with two macro backtraces attached to it.

The backtraces are now also more compact.

Closes: #894 
Closes: #890

### DIFF
--- a/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
+++ b/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
@@ -273,7 +273,7 @@ public class VadlTextDocumentService implements TextDocumentService {
       } catch (DiagnosticList dl) {
         log.debug("Raw diagnostics ({}): {}", document.uri, dl.getMessage());
         List<String> importedFileErrors = new ArrayList<>();
-        for (vadl.error.Diagnostic item : dl.items) {
+        for (vadl.error.Diagnostic item : dl.deflateSimilar().items) {
           Path itemPath = item.multiLocation.primaryLocation().location().path();
           if (!Objects.equals(itemPath, path)) {
             if (itemPath == null) {
--- a/vadl/main/vadl/error/Diagnostic.java
+++ b/vadl/main/vadl/error/Diagnostic.java
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText : © 2025 TU Wien <vadl@tuwien.ac.at>
+// SPDX-FileCopyrightText : © 2025-2026 TU Wien <vadl@tuwien.ac.at>
 // SPDX-License-Identifier: GPL-3.0-or-later
 //
 // This program is free software: you can redistribute it and/or modify
@@ -16,10 +16,12 @@
 
 package vadl.error;
 
+import com.google.common.collect.Streams;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Supplier;
 import org.jetbrains.annotations.Contract;
+import vadl.utils.Pair;
 import vadl.utils.SourceLocation;
 import vadl.utils.WithLocation;
 
@@ -45,6 +47,12 @@ import vadl.utils.WithLocation;
  *   3 │ constant flo = 18                      // A secondary location of the multilocation
  *     │          --- First definition          // Which also has a message
  *     │
+ *     ├─ From these 2 model invocation (outermost call first): // This code was generated from a
+ *     │    1) file.vadl:11:2                   // The first incocation started here
+ *     │       file.vadl:9:4                    // Then called the actual model here
+ *     │    2) file.vadl:14:2                   // A second invocation started here
+ *     │
+ *     │
  *     Each variable can only be declared once. // Messages can also belong to the whole diagnostic
  *     help: Find another awesome name.
  * </pre>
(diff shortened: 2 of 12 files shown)


vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -273,7 +273,7 @@ private void publishDiagnosticsForOneDocument(Document document,
       } catch (DiagnosticList dl) {
         log.debug("Raw diagnostics ({}): {}", document.uri, dl.getMessage());
         List<String> importedFileErrors = new ArrayList<>();
-        for (vadl.error.Diagnostic item : dl.items) {
+        for (vadl.error.Diagnostic item : dl.deflateSimilar().items) {
```
flofriday, line 276: **question:** @jaraonthe-dot-net is this clear how this works? 

I can also leave this change for you but in specs with a lot of invokes it's really helpful if only one diagnostic is reported

vadl/main/vadl/error/Diagnostic.java
```diff
@@ -122,6 +137,27 @@ public static void ensure(boolean condition, Supplier<DiagnosticBuilder> builder
     }
   }
 
+  /**
+   * Diagnostics that only differ in the macro trace will get collapsed into one.
+   *
+   * @param diagnostics to deflate.
+   * @return the deflated diagnostics.
+   */
+  public static List<Diagnostic> deflateSimilar(List<Diagnostic> diagnostics) {
```
flofriday: **question:** Do you have a better suggestion for a name here?

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
Jozott00: I am not sure if we should report only one location in the IDE. I don't think this is commonly done by other languages, and I think it might confuse people that one invocation causes an error while the other one apparently doesn't. 

So I think every location should be marked in the IDE, although when hovering, only one error message should be shown (I think currently the shown error message is replicated n times).

vadl/main/vadl/error/Diagnostic.java
Jozott00: I am fine with this one. An alternative would be `collapseSimilar`.

Jozott00 (approved):
LGTM

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
flofriday: I totally agree, however I don't think this relates to this code change.

At the moment the LS only reports the primary location, ignoring all secondary locations and expandedFrom traces. This PR doesn't change that. I've created #902 to track that progress.

If however, there is one model with an error and multiple invocations, the multiple diagnostics reported by for example the typechecker that have the same primary location and messages are now collapsed into one diagnostic. So yes only one message is now shown. The macroTraces are now (as they were before) still ignored.

vadl/main/vadl/error/Diagnostic.java
jaraonthe-dot-net: **Opinion**: I'm really not a fan of "deflate", as it is very vague. Better: `deduplicateSimilar`, `combineSimilar`, or the already suggested `collapseSimilar`.
I'd say "combine" is the most precise term here.

vadl/main/vadl/error/Diagnostic.java
```diff
@@ -59,6 +61,12 @@ public class Diagnostic extends RuntimeException {
   public final MultiLocation multiLocation;
   public final List<Message> messages;
 
+  /**
+   * The same diagnostic can be created from multiple macro invocations. They might get collapsed
+   * into one diagnostic, in which case this field is used to store them all.
+   */
+  public final List<List<SourceLocation>> macroTraces;
```
jaraonthe-dot-net, line 123: Suggestion: Please describe in the javadoc how this data structure works; i.e. what does the order of locations mean, aka which location is listed first?

jaraonthe-dot-net (commented):
Sorry to interject here, but I have two comments I'd like you to consider.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
jaraonthe-dot-net: Before this PR, your example from #890 would cause the LSP to deliver the same diagnostic message for the same location three times. So this is definitely an improvement.

I'm fine with this change to the language server.

Jozott00 (approved):

PR #905, opened 2026-04-22
lcb: Deduplicate generated predicate functions

Tbswss:
The goal of this PR is to reduce the number of duplicate predicate functions which are generated for the tablgen-immediate-records, enabling the generated aarch64 compiler to be compilable. This is done by reusing existing predicate-functions (and their identifiers) of other immediate-records, which have the same behaviour.

This is necessary, as LLVM apparently has an implicit limit on the number of predicate functions it supports (at most 256). LLVM generates a switch-case statement for all predicates, where each case is identified by a given index. These indices are stored in another generated "MatcherTable", which only supports "unsigned chars", so a maximum of 256 different predicates. Unfortunately, I did not find any reference for this limit.

For reference, where the aarch64 generated compiler had 505 predicate-functions before this change, it now has 21.

Changes to the implementation:
- The pass "GenerateTableGenImmediateRecordPass" still generates an immediate record for each found immediate, but reuses predicate-functions of other immediates for new immediates, by recognising identical behaviour
- "CreateFunctionsFromImmediatePass" was adapted to only generate cpp-predicate-functions for unique predicate-functions, not for every immediate.
- "TableGenImmediateRecord"s are no longer created by classes like "LlvmFieldAccessRef" or similar, but passed via the constructor
  - this caused changes to the `LlvmInstructionLoweringStrategy` and all subclasses, which make up most of the changes
- The "PredicateCodeGenerator..." test cases were adapted
  - these partially used to the "CreateFunctionsFromImmediates" pass to generate the predicates, but still regenerated the predicate-functions, which caused an error.
- Snapshots and similar were adapted

### DIFF
--- a/vadl/main/vadl/lcb/passes/llvmLowering/immediates/GenerateTableGenImmediateRecordPass.java
+++ b/vadl/main/vadl/lcb/passes/llvmLowering/immediates/GenerateTableGenImmediateRecordPass.java
@@ -19,17 +19,14 @@ package vadl.lcb.passes.llvmLowering.immediates;
 import static vadl.utils.GraphUtils.getSingleNode;
 import static vadl.viam.ViamError.ensurePresent;
 
+import com.google.common.collect.Streams;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-
 import javax.annotation.Nullable;
-
-import com.google.common.collect.Streams;
-
 import vadl.configuration.GeneralConfiguration;
 import vadl.cppCodeGen.CppTypeMap;
 import vadl.error.Diagnostic;
@@ -42,6 +39,7 @@ import vadl.pass.PassResults;
 import vadl.types.BitsType;
 import vadl.types.BoolType;
 import vadl.utils.Pair;
+import vadl.viam.Format;
 import vadl.viam.Function;
 import vadl.viam.Identifier;
 import vadl.viam.Instruction;
@@ -72,9 +70,12 @@ public class GenerateTableGenImmediateRecordPass extends Pass {
   }
 
 
+  /**
+   * Output of the pass.
+   */
   public record Output(
-    List<TableGenImmediateRecord> immediates,
-    Map<Function, TableGenImmediateRecord> immediatesByPredicates
+      List<TableGenImmediateRecord> immediates,
+      Map<Format.FieldAccess, TableGenImmediateRecord> immediatesByFieldAccess
   ) {}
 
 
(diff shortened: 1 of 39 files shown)


vadl/main/vadl/lcb/passes/llvmLowering/immediates/GenerateTableGenImmediateRecordPass.java
```diff
@@ -68,6 +97,9 @@ public List<TableGenImmediateRecord> execute(PassResults passResults,
             GenerateTableGenRegistersPass.class));
     var smallestRegisterClassType = generateTableGenRegistersPassOutput.smallestRegisterClassType();
 
+    var predicateMethodIdentifiers = new HashMap<Function, Identifier>();
+    var immediatesByKey = new HashMap<ImmediateKey, TableGenImmediateRecord>();
```
kper, line 101: major: wrong kind of map

Please use https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/IdentityHashMap.html

vadl/main/vadl/lcb/passes/llvmLowering/immediates/GenerateTableGenImmediateRecordPass.java
```diff
@@ -107,10 +139,32 @@ public List<TableGenImmediateRecord> execute(PassResults passResults,
                 upcastedValueType = upcastedValueTypeBitwidth < smallestRegisterClassBitwidth
                     ? smallestRegisterClassType
                     : upcastedValueType;
-                immediates.add(new TableGenImmediateRecord(instruction,
-                    fieldAccess,
-                    upcastedValueType));
 
+                // do not create a predicate method for every immedate record,
```
kper, line 143: major: I am not that convinced that you need to do this. Why isnt it enough to do it for every format and then reuse it?

kper (changes requested):

vadl/main/vadl/lcb/passes/llvmLowering/immediates/GenerateTableGenImmediateRecordPass.java
Tbswss: The advantage of this should be, that regardless of the format the field-access is of, predicate functions with identical behaviour will only be generated once, whereas generating the predicate functions for all formats will probably (or at least its possible) yield some duplicate generated predicate functions.

But it should suffice to generate the predicate functions for every format and then reuse these, and it is probably cleaner than the current solution.

Tbswss:
I am closing this PR in favour of https://github.com/OpenVADL/openvadl/pull/910

vadl/main/vadl/lcb/passes/llvmLowering/immediates/GenerateTableGenImmediateRecordPass.java
Tbswss: Reusing the predicate functions per format is actually cleaner and way less intrusive, I have created a new PR.

PR #907, opened 2026-04-22
Refactor CLI decoder options

rascmatt:
Refactored the decoder options a bit in general, since not all options make sense for all backends/commands.

- Introduce `skip=all` option to disable all decoder passes for the `check` command.
- restrict the scope of some options to the relevant commands (e.g. `rtl-table` to the `rtl` command)

Example of a `check` command which skips all decoder passes (also including timings, so you can verify that the passes are actually ignored):
```
openvadl check --decoder skip=all --timings sys/risc-v/rv64im.vadl
```
or with gradle run, if you prefer:
```
run --args="check --decoder skip=all --timings sys/risc-v/rv64im.vadl"
```


AndreasKrall (approved):
looks fine to me

PR #908, opened 2026-04-23
Add logback config to native image

rascmatt:
The logback configuration was not included in the native build, thus logback fell back to `DEBUG` level. This change will fix the issue of noisy logs when using the native image build.

Fyi, the log level can still be manually changed, by passing the `vadl.log.level` option. E.g.:

```
openvadl check sys/hexagon/hexagon.vadl -Dvadl.log.level=DEBUG
```


flofriday (approved):
LGTM, thanks that really helps :)

PR #911, opened 2026-04-26
parser: Fix crashes with invalid record invocations

flofriday:
I've also moved 4 tests from the parser directory to the macro one. Otherwise they are unchanged.

@AndreasKrall this fixes the crash you sent me by mail.


AndreasKrall (approved):
looks good to me

PR #912, opened 2026-04-27
frontend: Group type checking and VIAM lowering

rascmatt:
- Extend the type checker to check operation groups
- Introduce VIAM group definition


vadl/main/vadl/utils/Unit.java
```diff
@@ -0,0 +1,25 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package vadl.utils;
+
+/**
+ * Allow passing a unit value, as opposed to {@link  Void}, which has no instances. This avoids
+ * problems with the nullability checks.
+ */
+public enum Unit {
```
Jozott00: **question:** Why does using `Void` cause nullability issues?

Looking at the `AstVisitor.java` it seems like you can just `return null` even though the return type is `Void` without `@Nullable`. So apparently nullaway is treating `Void` as a special case.

vadl/main/vadl/viam/Group.java
```diff
@@ -0,0 +1,304 @@
...
+    public void accept(DefinitionVisitor visitor) {
+      visitor.visit(this);
+    }
+
+    public Operation op() {
+      return operation;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
Jozott00: **question:** Do we need the `equals` override here?

In general I always avoided implementing it, as every definition in the VIAM should exist only as a single object, so a definition in the VIAM is always identified by the object identity. If some logic is required to match definitions that behave equally, I think we should use a different method like `matches` instead.

(Same for the other definitions below)

vadl/main/vadl/utils/Unit.java
```diff
@@ -0,0 +1,25 @@
...
+  UNIT
+}
```
flofriday: **thought:** It seams a bit odd that we would need this. We already have quite a few visitors that where `Void` seems to work quite fine.

vadl/main/vadl/ast/ViamLowering.java
```diff
@@ -1554,6 +1624,7 @@ private InstructionSetArchitecture visitAndMergeIsa(InstructionSetDefinition def
     var memories = filterAndCastToInstance(allDefinitions, Memory.class);
     // TODO: @flofriday compute artifical resources
     var artificialResources = filterAndCastToInstance(allDefinitions, ArtificialResource.class);
+    var group = checkOneGroupDefinition(allDefinitions);
```
flofriday: **chore:** Let's move that to the typechecker. In general the Lowering shouldn't do any checks if possible and since it doesn't get run in the LSP this error wouldn't show.

vadl/main/vadl/ast/Group.java
flofriday, line 1: **rambling:** don't love this file, to be honest I didn't even notice it till now.

Why are groups so different from Definitions and Expressions?
Why do they need their own visitor?

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -75,7 +77,7 @@
  * of) them.
  */
 @SuppressWarnings("checkstyle:OverloadMethodsDeclarationOrder")
-public class TypeChecker
+public class TypeChecker extends DefaultGroupVisitor
```
flofriday: **minor:** I really don't like that we are extending a different visitor now. We only can this once in Java and I _really_ think Groups aren't important enough for that.

flofriday (commented):
I know the notes here are quite nitpicky but I think we already have too many edge cases in the frontend 😅

We can go over the changes together if you want to :)

vadl/main/vadl/utils/Unit.java
rascmatt: The problem occurred when having `Void` as an argument type to the visit method. In order to visit a node, you then have an expression like:
```java
expression.accept(this, (Void) null);
```
The cast to `Void` does nothing here, except showing the intended type. `NullAway` complains that you're not allowed to pass `null` to a non-nullable parameter, even though nobody will ever do anything with an argument of type `void`.

For return types of type `Void` NullAway does accept returning `null` as a return value though. So maybe this is a limitation by the framework?

Maybe we can the visitor entirely, by changing the Group Expressions to be actual AST Expressions?

vadl/main/vadl/ast/ViamLowering.java
rascmatt: Do you mean everything, or only checking the group definition? The problem here is, that I need the combined ISAs to determine if there is only applicable group definition, and they are only merged during the ViamLowering.

vadl/main/vadl/ast/Group.java
rascmatt: Agreed, maybe I can make them instances of Expressions? The thing is, that is is an expression syntax that we only have in group definitions, nowhere else.

vadl/main/vadl/ast/TypeChecker.java
rascmatt: Agreed :). 

I suppose I can at least make it an interface, but maybe we can entirely avoid it by making the Group-Expressions AST Expressions, and then reusing the Expression visitor.

rascmatt:
> I know the notes here are quite nitpicky but I think we already have too many edge cases in the frontend 😅
> 
> We can go over the changes together if you want to :)

Ty for the review 👍 ,

I'll see what I can do, and maybe we can talk about it on Wednesday.

Jozott00:
I forgot to actually submit my review 🙃

vadl/main/vadl/utils/Unit.java
Jozott00: Already answered above.

vadl/main/vadl/utils/Unit.java
rascmatt: Removed it again, as I was able to get rid of the non-nullable parameter type and using it with Void.

vadl/main/vadl/ast/Group.java
rascmatt: So, I did have a go at making the Group-Expression nodes proper Ast-Expressions, but this comes with its own challenges/ downsides imho.

Firs of all, let's clarify why they're so different from existing Definitions/ Expressions. A `group` is a definition in the traditional sense, i.e. it has a name and is a normal definition in the ISA. E.g.:
```vadl
group VLIW = O1 . ( O2<1..4> | {O3, O2, O1})
```
The expression itself is a (extended) regular expression over operations. We don't have this regular expression syntax anywhere else, and in fact, the grammar already restricts the regular expression syntax within a `group` definition. The same symbols used anywhere else, have different semantics behind them.

Integrating this into the existing Ast.Expr, would require us (as far as I can see) to give it a type, so we would probably have to introduce a new RegularExpression type and implement proper handling. (not sure about the exact steps here tbh). Also it would require use to integrate it into the BehaviorLowering, which doesn't really make sense, because the group expression does not express a behavior by itself.

Analogously I discussed with @Jozott00 to not make these expression nodes part of the VIAM Graph, but keep them as a separate structure. Since the RegExp structure is very basic anyways, this also simplifies the usage in the Generators.

vadl/main/vadl/ast/TypeChecker.java
rascmatt: I removed the abstract class and the TypeChecker now implements the appropriate `GroupVisitor<Void>` interface.

vadl/main/vadl/viam/Group.java
rascmatt: No, this was an artifact of IntelliJ converting my previous `record` into a `class`. I removed it.

vadl/main/vadl/ast/ViamLowering.java
rascmatt: Done, moved to the TypeChecker

vadl/main/vadl/ast/Group.java
rascmatt: Done, as discussed.

vadl/main/vadl/ast/Definition.java
```diff
@@ -962,6 +962,14 @@ List<InstructionSetDefinition> extendingNodes() {
         .toList();
   }
 
+  public <T extends Node> Stream<T> allInheritedNodesOf(Class<T> type) {
```
flofriday, line 972: **chore:** Let's add a comment here

I think that is also what blocks this right now cause checkstyle requires it.

flofriday (commented):
Almost there ✨

vadl/main/vadl/ast/Definition.java
rascmatt: done

flofriday (approved):

PR #913, opened 2026-04-28
iss: Save common expressions in helper instructions

Jozott00:
This PR reduces helper-code explosion for repeated helper-side expressions by materializing reused scalar subexpressions as `ExprSaveNode` temporaries before C emission.

It adds an ISS pass that identifies repeated helper-only scalar expressions, places a save at the latest dominating control location, and reuses that value instead of regenerating the same subtree at every use site. The pass stays conservative: it skips loop-index-dependent expressions, helper-preload-fed expressions, memory reads, function calls, and trivial aliases.

The change also cleans up the supporting infrastructure around it: dominator analysis was extracted into a shared ISS utility, helper codegen now avoids emitting pointless alias saves for already identifier-like values, and graph traversal helpers were centralized so the save pass and helper codegen use the same dependency-query logic.

The AArch64 SVE spec now works for all VL (128 - 2048 bits).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,8 @@ bin/
 
 ### Python ###
 .venv
+**/__pycache__/
+*.pyc
 
 ### OTHER ###
 .swp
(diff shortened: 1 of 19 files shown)


AndreasKrall (approved):
Looks fine to me. I noticed that also the tutorial was updated. I assume for AArch32 a lot of code sharing or other abstractions have to be added to keep the generated ISS small.

Issue #914, opened 2026-04-28
frontend: Crash with not handling of symbol expression


AndreasKrall:
When trying to typecheck the newest AArch32 spec in master in `sys/aarch32/armv6.vadl` there is a crash in the frontend.
The faulting line is 459:
```
model LdUH() : Ex = {(MEM<2> as  Word)}
```
which is called over multiple macro invocations from line 631:
```
//  $CondInstr          (HalfMemLoadImmInstr  ; (LDRHI ; "ldrh"  ; LdUH  ; 0b11'1011  ))
```
You have to uncomment this line to produce the following crash:
```
nativevadl check --expand-macros --decoder skip=all armv6.vadl
                             ___ ___    _   ___ _  _
                            / __| _ \  /_\ / __| || |
                           | (__|   / / _ \\__ \ __ |
                            \___|_|_\/_/ \_\___/_||_|

                       🔥 The OpenVADL compiler crashed 🔥

This shouldn't have happened, please open an issue with the stacktrace below at:
https://github.com/OpenVADL/open-vadl/issues/new


Before the crash, the following dumps were generated:
	- output/dump/expanded-macros.vadl

java.lang.RuntimeException: The typechecker doesn't know how to handle `symbol expression` yet, found in armv6.vadl:459:25 .. 459:30
	at vadl.ast.TypeChecker.throwUnimplemented(TypeChecker.java:336)
	at vadl.ast.TypeChecker.visit(TypeChecker.java:3988)
	at vadl.ast.TypeChecker.visit(TypeChecker.java:77)
	at vadl.ast.SymbolExpr.accept(Expr.java:1678)
	at vadl.ast.TypeChecker.checkWith(TypeChecker.java:185)
	at vadl.ast.TypeChecker.visit(TypeChecker.java:3970)
	at vadl.ast.TypeChecker.visit(TypeChecker.java:77)
	at vadl.ast.CastExpr.accept(Expr.java:2258)
	at vadl.ast.TypeChecker.checkWith(TypeChecker.java:185)
	at vadl.ast.TypeChecker.visit(TypeChecker.java:4264)
	at vadl.ast.TypeChecker.visit(TypeChecker.java:77)
	at vadl.ast.AssignmentStatement.accept(Statement.java:336)
	at vadl.ast.TypeChecker.check(TypeChecker.java:231)
	at java.base@25.0.2/java.util.ArrayList.forEach(ArrayList.java:1604)
[...]

PR #915, opened 2026-04-28
typechecker: Convert unimplemented crashes to errors

flofriday:
Now the LSP can also show them, and the typechecker can recover it's work.

This is related to: https://github.com/OpenVADL/openvadl/issues/914#issuecomment-4339070055

Such an error now looks like this with the LSP:

Screenshot 2026-04-29 at 00 35 12


AndreasKrall (approved):
looks fine to me

PR #916, opened 2026-04-29
cli: Disable semantic encoding verification by default

rascmatt:
Refactors the CLI options for the optional decoder steps to en- or disable options, with a certain default. It disables the semantic encoding verification pass by default, and allows to optionally include it like this:

```sh
openvadl check sys/spec.vadl --decoder with=encoding-verification
```


AndreasKrall (approved):
looks fine to me

PR #918, opened 2026-04-29
frontend: Group expressions and annotations

rascmatt:
This PR introduces the group annotations `assert` and `stop` by extending the type-system with support for quantified predicate expressions over operation sets. E.g.:

```vadl
[assert: forall i in {O1, O2} then i.par = 0b1]
group VLIW = O1.O2
```

These new language constructs are intentionally limited to group annotations.

---

The main type-system change is the introduction of `PseudoFormatType`, which represents variables bound by quantified group expressions. For each bound variable, the type checker derives a the pseudo format from the intersection of the formats of all instructions reachable through the referenced operation set. 

The participating formats need not be necessarily identical (e.g. in width), and shared fields do not need to be at the same bit position, which is required e.g. for Hexagon. The quantified variables expose the fields that are common to the whole set. Once such a field access is resolved, the field itself has a regular bits-type, so it integrates well into the existing type system.

On top of that, the parser, symbol resolution, type checking, and lowering were extended so that the new `forall` and `exist` expressions can be carried through the frontend and lowered into VIAM. The new predicate expressions themselves have result type `Bool`.


rascmatt:
Hi @Jozott00, @flofriday,

I would appreciate your feedback on the changes so far. I tried to document the requirements / changes in the PR description, but am happy to answer additional questions.

@flofriday we discussed the additions to the type system last week, but one notable difference to what we discussed is, that the restriction to the new `PseudoFormatType` we discussed are a bit too strict. For hexagon we will have to relax it, such that different-length formats and fields at different positions may still be combined in the operation set.

Thank you :)

vadl/main/vadl/viam/graph/dependency/GetFieldNode.java
```diff
@@ -0,0 +1,96 @@
...
+import vadl.viam.graph.Node;
+
+/**
+ * An expression node representing a field access, e.g. on an index of the
+ * forall-then expression.
+ * <pre>
+ *   forall x in {O1, O2} then x.a
+ * </pre>
+ */
+public class GetFieldNode extends ExpressionNode {
```
Jozott00: **note**: We will combine it with the TupleGetFieldNode

vadl/main/vadl/viam/graph/dependency/ForAllThenIdxNode.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+import vadl.types.Type;
+import vadl.viam.Operation;
+import vadl.viam.graph.Node;
+
+/**
+ * An expression node representing an index of the forall-then expression. That is, a
+ * bound variable restricted to some operation set.
+ */
+public class ForAllThenIdxNode extends ExpressionNode {
```
Jozott00: Move this into the `ForAllThenNode`

vadl/main/vadl/viam/graph/dependency/ForAllThenIdxNode.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private final List<Operation> operations;
```
Jozott00: Add `@DataValue` annotation

vadl/main/vadl/viam/graph/dependency/ForAllThenNode.java
```diff
@@ -0,0 +1,86 @@
...
+import vadl.types.Type;
+import vadl.viam.graph.GraphVisitor;
+import vadl.viam.graph.Node;
+import vadl.viam.graph.NodeList;
+
+/**
+ * An expression node representing the {@code forall i in {<operations>} then <body>}
+ * expression.
+ */
+public class ForAllThenNode extends ExpressionNode {
```
Jozott00: Some domain specific prefix would be nice

Jozott00: Maybe group

Jozott00: Or operations

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -1450,6 +1450,10 @@ private static BuiltIn instr(String name) {
       CONCATENATE_STRINGS
   );
 
+  public static final List<BuiltIn> equalityPredicates = List.of(
```
Jozott00: Use custom equality built-ins instead of using BitType equal ones.

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -122,7 +122,14 @@ static class StopPartialCheckingSignal extends RuntimeException {
...
    */
-  private final Deque<Integer> currentlyVisiting = new ArrayDeque<Integer>();
+  private final Deque<Integer> currentlyVisiting = new ArrayDeque<>();
+
+  /**
+   * For expressions within annotations, the top-level annotation definition providing
+   * context.
+   */
+  @Nullable
+  private AnnotationDefinition currentAnnotation = null;
```
Jozott00: @flofriday we could turn the `currentlyVisiting` stack into a `Deque<Node>`, so @rascmatt doesn't need the `currentAnnotation` anymore, as he could check that the first found definition on the stack is a `AnnotationDefinition`. What do you think?

Jozott00 (changes requested):

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -1610,3 +1624,12 @@ void resolveName(AnnotationDefinition definition, SymbolTable.SymbolResolver res
...
 }
+
+class GroupAssertionAnnotation extends ExprAnnotation {
+
+  @Override
+  void resolveName(AnnotationDefinition definition, SymbolTable.SymbolResolver resolver) {
+    // TODO:
+    super.resolveName(definition, resolver);
+  }
+}
```
flofriday: **thought:** This looks like it's a left over from an experiment, let's remove it or add a comment what it's meant to be.

vadl/main/vadl/ast/TypeChecker.java
flofriday: Yes, that would work, but checking membership will then be a bit uglier.

**From:**
```java
currentlyVisiting.contains(nodeId);
```

**To:**
```java
deque.stream().anyMatch(node -> node == target);
```

vadl/test/resources/frontend-snapshots/typechecker/invalidForallThenNotInsideGroupAnnotation.vadl
```diff
@@ -0,0 +1,44 @@
...
+  operation OpB = {B}
+
+  [assert : forall i in {OpA, OpB} then i.a = 0b01]
+  group VLIW = OpA.OpB
+}
+
+
+//  Reported Diagnostics:
+//
+//  error: The forall-then expression is only permissible for annotations on the `group` definition.
```
flofriday: **nitpick:** Let's move this into the description or locationdescription and create a shorter title `Invalid Forall Then Expression` or something.

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -746,6 +758,45 @@ private BuiltInCheckResult checkBuiltin(BuiltInTable.BuiltIn builtIn, List<Expr>
...
+                  location)
+                  .description("The sets of instructions bound by `%s` and `%s` do not overlap",
+                      ((Identifier) l).name, ((Identifier) r).name)
+                  .build());
+        }
+
+        return new BuiltInCheckResult(Type.bool(), args);
+      }
+    }
+
```
flofriday, line 807: **chore:** Let's add tests for this.

vadl/main/vadl/ast/PseudoFormat.java
```diff
@@ -0,0 +1,227 @@
...
+import javax.annotation.Nullable;
+import vadl.types.DataType;
+import vadl.types.Type;
+import vadl.utils.SourceLocation;
+import vadl.utils.WithLocation;
+
+/**
+ * A pseudo format constructed by intersecting multiple formats.
+ */
+public class PseudoFormat extends Definition implements TypedNode {
```
flofriday: **thought:** I don't think the PseudoFormat makes sense.

For one I'm not sure if it should be a definition because it cannot be typed by a programmer, it is only an intermediate datastructure for the compiler and it doesn't participate in the contracts like other Definitions do. Like in most visitors.

I think we can merge this with the PseudoFormatType and only rely on this one datastructure.

I might be wrong here though, and maybe have missed something.

vadl/main/vadl/ast/Ungrouper.java
```diff
@@ -204,6 +204,10 @@ public Expr visit(ExistsInThenExpr expr) {
     return expr;
   }
 
+  @Override
+  public Expr visit(ForallThenExpr expr) {
+    return expr;
+  }
```
flofriday, line 214: **minor:** I'm pretty sure this is wrong, because we can have subexpressions here that could have unnecessary groups.

vadl/main/vadl/ast/TypeChecker.java
Jozott00: For this we could just define a function that makes it easy to read it. 
@BadGraphixD also came up with a case where we have to know in which definition a behavior is currently handled: for next subcalls we can only determine the instruction with, if the call is done in an instruction, which must be checked in the typechecker. Havinng the visiting deque with the definitions would also allow this.

vadl/main/vadl/viam/graph/dependency/ForAllThenIdxNode.java
rascmatt: done

vadl/main/vadl/viam/graph/dependency/ForAllThenIdxNode.java
rascmatt: done

vadl/main/vadl/viam/graph/dependency/ForAllThenNode.java
rascmatt: done, I chose `OperationForAllNode`

vadl/main/vadl/types/BuiltInTable.java
rascmatt: Done.

The check for choosing the correct overload of the `=` operator is a bit strange, especially since there is already a `takes` method supported by every BuiltIn. Unfortunately it's not that simple, as the `takes` method mainly checks `Bits` length compatibility, and is not fully capable of correctly handling singed/unsigned cases.

So for now I specifically handle the new overload case.

vadl/main/vadl/ast/AnnotationTable.java
rascmatt: Spot on, this is not needed anymore. Thx.

vadl/test/resources/frontend-snapshots/typechecker/invalidForallThenNotInsideGroupAnnotation.vadl
rascmatt: done

vadl/main/vadl/ast/TypeChecker.java
rascmatt: Done.

To make warnings available to the frontend snapshot tests I additionally added diagnostics from the `DeferredDiagnosticStore`.

vadl/main/vadl/ast/TypeChecker.java
rascmatt: @flofriday | @Jozott00 

Added an `IdentityDeque` to allow for easy identity-based membership checks, but allow access to the full stack of currently visiting nodes.

For now I implemented two helpers to access the visiting context, but we can of course extend/change it if we find it to be not helpful for @BadGraphixD's use case.

vadl/main/vadl/viam/graph/dependency/GetFieldNode.java
rascmatt: Done with #963

vadl/main/vadl/ast/PseudoFormat.java
rascmatt: Removed the two separate definitions, we only have the `PseudoFormatType` now (for lack of a better name). I was also able to get rid of this synthetic definition.

vadl/main/vadl/ast/Ungrouper.java
rascmatt: Good catch, thank you. I added ungrouping for the expression body, for the indices it's not necessary.

rascmatt:
@Jozott00 | @flofriday 

Thank you for the reviews, I addressed the comments. If there's nothing further, I'd like to merge it soon, so I don't run into merge conflicts every few days.

Jozott00 (approved):
LGTM

PR #919, opened 2026-04-30
vadl: Convert all pass timinigs to nanoseconds

flofriday:
During some performance evaluation we discovered that the previous tracking in milliseconds wasn't granular enough as the native build on small specs is simply to fast.


Jozott00 (approved):

AndreasKrall (approved):
looks fine to me

PR #921, opened 2026-04-30
iss: Add benchmark workflow for embench

Jozott00:
This PR adds dedicated ISS Embench benchmark execution classes.
Additionally, it adds (and refactors) the benchmark publication to Resselpark.


.github/workflows/reusable-benchmark-publish.yml
```diff
@@ -0,0 +1,68 @@
...
+        if: ${{ inputs.extra-env != '' }}
+        run: printf '%s\n' "${{ inputs.extra-env }}" >> "$GITHUB_ENV"
+      - name: Run benchmark
+        run: ${{ inputs.benchmark-command }}
+        env:
+          isCI: true
+      - name: Prepare artifact
+        run: cp "${{ inputs.artifact-source }}" "${{ github.workspace }}/${{ inputs.artifact-name }}"
+      - name: Pushes artifact
+        uses: dmnemec/copy_file_to_another_repo_action@main
```
rascmatt: Will this work on ARM? The comment from the `benchmark-lcb.yml` says the plugin requires linux, and fixed the x64 runner...

Jozott00: Oh yeah right... the repo of the action doesn't mention it though. 

If it doesn't work, I will try an other approach.

Jozott00: Well...
> Run dmnemec/copy_file_to_another_repo_action@main
Error: Container action is only supported on Linux

Jozott00: Replaced it by pure git comments

rascmatt (approved):

PR #924, opened 2026-05-01
frontend: Add native frontend benchmark for aarch64

flofriday:
Add automatic benchmarks.

@Jozott00 can you look over this?

### DIFF
--- a/.github/workflows/benchmark-frontend.yml
+++ b/.github/workflows/benchmark-frontend.yml
@@ -32,10 +32,45 @@ jobs:
     uses: ./.github/workflows/reusable-benchmark-publish.yml
     with:
       runner: self-hosted-arm64
-      benchmark-command: python3 vadl/test/resources/scripts/frontend/benchmark_frontend.py sys/aarch64/aarch64.vadl
+      benchmark-command: python3 scripts/benchmark/benchmark_frontend.py sys/aarch64/aarch64.vadl
       artifact-source: vadl/output/timings-mean.csv
       artifact-name: ${{ github.event.repository.updated_at }}_${{ github.sha }}.csv
       destination-folder: data/frontend/aarch64/native/results
       commit-message: 'data: Pushed frontend aarch64 native data'
     secrets: inherit
 
+  rv32i:
+    needs: build
+    uses: ./.github/workflows/reusable-benchmark-publish.yml
+    with:
+      runner: self-hosted-arm64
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark-frontend.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: self-hosted-arm64
+    steps:
+      - uses: actions/checkout@v4
+      - uses: graalvm/setup-graalvm@v1
+        with:
+          java-version: '25'
+          distribution: 'graalvm'
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build
+        run: ./gradlew nativeBuild
```
flofriday, line 28: **question:** Will the artifacts generated here be available in the other builds?

.github/workflows/benchmark-frontend.yml
```diff
@@ -0,0 +1,76 @@
...
+
+  aarch64:
+    needs: build
+    uses: ./.github/workflows/reusable-benchmark-publish.yml
+    with:
+      runner: self-hosted-arm64
+      benchmark-command: python3 vadl/test/resources/scripts/frontend/benchmark_frontend.py sys/aarch64/aarch64.vadl
+      artifact-source: vadl/output/timings-mean.csv
```
flofriday, line 36: **question:** All benchmarks produce the same file, are they run in sequence because otherwise there might be race-conditions.

scripts/benchmark/benchmark_frontend.py
Jozott00, line 1: Maybe we should move the script to scripts/benchmarks/ ? It is not really a resource used by the Java program execution.

Jozott00 (approved):
LGTM

.github/workflows/benchmark-frontend.yml
Jozott00: What do you mean by "in other builds"?

.github/workflows/benchmark-frontend.yml
Jozott00: Yes, one runner can only execute a single job at once.

.github/workflows/benchmark-frontend.yml
flofriday: We build the native binary here and use it below in the benchmark jobs. Or do we need to build it there?

Jozott00: Hmm, I think it should be cached.

Jozott00 (approved):

PR #925, opened 2026-05-03
iss: Move bi-endian condition evaluation to DisasContext creation

BadGraphixD:
- If the memory definition is bi-endian, the DisasContext gets a new field `endian_cond`, which stores the value of the bi-endian condition
  - The condition is evaluated once per TB instead of per instruction (fetch and memory op)
  - Added IssStaticEndianConditionNode, which is translated to `ctx->endian_cond`


Jozott00 (approved):
That's a nice improvement!

Jozott00 (approved):

PR #926, opened 2026-05-04
rtl: Use forwarding spec in MiA

linushdot:
Use the `readOrForward` and `results` calls in the MiA specification to generate partial forwarding logic.


rascmatt:
lgtm

rascmatt (approved):

PR #930, opened 2026-05-04
setup: Enable GC for native builds

### LINKED ISSUE #927: setup: Is disabling the GC reasonable?
At the moment we disable the GC for native builds.

The [official documentation](https://www.graalvm.org/latest/reference-manual/native-image/optimizations-and-performance/MemoryManagement/) mentions that: 

> [EpsilonGC is used for] very short running applications that only allocate a small amount of memory.

While OpenVADL is certainly designed to be short running in most cases (more on that later) non-trivial specs do allocate a lot of memory. 
- sve: 511MB
- miniARMv7: 4.85GB
- aarch32: 4.88GB

And this is just the check command, for the generators it's even more likely that they will run out of memory.

At least for my thesis one big motivator for the performance is the responsiveness of an LSP. However, the LSP is no longer a _very short running application_ and therefore will always need a GC to avoid crashing every couple of minutes.

Obviously we could generate a separate binary for the LSP with the GC enabled, but my bigger problem is that the results with the GC disabled aren't representative for an environment where a GC is inherently required.

**I suggest** enabling the default `Serial GC` and tuning it's parameters for throughput (since latency isn't that important for compilations, and even for the LSP humans are quite forgiving). I would love to use `G1` but it isn't available for GraalVM Community Edition.

flofriday:
Closes: #927

The GC is now (as good as possible) the same between native and normal graalvm.

I've decided to set the limits in a way that is quite liberal with memory, and focuses on throughput rather than latency which does make sense for compilations. 

For the LSP this might be different, as it is right now for most specs not a single GC cycle is run, for aarch32 one is. However even then almost 40% is spent in the GC.

### DIFF
--- a/vadl-cli/build.gradle.kts
+++ b/vadl-cli/build.gradle.kts
@@ -34,7 +34,14 @@ graalvmNative {
             }
             imageName.set("openvadl")
             mainClass.set(application.mainClass)
-            buildArgs.addAll("-O2", "--gc=epsilon")
+            buildArgs.addAll(
+                "-O2",
+                // This is different from the GC as for graalvm builds, which is mainly because this is the only one that
+                // also works outside Linux.
+                "--gc=serial",
+                "-R:MinHeapSize=4g",
+                "-R:MaxNewSize=2g",
+            )
             buildArgs.add("--enable-url-protocols=https")
             buildArgs.add("--enable-native-access=ALL-UNNAMED")
         }
@@ -45,6 +52,9 @@ tasks.startScripts {
     defaultJvmOpts = listOf(
         "-XX:TieredStopAtLevel=1",
         "--enable-native-access=ALL-UNNAMED",
+        "-Xms4g",
+        "-Xmn2g",
+        "-XX:+UseParallelGC"
     )
 }
 


Jozott00:
Shouldn't be the requirement for a short-running compiler more like small latency over high throughput? So that we reduce the compilation time (latency) for small specs. (maybe I understood something wrong)

flofriday:
> Shouldn't be the requirement for a short-running compiler more like small latency over high throughput? So that we reduce the compilation time (latency) for small specs. (maybe I understood something wrong)

I see where you are coming from but in the context of GCs these terms are somewhat confusingly named. _Latency_ is the negative impact on responsiveness, most commonly measured in how long each pause time is where the GC stops the mutator (our program). For video games, web browsers, video players, air bag sensors that is important because they will eventually drop frames or cause human casualties. In that context it's often better to stop more often but only short times to stay responsive, even if it causes more time spent in the GC overall. 

In applications (like a compiler) where we don't care about responsiveness because we only care when we will get to the end and not _how_ we get there we only care about _Throughput_, the total amount of time spent in the GC, less time in the GC ==> higher throughput.

The graalvm docs explain that also quite well: https://www.graalvm.org/latest/reference-manual/native-image/optimizations-and-performance/MemoryManagement/#performance-considerations

Jozott00 (approved):
LGTM!

PR #932, opened 2026-05-04
iss: Implement alias program counter

BadGraphixD:
- `alias program counter` may refer to another single register (also inside register files/tensors, but not register aliases)
  - This implicitly creates a Counter, which references the register
  - Any reads/writes to the alias AND the referenced register are interpreted as reads/writes to the counter
    - If such an access involves non-constant indices, then the access node is duplicated and a runtime (usually tcg-translation-time) check is inserted such that we statically know for each node, whether it is a PC access
  - Counter subcalls (such as `current`, `next` and `nextnext`) can only be performed on the alias, not the referenced register


AndreasKrall:
> * `alias program counter` may refer to another single register (also inside register files/tensors)
>   
>   * Counter subcalls (such as `current`, `next` and `nextnext`) can only be performed on the alias, not the referenced register

It is the correct solution when counter subcalls only can be done on the counter register. But the counter annotation (nextnext) on the counter register must be obeyed also on the referenced register. This means that always the unmodified value is stored in the register, but when the register is read, the value has to be 2 times the instruction size higher. And the referenced register has to be updated after every instruction equivalent to the counter.

vadl/main/vadl/ast/AnnotationTable.java
```diff
@@ -110,6 +110,22 @@ public class AnnotationTable {
...
+        .check(ctx -> {
+          var alias = ctx.astTargetDef;
+          ensure(alias.kind.equals(AliasDefinition.AliasKind.PROGRAM_COUNTER),
+              () -> error("Invalid annotation target", alias)
+                  .locationDescription(alias,
+                      "Program counter annotations can only be applied on program counters "
+                          + "and program counter aliases"));
+          ctx.verifyOnlyOneOfGroup();
+        })
+        // FIXME: Apply to AST
```
Jozott00: **question:** Are you going to implement PC offset annotations in a separate PR?

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -1591,157 +1591,167 @@ public Void visit(AliasDefinition definition) {
...
-                  error("Constant accesses cannot occure after param accesses.",
-                      argVal).build());
-            }
-            dummyArgs.add(arg);
+          wildcardIndices.add(i);
+        } else if (argVal instanceof RangeExpr) {
+          if (definition.kind == AliasDefinition.AliasKind.PROGRAM_COUNTER) {
+            // FIXME: should program counters be able to be a slice of a register?
+            //        eg. the lower 32 bits of a 64-bit register
+            addErrorAndStopChecking(
```
Jozott00, line 1661: **minor**: Should be fine for now. Please create an issue and reference it in this FIXME comment

vadl/main/vadl/gcb/passes/DetermineBuiltinAttributesPass.java
```diff
@@ -94,6 +94,9 @@ public Object execute(PassResults passResults, Specification viam) throws IOExce
   private boolean isRedFlag(Specification viam, Graph snapshot) {
     var pc =
         Objects.requireNonNull(ensurePresent(viam.isa(), "must be present").pc()).registerTensor();
+    // FIXME: Checking if the registers are equals is no sufficient for checking if the PC is read.
+    //        A PC can be only a single register file entry -> also check indexing or use
+    //        CanAccessCounter::isPcAccess
     var hasPc = snapshot.getNodes(ReadsRegisterTensor.class).filter(
```
Jozott00, line 101: **minor**: Please add an issue, assign LCB people and add the issue number to this FIXME comment

vadl/main/vadl/gcb/passes/GenerateGcbIntrinsicsPass.java
```diff
@@ -143,6 +143,9 @@ public Object execute(PassResults passResults, Specification viam) throws IOExce
   private boolean isRedFlag(Specification viam, Graph snapshot) {
     var pc =
         Objects.requireNonNull(ensurePresent(viam.isa(), "must be present").pc()).registerTensor();
+    // FIXME: Checking if the registers are equals is no sufficient for checking if the PC is read.
+    //        A PC can be only a single register file entry -> also check indexing or use
+    //        CanAccessCounter::isPcAccess
     return snapshot.getNodes(ReadsRegisterTensor.class).anyMatch(
```
Jozott00, line 150: **minor**: Reference the same issue as created above

vadl/main/vadl/iss/passes/extensions/RegInfo.java
```diff
@@ -294,6 +298,30 @@ public String cArrayIndex(String indexPrefix) {
...
+  /**
+   * Returns the array access for registers in the cpu state.
+   * CPU registers are rendered as a single value or 1D array, where all dimensions
+   * are flattened to one.
+   *
+   * @param indices the concrete indices to access with
+   * @return C array index expression or empty string if no indices
+   */
+  @SuppressWarnings("MethodName")
+  public String cArrayIndex(List<Constant.Value> indices) {
```
Jozott00: **minor**: I think this should be moved to the IssTemplateRenderingPass (caller side) as the only RegInfo state it uses is reg() for an argument check.

vadl/main/vadl/viam/graph/CanAccessCounter.java
```diff
@@ -0,0 +1,55 @@
...
+import vadl.viam.passes.staticCounterAccess.CounterAccessResolvingPass;
+
+/**
+ * Interface to indicate that the implementing class could access a {@link Counter}.
+ *
+ * <p>The {@link #staticCounterAccess()} indicates if this access is known to be
+ * (program) counter access. It is set by the
+ * {@link CounterAccessResolvingPass}</p>
+ */
+public interface CanAccessCounter extends HasRegisterTensor {
```
Jozott00, line 32: **minor**: Please also clarify that the CounterAccessResolvingPass ensures that if the access is non-static, it is known that the access to the register has no affect on the program counter (by inserting dynamic checks on the potential register indices).

vadl/main/vadl/viam/passes/staticCounterAccess/CounterAccessResolvingPass.java
```diff
@@ -0,0 +1,200 @@
...
+ * Now given a {@link WriteRegTensorNode} in a behavior, we can statically determine if it
+ * writes the {@code PC}, if all the indices used are constant (or when the user wrote
+ * {@code PC := ...}). If they are not (e.g. if
+ * they come from a format field), we insert a runtime check: We duplicate the node, mark the
+ * duplicate as a {@code PC} access and insert a runtime condition that decides which node
+ * is used. The runtime condition compares the indices of the node to the indices of the
+ * program counter. Now we can statically determine for each of the nodes, if they are
+ * a {@code PC} access.
+ */
+public class CounterAccessResolvingPass extends Pass {
```
Jozott00, line 73: Nice approach!

vadl/main/vadl/viam/passes/staticCounterAccess/CounterAccessResolvingPass.java
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  private static void handleIndexedAccess(CanAccessCounter node, Counter counter) {
+    var indexCount = node.indices().size();
+
+    // check if the indices match
+    if (indexCount != counter.indices().size()) {
+      // FIXME: what if we write to multiple registers (eg by leaving out the last index)
+      //        is that possible? If yes, then this check fails to notice a PC change!
+      return;
```
Jozott00: **minor**: Yes, this could happen. For now, provide a user friendly error that this is not yet supported for `register` definitions that are used as source register of an `alias program counter`. Also create an issue and reference the issue as `.note()` in the diagnostic.

vadl/main/vadl/lcb/passes/isaMatching/IsaMachineInstructionMatchingPass.java
```diff
@@ -174,6 +174,8 @@ public Result execute(PassResults passResults, Specification viam) throws IOExce
       return new Result(Collections.emptyMap(), Collections.emptyMap());
     }
 
+    // FIXME: a pc can be a single register in a register file. if that should be supported
+    //        here, use `pc.isSingleRegister()` instead to check that
     var pc = isa.pc();
```
Jozott00, line 180: **minor**: Reference the same github issue as the one above

vadl/main/vadl/pass/PassOrders.java
```diff
@@ -522,6 +522,8 @@ public static PassOrder iss(IssConfiguration config) throws IOException {
     order
         .add(new IssInfoRetrievalPass(config))
         .add(new IssConfigurationPass(config))
+        // TODO: this can add new conditional side effects. But SideEffectConditionResolvingPass
+        //       runs before...
         .add(new IssApplyMemoryEndiannessPass(config))
```
Jozott00: This shouldn't be a problem for the ISS, as we never rely on the condition. The hardware generator is the only one using it.

Jozott00 (changes requested):
Well done! Please add some small frontend tests to check diagnostics and different valid `alias program counter` definitions.

EDIT: We should probably also reject an `alias program counter` pointing to a `alias register`. (with a test case)

vadl/main/vadl/ast/AnnotationTable.java
BadGraphixD: Yes, I will do it in a separate PR. It's also required for the aarch32 spec afaik.

BadGraphixD: See #938

vadl/main/vadl/ast/TypeChecker.java
BadGraphixD: See #939

vadl/main/vadl/gcb/passes/DetermineBuiltinAttributesPass.java
BadGraphixD: See #941

vadl/main/vadl/iss/passes/extensions/RegInfo.java
BadGraphixD: Imo it should stay there because:
- the indexing depends on the dimensions of the reg tensor (see `IssUtils.cIndex()`)
- there already exists another method there (also called `cArrayIndex`), which does a similar thing

But I can move it if you want

Jozott00: Ok, thats also fine for me. Then call this overloaded method in the existing one, so they share the same core logic.

vadl/main/vadl/viam/passes/staticCounterAccess/CounterAccessResolvingPass.java
BadGraphixD: See #942

vadl/main/vadl/iss/passes/extensions/RegInfo.java
BadGraphixD: Nvm, a check was unnecessary. Moved it to the caller side.

docs/refman/tutorial.md
```diff
@@ -1033,7 +1033,7 @@ The memory characteristics can be detailed with different annotations.
 If no annotation is defined, the memory serves both as data and instruction memory, the memory access is carried out in little endian mode, there is no address translation and the memory consistency model is sequential consistency.
 With the annotation `[data]` a memory is only used for data.
 With the annotation `[instruction]` a memory is only used for instructions.
-With the annotation `[bigEndian]` a memory is only accessed in big endian mode.
+With the annotation `[big endian]` a memory is only accessed in big endian mode.
```
flofriday, line 1036: **praise:** Nice catch 🎉

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -1591,157 +1591,166 @@ public Void visit(AliasDefinition definition) {
...
+    if (reg == null) {
+      // if this does not directly reference a register,
+      // it might reference another alias definition
+      var alias = definition.symbolTable().findAs(targetIdent, AliasDefinition.class);
+      if (alias == null || alias.kind != AliasDefinition.AliasKind.REGISTER) {
+        throw addErrorAndStopChecking(
+            error("Unknown alias source register", targetIdent.location())
+                .locationDescription(targetIdent.location(), "Unknown register `%s`.",
+                    targetIdent)
+                .build());
```
flofriday, line 1611: **nitpick:** We could also provide suggestions that could work.

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -1591,157 +1591,166 @@ public Void visit(AliasDefinition definition) {
...
-                        + "https://github.com/OpenVADL/openvadl/issues/new")
-                .build()
-        );
-      }
+    if (definition.targetType != null) {
+      // FIXME: Support relational alias types on registers
+      addErrorAndStopChecking(
+          error("Unsupported Alias Type", definition)
+              .locationDescription(definition,
+                  "The typechecker doesn't know how such aliases yet.")
```
flofriday, line 1625: **chore:** Grammar, this isn't quite English 😅

Jozott00 (approved):

PR #933, opened 2026-05-05
frontend: Performance improvements

flofriday:
This PR cleans up some parts in the frontend, mostly with the goal to squeeze a bit more performance out of it.

Improvements:
- Fewer source locations to be allocated by implementing a rope
- Removed unused fields from symboltable
- Faster AST children collection 

Before-After on my M3 Max:
Screenshot 2026-05-06 at 15 21 07

I think this is especially good on native because it produces less garbage and the GC doesn't have that many references to trace, which would make sense because the graalvm build uses a more advanced GC than native.


Jozott00 (approved):
LGTM!

PR #937, opened 2026-05-07
pass: Refactor PassOrders and ISS pass packages

Jozott00:
The `PassOrders` class is becoming larger and contains the pass logic for all backends. This PR splits it into one class per backend.

Additionally, the pass packages of the ISS are now split into `common, scalar, helper` packages to separate different lowering techniques.


vadl/main/vadl/pass/order/CheckPassOrder.java
```diff
@@ -0,0 +1,48 @@
...
+  public static PassOrder create(GeneralConfiguration configuration) {
+    var order = new PassOrder();
+    order.add(new PassOrders.ViamCreationPass(configuration));
+
+    OrderSupport.addHtmlDump(order, configuration, "VIAM Creation",
+        "Dump directly after frontend generated VIAM.");
+
+    order.add(new ViamVerificationPass(configuration));
+    OrderSupport.addDecodePasses(order, configuration);
+    OrderSupport.addHtmlDump(order, configuration,
```
rascmatt, line 43: Since it's now possible to disable the decode passes entirely (for the check command), imo we should move this dump into the `addDecodePasses`.

--

fyi:

I'm anyways considering to rework the dump behavior for the decode passes a bit. E.g. I'd like to add the option (to the CLI) to dump the SMT format used to verify semantic & decode tree verifications etc. But that is not something that should necessarily be included in the general dump, which is already too big. I started a branch for that a few days ago, but did not finished it yet `feat/sem-ver-dump`.

Maybe we should generally have a more fine-granular way to select specific dump artifcats.

vadl/main/vadl/pass/order/OrderSupport.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    order.add(new VdtLoweringPass(config));
+
+    if (config.getDecoderOptions().isEnabled(OPT_DECODER_VERIFICATION)) {
+      order.add(new VdtVerificationPass(config));
+    }
+  }
+
+  static void addUserModeEmitPasses(PassOrder order, IssConfiguration config) {
```
rascmatt: Isn't this also rather ISS specifict? Looks to me like it belongs to the `IssPassOrder`, don't you think?

Jozott00: Yes definitely!

vadl/main/vadl/pass/order/CheckPassOrder.java
Jozott00: Yeah, I agree that the current approach of dumping everything is not the best. We should come up with some `--dump` flag extension (that does not bloat the command line argument space)

Jozott00: Regarding the dump: I would not move it to the `addDecodePasses`, because every other generator has its own dump phase. When adding it to the `addDecodePasses` method, it will implicitly create a second dump phase which may have big impact on performance.

rascmatt: Ok, let's resolve this for now

rascmatt (approved):

PR #946, opened 2026-05-11
ci: Updated gradlew command for rtl

kper:
The current command also runs the tests for the lsp. It fails because the lsp has no `RtlRisVBenchmarkTest` test. This PR changes the command to only run the vadl test suite.


Jozott00 (approved):
Thanks!

Jozott00:
(I think this can be merged without waiting for the CI to pass)

PR #947, opened 2026-05-11
frontend: Further performance improvements

flofriday:
This is some continuation of my work with #933 

These optimizations come from a mixture of educated guesses and spending too much time in the profiler, investigating memory allocations and time spent.

### Results
Here are some benchmark results from my M3 Max with the miniARMv7 spec on native.

- **master:**  507ms
- **Avoiding empty entries in RopeList** 483ms
- **Reusing rope in macro expander:** 479ms
- **Rewriting macro expander in imperative style and preallocate lists:** 372ms
- **Refactor location joining:** 369ms (not noteworthy but cleaner, and faster in lowering)
- **Better builtin lookups:** 321ms
- **Avoiding unecessary lists in format field access:** 297ms
- **Replacing many forEach with loops:** 290ms

### DIFF
--- a/vadl/main/vadl/ast/AstUtils.java
+++ b/vadl/main/vadl/ast/AstUtils.java
@@ -18,7 +18,6 @@ package vadl.ast;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import vadl.types.BuiltInTable;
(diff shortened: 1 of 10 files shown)


Jozott00 (approved):
Nice! 

Good to know that using the imperative for/if construct instead of stream/filter only results in negligible performance improvements.

flofriday:
> Good to know that using the imperative for/if construct instead of stream/filter only results in negligible performance improvements.

Well, sometimes, so for example in _Rewriting macro expander in imperative style and preallocate lists_ I first tried it without rewriting in an imperative style and it was **slower** than before, and only after rewriting in an imerative style it got considerbly faster (100ms). 

So I think it heavily depends on the kind of lambdas provided and whether or not they close over local variables.

PR #948, opened 2026-05-11
viam: Extract `Abi.RegisterRef` and remove `GeneratesRegisterFileName` interface

Jozott00:
This PR extracts most of the `Abi.RegisterRef` to `vadl.viam.RegisterRef`, so the datastructure can be reused in other definitions like the UME (#834). 

Additionally, it removes the `GeneratesRegisterFileName` which does not fit into the VIAM hierarchy and is already covered by the `RegisterResource` class.


Jozott00:
@arcane-quill this PR will allow you to use `RegisterRef` for UME stuff like `syscall register`.

rascmatt (approved):
lgtm

PR #949, opened 2026-05-11
ci: Set artifact name based on run timestamp

flofriday:
Unfortunatley, I couldn't find any good way to share the same timestamp accross multiple benchmarks.

### DIFF
--- a/.github/workflows/reusable-benchmark-publish.yml
+++ b/.github/workflows/reusable-benchmark-publish.yml
@@ -16,10 +16,6 @@ on:
         description: 'Path to the produced artifact'
         required: true
         type: string
-      artifact-name:
-        description: 'Published artifact filename'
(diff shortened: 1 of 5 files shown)


Jozott00 (approved):
I guess this small VM execution on Github won't be too bad

.github/workflows/reusable-benchmark-publish.yml
```diff
@@ -40,7 +36,16 @@ on:
         type: string
 
 jobs:
+  timestamp:
+    runs-on:  ${{ inputs.runner }}
```
Jozott00, line 40: Will this work on the mac-mini?

Jozott00: Oh yeah it will

Issue #764, opened 2026-02-27
ume: Write User Documentation


Jozott00:
To make user mode emulation usable, we must add extensive documentation and usage examples to our online docs in `docs/`.

Jozott00:
@arcane-quill I believe it’s a good idea to create an in-depth user documentation for the user mode emulation now. This will help you understand the full extent of the generic `user mode emulation` definition.

In `docs/refman/tutorial.md`, add a new section titled `## User Mode Emulation` at the bottom. Provide a detailed description of the definition, including examples, as is done for other definitions in the tutorial. Create a new branch and open a new PR where you work on this (no code changes, only documentation). Once this documentation is finalized and covers all user mode emulation specification cases, and is approved by @AndreasKrall, we can proceed with the further implementation of the feature.

arcane-quill:
> [@arcane-quill](https://github.com/arcane-quill) I believe it’s a good idea to create an in-depth user documentation for the user mode emulation now. This will help you understand the full extent of the generic `user mode emulation` definition.
> 
> In `docs/refman/tutorial.md`, add a new section titled `## User Mode Emulation` at the bottom. Provide a detailed description of the definition, including examples, as is done for other definitions in the tutorial. Create a new branch and open a new PR where you work on this (no code changes, only documentation). Once this documentation is finalized and covers all user mode emulation specification cases, and is approved by [@AndreasKrall](https://github.com/AndreasKrall), we can proceed with the further implementation of the feature.

okay good idea! will start this today

PR #950, opened 2026-05-11
ume: Add extensive documentation for user mode emulation 

arcane-quill:
This PR covers issue #764 and adds extensive documentation for the user mode emulation, including examples and descriptions for various specification cases

### DIFF
--- a/docs/refman/tutorial.md
+++ b/docs/refman/tutorial.md
@@ -2007,4 +2007,51 @@ In addition, several behavioral aspects of the cache can be specified, such as w
 
 \lbl{tut_ume_definition}
 
-The \ac{UserModeEmulation} definition determines how a VADL specification maps to QEMU's Linux user-mode emulation (UME). 
+The \ac{UserModeEmulation} definition determines how a specification from a VADL file maps to QEMU's Linux user mode emulation (UME). 
+The UME is used to run Linux ELF binaries without having to emulate the full machine. 
+To achieve this it needs to intercept system calls (syscalls) based on the syscall instruction of the architecture given by the VADL file. 
+The associated register conventions (the \ac{ABI}) need to be mapped to determine the system call number, its arguments, and where to
+place the return value.
+
+\listing{user_mode_emulation, User Mode Emulation Definition}
+~~~{.vadl}
+user mode emulation UME for RV64 with ABI = {
+  
+}
+~~~
+\endlisting
+
+A \ac{UME} section starts with the keyword `user mode emulation` followed by a unique identifier.
+Some elements inside of the \ac{UME} section rely on previously defined \ac{ISA} elements, so it needs to be referenced
+using the `for` keyword after the identifier.
+Definitions from the referenced \ac{RV64} are now available for use inside the \ac{UME} section.
+The `with` keyword binds the \ac{ABI} to the section. 
+
+### Exception Definition 
+
+Varying exceptions are defined in the VADL like:
+
+\listing{user_mode_emulation_exc, User Mode Emulation Definition Exceptions}
+~~~{.vadl}
+user mode emulation UME for RV64 with ABI = {
+  
+}
+~~~
+\endlisting
+
+It is also possible to select the desired OS in the .vadl file (e.g. Linux or BSD). In order for this to work, custom VADL annotations 
+and tokens are added to the .ATG file. 
+
+Templates are generated based on a RISCV implementation from QEMU's linux-user section, that has been interpolated using Thymeleaf to be viable for 
+any architecture specified in the VADL. 
+
+The cpu_loop intercepts the specific syscall instruction and dispatches to the syscall translator. 
+
+Syscalls are translated by utilizing QEMU's integrated do_syscall() function that maps guest syscall numbers to host syscall numbers and 
+writes the return value back into the guest's return register.
+
+The mechanics of how unix signals are handed to a guest process are handled in signal.c, which makes sure that the host signal gets intercepted, a signal frame
+gets set up on the guest stack and the guest PC gets returned to the registered signal handler of the guest whenever a guest program triggers a signal - for example
+by accessing an invalid address (SIGSEV). 
+
+Additional testing infrastructure ensures that the UME works as intended. 
--- a/docs/refman/tutorial.md
+++ b/docs/refman/tutorial.md
@@ -2016,7 +2016,7 @@ place the return value.
 \listing{user_mode_emulation, User Mode Emulation Definition}
 ~~~{.vadl}
 [os : linux] 
-user mode emulation UME from RV64IM with ABI = {
+user mode emulation UME for RV64IM with ABI = {
 syscall instruction = ECALL
 
 syscall number = a7
--- a/docs/refman/tutorial.md
+++ b/docs/refman/tutorial.md
@@ -2072,6 +2072,21 @@ key-value pairs are the name of the syscall and its respective number. A generic
 table already exists, so the table defined in the VADL file provides the architecture-specific syscalls 
 that should either be added to or overwrite the generic ones. 
 
+\listing{user_mode_emulation, User Mode Emulation Definition}
+~~~{.vadl}
+[os : linux] 
+user mode emulation UME for RV64UME with ABI = {
+...
+
+syscall table = LinuxSyscall
+
+}
+~~~
+\endlisting
+
+The `syscall table` declaration defines the previously mentioned enumeration `LinuxSyscall` as the syscall table that is
+being used for the \ac{UME}. 
+
 ### Exception Definition (work in progress)
 
 Varying exceptions are defined in the VADL like:
--- a/docs/refman/tutorial.md
+++ b/docs/refman/tutorial.md
@@ -2015,8 +2015,14 @@ place the return value.
 
 \listing{user_mode_emulation, User Mode Emulation Definition}
 ~~~{.vadl}
-user mode emulation UME for RV64 with ABI = {
-  
+[os : linux] 
+user mode emulation UME from RV64IM with ABI = {
+syscall instruction = ECALL
+
+syscall number = a7
+syscall argument = a{0..5}
+syscall return = a0
+
 }
 ~~~
 \endlisting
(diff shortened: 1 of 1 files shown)


AndreasKrall:
@arcane-quill The language tutorial is a tutorial for the user of VADL. It has to explain how to use UME and has to include some examples. Therefore, in the tutorial there should not be any text about the implementation. So delete everything starting with `In order for this to work, custom VADL annotations` and all the following lines.

docs/refman/tutorial.md
```diff
@@ -2002,3 +2002,67 @@ In addition, several behavioral aspects of the cache can be specified, such as w
...
+The \ac{UserModeEmulation} definition determines how a specification from a VADL file maps to QEMU's Linux user mode emulation (UME). 
+The UME is used to run Linux ELF binaries without having to emulate the full machine. 
+To achieve this it needs to intercept system calls (syscalls) based on the syscall instruction of the architecture given by the VADL file. 
+The associated register conventions (the \ac{ABI}) need to be mapped to determine the system call number, its arguments, and where to
+place the return value.
+
+\listing{user_mode_emulation, User Mode Emulation Definition}
+~~~{.vadl}
+[os : linux] 
+user mode emulation UME from RV64IM with ABI = {
```
AndreasKrall:
```suggestion
user mode emulation UME for RV64IM with ABI = {
```

AndreasKrall (approved):
Fixed typo, otherwise it is fine

docs/refman/tutorial.md
```diff
@@ -2002,3 +2002,67 @@ In addition, several behavioral aspects of the cache can be specified, such as w
...
+
+\listing{user_mode_emulation_exc, User Mode Emulation Definition Exceptions}
+~~~{.vadl}
+user mode emulation UME for RV64 with ABI = {
+  
+}
+~~~
+\endlisting
+
+It is also possible to select the desired OS in the .vadl file (e.g. Linux or BSD).
```
Jozott00: This section is not yet finished

Jozott00 (changes requested):
I think there are some parts missing, e.g. syscall number matching and other kinds of register definitions. E.g. using non-ABI registers, such as the actual source registers from the ISA, like X(1). Or should it only be possible to use alias registers defined in the ABI?

docs/refman/tutorial.md
arcane-quill: yes, I also thought I should finish this section before its ready to be merged

Jozott00: I disabled the auto-merge, so it is not accidentally merged automatically.

Jozott00:
An other thing I noticed: Typically syscalls are triggered by exceptions (which are triggered by instructions). And in QEMU the execution typically checks for a specific exception and state condition to determine whether a syscall was triggered or not.

Therefore, I think the `syscall instruction` is not really necessary and meaningful. I am not sure if there is any architecture that has a non-exception syscall trigger mechanism, but for now I would concentrate on exceptions. 

Speaking of exceptions, the exception definition alone won't be sufficient to recognize a syscall. E.g. the `sys/risc-v/rvcsr.vadl` which defines the `RV32Zicsr` uses only a single exception definition to handle all kinds of exceptions. The actual exception code is given as argument to the exception:
```
enumeration ExcCode: ExcCodeSize =
  { ILLEGAL_INSTR = 0x02
  , BREAKPOINT    = 0x03
  , M_ECALL       = 0x0B
}

exception Exc(cause: ExcCodeSize) = {
// ...
}
```
This means the syscall trap mechanism must be a combination of an initial exception + condition check on some values. In case of `RV32Zicsr` it would be sufficient to have access to the `cause` argument, e.g. like this:
```
[ condition : cause = ExcCode::M_ECALL ]
syscall exception = Exc
```
However, it might be necessary to also have access to the CPU state to evaluate the condition. Unfortunately I don't have enough knowledge about different syscall mechanisms used in practice. 
Also whether this should be an annotation like in the example above, or a more sophisticated definition depends on the potential complexity of such trigger mechanisms.

AndreasKrall:
> An other thing I noticed: Typically syscalls are triggered by exceptions (which are triggered by instructions). And in QEMU the execution typically checks for a specific exception and state condition to determine whether a syscall was triggered or not.
> 
It is the other way around. A syscall triggers a privileged exception changing the privilege level.
> Therefore, I think the `syscall instruction` is not really necessary and meaningful. I am not sure if there is any architecture that has a non-exception syscall trigger mechanism, but for now I would concentrate on exceptions.
> 
As it is the other way around, the syscall instruction does make sense and syscalls can be implemented without exceptions. Of course syscalls and exceptions are caught in privileged mode in a similar way. But in UME there is no privileged mode because the privileged mode is emulated. It is possible to implement UME without exceptions, but because the privileged exception handling is emulated and shared, they should be implemented together.

For exceptions you also have to distinguish between user mode exceptions and privileged exceptions. User mode exceptions can be caught by a user mode exception handler, privileged exceptions change to privileged mode. User mode exceptions are caught be an exception handler written in assembly language which is invoked from a certain address. I believe that only few of our VADL specifications have the user mode exception handling implemented. It should be defined in the processor section.

I did not look into the privileged instruction set of RISC-V to know the details. But exceptions are only implemented in a very rudimentary way in our VADL specifications.
Privileged mode is not defined at all. In the AArch64 spec there is some privilege level checking for access to system registers. 
> Speaking of exceptions, the exception definition alone won't be sufficient to recognize a syscall. E.g. the `sys/risc-v/rvcsr.vadl` which defines the `RV32Zicsr` uses only a single exception definition to handle all kinds of exceptions. The actual exception code is given as argument to the exception:
> 
> ```
> enumeration ExcCode: ExcCodeSize =
>   { ILLEGAL_INSTR = 0x02
>   , BREAKPOINT    = 0x03
>   , M_ECALL       = 0x0B
> }
> 
> exception Exc(cause: ExcCodeSize) = {
> // ...
> }
> ```
> 
> This means the syscall trap mechanism must be a combination of an initial exception + condition check on some values. In case of `RV32Zicsr` it would be sufficient to have access to the `cause` argument, e.g. like this:
> 
> ```
> [ condition : cause = ExcCode::M_ECALL ]
> syscall exception = Exc
> ```

I do not think that a specification in such a way is useful. I have to think deeper about it and read the processor documentations.
> However, it might be necessary to also have access to the CPU state to evaluate the condition. Unfortunately I don't have enough knowledge about different syscall mechanisms used in practice. Also whether this should be an annotation like in the example above, or a more sophisticated definition depends on the potential complexity of such trigger mechanisms.

docs/refman/tutorial.md
AndreasKrall: > I think there are some parts missing, e.g. syscall number matching and other kinds of register definitions. E.g. using non-ABI registers, such as the actual source registers from the ISA, like X(1). Or should it only be possible to use alias registers defined in the ABI?

I think it is not necessary to directly support ISA registers. I think it is always a good idea to use an ABI. The problem with our current ABI is that it is meant only for the compiler and requires things which do not make sense in the UME like sequences and certain registers.

Jozott00:
> It is the other way around. A syscall triggers a privileged exception changing the privilege level.

The trigger mechanism I was talking about was on the QEMU implementation side. In QEMU the UME is an execution loop (similar to full system emulation), where each iteration execution checks for an interrupt or exception. 
So in this loop body there is essentially a switch over all exceptions that have to be handled by the UME. One of them is handling the syscall call exception (in RISC-V ECALL) which will trigger the underlying syscall handling.

This means from the generator perspective, it doesn't matter which instruction causes the exception, it only matters which exception (under which condition) should be recognized as syscall. Therefore, the `syscall instruction` per se, has no relevance to the generated execution loop that looks for a user-defined exception for syscall handling.

As we control the whole generation, we could of course also handle syscalls by defining the `syscall instruction` and automatically throwing a synthetic exception in the QEMU implementation. However, this is way more implicit and does not allow as much control by the user (e.g. if the syscall is only triggered on some specific condition in the syscall instruction). And because (I think) most architectures are using exceptions for syscalls, I would concentrate on them and later add alternative mechanisms when we need them.

Jozott00:
@arcane-quill I had a meeting with @AndreasKrall to discuss some aspects of the UME VADL definition and QEMU generation.

We habe agreed on the following:
- We use `syscall instruction` instead of a `syscall exception` as proposed above. This means, that under the hood, the UME generator will replace the instruction's behavior with a raise of a synthetic SYSCALL exception, which is used in the the cpu loop to recognize the syscall.
- For now, we skip the exception behavior definition, as it is not yet clear how to define them in VADL. So the exception section in the docs can be stated as "work in progress".
- The ABI definition is mandatory for the UME definition, as it contains the stack pointer. However, registers, used by the UME definition, such as `syscall argument` can be both, an `alias register` from the ABI and an actual register from the ISA definition.
- For now all exceptions that are not the synthetic syscall exception lead to an UME abort.
- We will need some kind of table in VADL to define the linux syscall numbers, as they may differ between architectures. I’ll take a closer look at the `syscall.tbl` files in the upstream QEMU code so that we can develop a suitable design for it.

@arcane-quill If anything is unclear, please let me know.

arcane-quill:
I was thinking about how to handle the syscalls, maybe we can use the specific vadl files to define the non-generic syscalls and then overwrite/add them to the existing table? Maybe we can put those existing syscalls in a separate vadl file to make it easier to generate everything correctly via the interpolation? @Jozott00

Jozott00:
> I was thinking about how to handle the syscalls, maybe we can use the specific vadl files to define the non-generic syscalls and then overwrite/add them to the existing table? 

You mean the syscall numbers right? 
Yeah, I would define some kind of `syscall table` that can be extended, so one table is able to overwrite syscalls of the super table. And VADL provides a built-in generic table that has all generic syscalls already defined. 
Something like 
```
syscall table riscv64_syscalls extending VADL::GenericSysCallTable {
  // define syscall overwrites and add specific ones ...
} 

syscall table some_other_table extending riscv64_syscalls {
  // inherit riscv64 ones
}
```

However, I think we should not interpolate the existing table, but render the table from scratch based on the VADL definition.

> Maybe we can put those existing syscalls in a separate vadl file to make it easier to generate everything correctly via the interpolation? @Jozott00

I’m not entirely sure what you mean by `to make it easier to generate everything correctly via the interpolation`, but generally, placing the syscall table in a separate VADL file and then importing it is probably the best approach. However, it’s not mandatory (it’s a user’s choice).

Jozott00:
I just saw your suggested approach in the documentation (with the enumeration). While an enumeration would also work, it has a drawback: it lacks a declaration side type, which prevents the LSP from recognizing it as a syscall table. Consequently, the LSP cannot provide autocompletion or on-the-fly diagnostics (at least in most cases). 

I personally don't care that much about those features, but that is just a consideration. Of course, the benefit of such enumeration is that it’s already implemented in the frontend, so there is no need to add the syscall table definition to the language.

Also you have to check if there are syscalls that need more than just the number assignment. If that is the case, the enumeration is not a sufficient type.

@AndreasKrall if you are fine with enumerations for syscall tables, I am too.

AndreasKrall:
> I just saw your suggested approach in the documentation (with the enumeration). While an enumeration would also work, it has a drawback: it lacks a declaration side type, which prevents the LSP from recognizing it as a syscall table. Consequently, the LSP cannot provide autocompletion or on-the-fly diagnostics (at least in most cases).
> 
> I personally don't care that much about those features, but that is just a consideration. Of course, the benefit of such enumeration is that it’s already implemented in the frontend, so there is no need to add the syscall table definition to the language.
> 
> Also you have to check if there are syscalls that need more than just the number assignment. If that is the case, the enumeration is not a sufficient type.
> 
> @AndreasKrall if you are fine with enumerations for syscall tables, I am too.

I suggest to start with the enumeration, it reuses existing constructs and we can expand it in the future, when we see that we have further requirements.

Jozott00 (approved):
LGTM

docs/refman/tutorial.md
```diff
@@ -2002,3 +2002,84 @@ In addition, several behavioral aspects of the cache can be specified, such as w
...
+     sys_riscv_hwprobe      = 258
+  , sys_riscv_flush_icache  = 259
+}
+~~~
+\endlisting
+
+The `enumeration` keyword is used to define a set of named constants, in this case the 
+key-value pairs are the name of the syscall and its respective number. A generic syscall
+table already exists, so the table defined in the VADL file provides the architecture-specific syscalls 
+that should either be added to or overwrite the generic ones.
```
Jozott00, line 2073: @arcane-quill just noticed that you don't describe how to set the syscall table in the UME definition.

arcane-quill: you mean by adding something like "syscall table" (or syscall base table and syscall arch table to differentiate) within the Ume section?

Jozott00: Yes, somehow the user must be able to define that the enumeration definition is actually the syscall table. I think `syscall table` is fine for that.

PR #951, opened 2026-05-11
iss: Implement PC annotations [current], [next] and [next next]

BadGraphixD:
- Implemented `[current]`, `[next]` and `[next next]` on pc and alias pc definitions
- The pc subcalls (like `.current`) overrule the annotations

The offsets caused by the annotations are handled in the viam pass PcOffsetPass (it cannot be in the frontend, because there we have not yet determined which register (file) reads are counter accesses).

The offsets caused by the subcalls are handled in the frontend in BehaviorLowering, where the subcalls overwrite (or rather nullify) the offset of the annotations by also adding the equivalent negative value thereof.

The frontend does not yet support getting the instruction and memory byte size during behaviour lowering, so the pc subcalls use a hardcoded instruction size of 32 bit and byte size of 8 bit for now. This will change soon.


BadGraphixD:
@flofriday could you please review the first commit. Since I already merged #932, your requested changes from that review are here.

vadl/main/vadl/ast/TypeChecker.java
```diff
@@ -1603,13 +1603,26 @@ public Void visit(AliasDefinition definition) {
...
+            node -> node instanceof RegisterDefinition
+                || (node instanceof AliasDefinition aliasDef
+                && aliasDef.kind == AliasDefinition.AliasKind.REGISTER)
+        ));
+        var suggestions = Levenshtein.suggestions(targetIdent.pathToString(), candidates);
         throw addErrorAndStopChecking(
             error("Unknown alias source register", targetIdent.location())
                 .locationDescription(targetIdent.location(), "Unknown register `%s`.",
                     targetIdent)
+                .suggestions(suggestions)
```
flofriday, line 1676: **praise:** Awesome, thank you :)

flofriday:
> @flofriday could you please review the first commit. Since I already merged #932, your requested changes from that review are here.

Honestly, that whole situation is on me, I've wrote that review before you merged but forgot to submit it on the previous PR 😅
But thanks anyway for incorporating it here, looks good :)

Jozott00 (approved):
LGTM

PR #952, opened 2026-05-12
iss: Implement CPU-vector storage metadata foundation

Jozott00:
This PR is vector metadata and renaming only.

- Rename helper-only register execution class to CPU_VECTOR
- Add CPU-vector and gvec-capable register metadata
- Align CPU-vector state arrays and emit future offset helpers
- Add RV64V metadata/template regression coverage
- Preserve existing helper-call vector execution behavior


BadGraphixD (approved):
Looks good to me!

PR #954, opened 2026-05-12
setup: Add a no gc native image build

flofriday:


AndreasKrall:
For me this would be enough. This setup stores the epsilon build in the same build directory as the native build. This requires the copying of the binary if you want both native versions, one with serial GC and one with epsilon. What do you think about adding an additional build directory for the epsilon build? Than no copying is necessary if you want to use all three versions in parallel.

AndreasKrall (approved):
Fine for now, but perhaps we add later an additional build directory.

PR #955, opened 2026-05-13
ci: Remove redundant timestamp job in benchmark workflow

Jozott00:
- Directly set artifact name using `date` command, simplifying the process by removing the `timestamp` job dependency.


flofriday (approved):
Awesome, that's much better

PR #956, opened 2026-05-13
viam: Optimize Graph API to reduce memory consumption

Jozott00:
This PR optimizes the VIAM Graph API and node implementations to reduce memory consumption.

1. Add specialized node iterator to directly filter nodes for the `Graph.getNodes` API (200dafe0c0786c9da6c199ad31ac07533a3f1bc8)
2. Change some inefficient client side usages of getNodes that collect all nodes into a list for existential checks (c1342f790ccbe7a55a1aedbe95643e2140185ad4)
3. Change NodeList input/successor application implementations to copy-on-write instead of always creating a copy on every application (f77ad3169742d9db6dc466ae77b04ebe8075d275)


flofriday (approved):
That's awesome ✨🚀

PR #961, opened 2026-05-15
lcb: Implement generation of indirect-call patterns for aarch64

Tbswss:
This PR adds a new MachineInstructionLabel `BLR`, which describes instructions, which represent an unconditional jump with a register, which sets the linking register. Additionally, a new LlvmInstructionLowering-Strategy is added, which implements the generation of LLVM "PseudoCALLIndirect" and "BRIND" patterns for instructions labeled as `BLR`.

### DIFF
--- a/vadl/main/vadl/gcb/passes/MachineInstructionLabel.java
+++ b/vadl/main/vadl/gcb/passes/MachineInstructionLabel.java
@@ -94,6 +94,7 @@ public enum MachineInstructionLabel {
   UNCONDITIONAL JUMPS
    */
   JALR, //  unconditional indirect jump without immediate. With linking register.
+  BLR, //  unconditional indirect jump without immediate. With linking register (aarch64-style).
   JAL, // the difference between JAL and J is that JAL also writes a linking register.
   J, // unconditional jump with no linking register.
   JR, //  unconditional indirect jump without immediate. No linking register.
--- a/vadl/test/resources/snapshots/aarch64/InstrInfoTableGen.td
+++ b/vadl/test/resources/snapshots/aarch64/InstrInfoTableGen.td
@@ -14669,7 +14669,7 @@ let Inst{4-0} = op{4-0};
 let Inst{31-10} = op{26-5};
 let Inst{9-5} = rn{4-0};
 
-let isTerminator       = 1;
+let isTerminator       = 0;
 let isBranch           = 0;
 let isCall             = 0;
 let isReturn           = 0;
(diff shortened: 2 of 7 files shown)


vadl/test/resources/snapshots/aarch64/InstrInfoTableGen.td
```diff
@@ -14669,7 +14669,7 @@ let Inst{4-0} = op{4-0};
 let Inst{31-10} = op{26-5};
 let Inst{9-5} = rn{4-0};
 
-let isTerminator       = 1;
```
kper, line 14672: major: hmmm 

Is that really correct?

kper (changes requested):

vadl/test/resources/snapshots/aarch64/InstrInfoTableGen.td
Tbswss: The upstream compiler also marks "isTerminator = 0" (implicitly) for BLR, as far as I can see:

```
let isCall = 1, Defs = [LR], Uses = [SP] in {
  def BLR : BranchReg<0b0001, "blr", []>;
  ...

class BaseBranchReg<bits<4> opc, dag oops, dag iops, string asm,
                    string operands, list<dag> pattern>
    : I<oops, iops, asm, operands, "", pattern>, Sched<[WriteBrReg]> {
  let Inst{31-25} = 0b1101011;
  ....
  let Inst{4-0}   = 0b00000;
}

class BranchReg<bits<4> opc, string asm, list<dag> pattern>
    : BaseBranchReg<opc, (outs), (ins GPR64:$Rn), asm, "\t$Rn", pattern> {
  bits<5> Rn;
  let Inst{9-5} = Rn;
}
```

And according to [this](https://discourse.llvm.org/t/does-call-instruction-llvm-ir-terminate-a-basic-block/23589) and [this](https://llvm.org/docs/LangRef.html#terminator-instructions) (even though that talks about LLVM IR) call-instructions are not seen as terminators.

kper: fine for me

vadl/main/vadl/gcb/passes/MachineInstructionLabel.java
```diff
@@ -94,6 +94,7 @@ public enum MachineInstructionLabel {
   UNCONDITIONAL JUMPS
    */
   JALR, //  unconditional indirect jump without immediate. With linking register.
+  BLR, //  unconditional indirect jump without immediate. With linking register (aarch64-style).
```
kper, line 97: question: whats the difference to JALR? Can't we merge it?

Tbswss: The comment for JALR is a bit misleading. JALR has three operands: the register, in which the return address is stored, the register to which to jump to, and an immediate, which gets added to the register before branching.

BLR on the other hand only takes in a single register, which describes the location to branch to.

Initially I tried to label aarch64-BLR instructions as JALR too, but I found this a bit unclean, because the aarch64 BLR does not correctly model the JALR instruction. And in the respective LlvmLoweringStrategy for JALR instructions, the assumption that an instruction which is marked as JALR has an immediate would have been broken. So a separate label seemed to be the best option.

kper: Fair enough

But then the comment above is wrong?

Tbswss: Yes, that comment is wrong, because the associated "findJALR" method does not check whether there are immediates or not, just if there is an operation (ADD, SUB, etc.). Furthermore, the LlvmLower...Strategy expects an immediate to be present.

But I am refactoring / adding comments in [that PR](https://github.com/OpenVADL/openvadl/pull/964) already, I would leave it as is here.

kper (approved):

PR #963, opened 2026-05-15
chore: Refactor tuple nodes and access

rascmatt:
As discussed in the PR (See #918 and [this thread](https://github.com/OpenVADL/openvadl/pull/918#discussion_r3194673715)) regarding group annotations, this refactors the current usages of the Tuple nodes to a more general `Struct` which allows field access as well.

Tuples are currently used in two places, namely `LetExpr`essions and `LetStatements` as well as `Status` types. For `Let` nodes, the result of the value expression may have a `StructType` with fields `result` and `status`. E.g.:

```vadl
let next, status = VADL::adds(PC, 4 as Bits<32>) in
  if status.zero then
  PC := next
```

During behavior lowering the name `next` will be mapped to a `StructGetFieldNode`, accessing the field `result` on the expression `VADL::adds(PC, 4 as Bits<32>)`.

`StatusType` now exposes fields `negative`, `zero`, `carry` and `overflow`, as opposed to positional indexing.


AndreasKrall:
I have noticed the use of streams with lambda where I am quite sure that it is slower than a for loop and it generates a list of one or two elements. Is this really necessary? I am not convinced that the functional style is easier to understand.

Regarding performance you should read:
[How Fast Are Streams Really?](https://belief-driven-design.com/how-fast-are-streams-really-ad9cc/)

rascmatt:
@AndreasKrall 
> I have noticed the use of streams with lambda where I am quite sure that it is slower than a for loop and it generates a list of one or two elements. Is this really necessary? I am not convinced that the functional style is easier to understand.
> 
> Regarding performance you should read: [How Fast Are Streams Really?](https://belief-driven-design.com/how-fast-are-streams-really-ad9cc/)

Thank you, that was a good read. 

I agree that in most streams introduced in this commit they're mapping / filtering over lists of 2-4 elements, which will be slower than using a for loop instead. I replaced those with loops, even if they may be not be on a critical path.

In general I personally like the functional style, but I do agree that it can have the opposite effect in some cases. E.g. this one I think benefits from the stream: (both examples are form this PR)

```diff
-return "Struct<%s>".formatted(types.entrySet().stream()
-        .map(e -> e.getKey() + ": " + e.getValue().name())
-        .collect(Collectors.joining(", ")));
+final var sb = new StringBuilder("Struct<");
+
+boolean first = true;
+for (var entry : types.entrySet()) {
+  if (!first) {
+    sb.append(", ");
+  }
+  sb.append(entry.getKey())
+      .append(": ")
+      .append(entry.getValue().name());
+  first = false;
+}
+
+sb.append(">");
+return sb.toString();
```

And this is definitely less readable than the loop-based alternative:

```diff
-values.entrySet().stream()
-    .collect(Collectors.toMap(
-        Map.Entry::getKey,
-        e -> e.getValue().type(),
-        (a, b) -> b,
-        LinkedHashMap::new))
+final Map<String, Type> types = new LinkedHashMap<>();
+for (var entry : values.entrySet()) {
+  types.put(entry.getKey(), entry.getValue().type());
+}
+return types;
```

AndreasKrall:
```
  String mapName(String name) {
    var valType = valueExpr.type;
    if (!(valType instanceof StructType struct)) {
      throw new IllegalStateException("Expected StructType but got " + valType);
    }
    final List<String> fields = struct.fieldNames();
    for (var i = 0; i < identifiers.size(); i++) {
      if (name.equals(identifiers().get(i).name)) {
        return fields.get(i);
      }
    }
    throw new IllegalStateException("Let statement does not have a name `%s`.".formatted(name));
  }
```
This code appears once in Expr and once in Statement. I do not know if it is easy to use a common function as there are different this pointers. But if you do not share the code, please change the error message in Expr from `statement` to `expression` and check all comments in Expr where `statement` should be replaced by `expression`.

AndreasKrall:
```
-return "Struct<%s>".formatted(types.entrySet().stream()
-        .map(e -> e.getKey() + ": " + e.getValue().name())
-        .collect(Collectors.joining(", ")));
+final var sb = new StringBuilder("Struct<");
+
+boolean first = true;
+for (var entry : types.entrySet()) {
+  if (!first) {
+    sb.append(", ");
+  }
+  sb.append(entry.getKey())
+      .append(": ")
+      .append(entry.getValue().name());
+  first = false;
+}
+
+sb.append(">");
+return sb.toString();
```
Here the use is not performance critical. Without nearly no loss of performance you could change
```
  sb.append(entry.getKey())
      .append(": ")
      .append(entry.getValue().name());
```
to
```
  sb.append(entry.getKey() + ": " + entry.getValue().name());
```

vadl/main/vadl/iss/passes/common/opDecomposition/IssOpDecompositionPass.java
```diff
@@ -424,14 +424,17 @@ private void handleSliceAcrossUpperLowerBoundary(BuiltInCall longMul, SliceNode
     var argWidth = arg1.type().asDataType().bitWidth();
 
     var targetType = Type.bits(argWidth);
-    var tupleType = Type.tuple(targetType, targetType);
+    var structType = Type.struct(
+        "fst", targetType,
+        "snd", targetType
```
Jozott00: I am not sure if `fst` and `snd` are the most obvious names for this tuple like struct ^^ 
I would prefer either "first" and "second" or just "1" and "2".

AndreasKrall: Would not be the correct names here `low` and `high` or `lower` and `upper`?

Jozott00: True, in this particular case `low` `high` are the best fitting names.

rascmatt: Done, changed to `low` and `high`.

vadl/main/vadl/viam/passes/statusBuiltInInlinePass/Inliner.java
```diff
@@ -183,12 +189,12 @@ private void initUsers(BuiltInCall builtInCall) {
...
-            case 1 -> zeroUsers.add(getField);
-            case 2 -> carryUsers.add(getField);
-            case 3 -> overflowUsers.add(getField);
+        if (usage instanceof StructGetFieldNode getField) {
+          switch (getField.field()) {
+            case NEGATIVE -> negativeUsers.add(getField);
+            case ZERO -> zeroUsers.add(getField);
+            case CARRY -> carryUsers.add(getField);
+            case OVERFLOW -> overflowUsers.add(getField);
             default -> throw new ViamGraphError("User of status tuple accesses non existing index.")
```
AndreasKrall: `"User of status tuple accesses non existing index."` is not a useful error message any more.

Please check also all other default cases in struct switches.

rascmatt: Right, thank you. I adapted the message.

AndreasKrall: I still found the following messages:
```
"Tuple unpacking only works on tuples but the type was `%s`"

"Get tuple index out of bounds."
```
Do a file search on the word `tuple` in all cases.

rascmatt: I should have all occurrences now, I still found a few others as well.

AndreasKrall (approved):
Looks good to me now.

PR #965, opened 2026-05-16
frontend: Fix macro expansion missing encoding copy

flofriday:


AndreasKrall (approved):
Looks good to me, but is there a test case which does not trigger an error anymore?

flofriday:
I don't think it's possible, because except for the source location the copy doesn't change anything that could be observed and even the source location isn't often used :/

PR #966, opened 2026-05-17
frontend: More even further performance improvements

flofriday:
Again even more performance improvements focused on memory pressure.

**Protocol:**
```
master:
1804M

Improve synthetic cast expression creation
1790M

Avoid redundant string concat and spliting
1734M

Optimize pathToSegments method
1730M

Optimize pathToString method on IdentifierPath:
1724M

Optimize hot BuiltinTable:argsCompatible method:
1651M
```

### DIFF
--- a/vadl/test/resources/frontend-snapshots/lowering/aliasConstantRegister.vadl
+++ b/vadl/test/resources/frontend-snapshots/lowering/aliasConstantRegister.vadl
@@ -294,7 +294,7 @@ assembly Test = ""
 //    . : ' Inputs: A, B, C, D
 //    . : ' Outputs: A, B, C, D
 //    . : ' Encoding: encoding
-//    . : ' Assembly: TEST::Test::assembly
+//    . : ' Assembly: assembly
 //    . : ' Behavior Graph:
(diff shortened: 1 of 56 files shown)


vadl/test/resources/frontend-snapshots/lowering/aliasConstantRegister.vadl
```diff
@@ -312,10 +312,10 @@ assembly Test = ""
 //    . : ' | n13 WritesRegisterTensor in: (n3 n4 n12)     succ: ()      reg: D   resource: U
 //    . : ' | n14 InstrEnd             in: (n2 n7 n10 n13) succ: ()      effects: 4
 //    . : ' | n15 Start                in: ()              succ: (n14)   next: set
-//    . : ' Assembly name: "TEST::Test::assembly"
```
Jozott00, line 315: question: Did you just adjust the VIAM snapshot dump or do you omit the scope parts from the assembly definition identifier? 

I would rather like to keep the identifiers in the VIAM as is.

flofriday: So the VIAM DUMP apperently only prints the last part of the names by default, however, there were some bugs earlier in the lowering where not _everywhere_ the text was split along the `::`.

Jozott00 (approved):

PR #967, opened 2026-05-18
gcb: Add comments / documentation to the MachineInstructionLabel enum.

Tbswss:
This PR adds documentation to the MachineInstructionLabel-Enum, which represents classifications which are assigned to instructions.

Furthermore, MachineInstructionLabels for Branches and Conditional-Moves were adapted. See [this](https://github.com/OpenVADL/openvadl/pull/967#issuecomment-4682621995) comment.


Tbswss:
Contrary to what I have said at yesterday's jour-fixe (18.05), currently only one classification via the MachineInstructionLabels per instruction is possible. @AndreasKrall

AndreasKrall:
Looks only partially fine to me.

The status branches need more care. You have to change the unsigned names. I would use carry set and carry clear. The reason is that there are two different kind of subtractions with status register, one with carry and one with borrow. See the VADL reference manual and/or Wikipedia where the difference is explained in detail. Depending on the kind of subtraction used, there is the opposite meaning of the carry flag. As many users can fall into this trap, please add an exact explanation in the comment. In the future we want to support both ARM and x86 and they have the opposite use of the flag registers. To understand ARM, you just can check the VADL source code, for understanding x86 I suggest you take a look into the AMD64.pdf from the Übersetzerbauübung on page 11.
 
Furthermore, for the status branches I would add negative (less than zero) and positive (greater or equal to zero) which are based on the sign bit too. I also would add the overflow branches, although they will not be used.

AndreasKrall:
In theory a user can define an ISA which supports both kind of subtractions. Therefore, I would design the classification algorithm in such a way, that the classification would also work in this kind of specifications.

Tbswss:
I removed all branch and conditional-move instructions with unsigned conditions and replaced them by more "specific" carry-set, carry-clear, negative-set, negative-clear, overflow-set, overflow-clear variants.

In total, following machine-instruction-labels were added now:

- B_CS
  - carry set
- B_CC
  - carry clear
- B_NS
  - negative set (sign bit set)
- B_NC
  - negative clear (sign bit clear)
- B_VS
  - overflow set
- B_VC
  - overflow clear

and equivalent variantions for conditional-moves

- CSEL_CS_64
- CSEL_CC_64
- ...

Furthermore, I removed the instructions with unsigned-conditons from the generated LLVM code, as these conditions cannot be deduced solely by the status-registers in a "platform independent" way apparently (as described above).

AndreasKrall (approved):
Looks good to me.

PR #970, opened 2026-05-20
lsp: Remove diagnostics debounce delay

flofriday:
Previously the delay was added to avoid over-reporting parsing errors during the writing of the document, however this is also done by state of the art IDEs like IntelliJ, so this commit removes it.

**Recording of default IntelliJ with Java**

https://github.com/user-attachments/assets/16858e72-cd22-4910-9234-ed55b1c1040c

**Recording of OpenVADL LSP in IntelliJ**

https://github.com/user-attachments/assets/5df4d308-e45c-470b-9905-6da8400916ac

**Recording of OpenVADL LSP in VSCode**

https://github.com/user-attachments/assets/8c184e14-2326-43e4-bef4-b592c44fdd0d

### DIFF
--- a/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
+++ b/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
@@ -62,14 +62,6 @@ import vadl.utils.SourceLocation;
  * Handles document-related features of the language server.
  */
 public class VadlTextDocumentService implements TextDocumentService {
-  /**
-   * How many milliseconds to wait before providing diagnostics for the latest change. If a new
-   * document version is pushed by the client within this time, the now outdated version will not
-   * receive diagnostics. This is intended to avoid showing transitory diagnostics while the
-   * developer is typing.
-   */
-  private static final int DIAGNOSTICS_DELAY_MS = 500;
-
   private static final Logger log = LoggerFactory.getLogger(VadlTextDocumentService.class);
 
   private final VadlLanguageServer server;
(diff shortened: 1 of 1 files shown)


vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
jaraonthe-dot-net: chore: remove mention of "debouncing"

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -229,13 +221,6 @@ private void publishDiagnostics(Document document, LspSnapshotFileSystem snapsho
...
 
     var unused = server.executor().submit(() -> {
-      try {
-        // TODO Consider to instead delay for remaining time *after* generating diagnostics, to have
-        //  more accurate delay timing
-        Thread.sleep(DIAGNOSTICS_DELAY_MS);
-      } catch (InterruptedException e) {
-        return;
-      }
       if (!documentVersionIsCurrent(document)) {
```
jaraonthe-dot-net: suggestion: Remove this version check - as we no longer do the delay, it is unlikely that the version will be outdated at this point (there's still the version check after diagnostics have been generated, so outdated-ness will still be caught).

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
jaraonthe-dot-net: nitpick: If you implement my suggestion above, then there is only one version check, thus the "(after)" qualifier can be removed here.
(I only added that so that the logs show which of the two version checks triggered.)

jaraonthe-dot-net (changes requested):

jaraonthe-dot-net (approved):

PR #972, opened 2026-05-21
parser: Improve invalid identifier errors

### LINKED ISSUE #82: Empty or Invalid Identifier Names should be disallowed

### LINKED ISSUE #127: ast: Invalid identifier when specifying functions
I get the following error

```
error: Parsing Error
     ╭──[vadl/test/resources/testSource/sys/risc-v/rv3264im.vadl:290:19]
     │
 290 │   function shift( value : Bits<64>, offset: Bit<32> ) -> Bits<12> = (value >> offset) as Bits<12>
     │                   ^ invalid identifier
     │
     The parser got confused at this point, probably because of a syntax error in your code.
```

when I add to `rv3264im.vadl`

```
  function shift( value : Bits<64>, offset: Bit<32> ) -> Bits<12> = (value >> offset) as Bits<12>
```

flofriday:
Closes: #82 
Closes: #127

### DIFF
--- a/vadl/main/vadl/ast/MacroExpander.java
+++ b/vadl/main/vadl/ast/MacroExpander.java
@@ -25,7 +25,6 @@ import java.util.Map;
 import java.util.Objects;
 import javax.annotation.Nullable;
 import vadl.error.Diagnostic;
-import vadl.error.DiagnosticList;
 import vadl.utils.RopeList;
 import vadl.utils.SourceLocation;
 
@@ -71,9 +70,9 @@ class MacroExpander
    */
   private Expr expandExprRaw(Expr expr) {
     var result = expr.accept(this);
-    if (!errors.isEmpty()) {
-      throw new DiagnosticList(errors);
-    }
+    //if (!errors.isEmpty()) {
+    // throw new DiagnosticList(errors);
+    //}
     if (result instanceof BinaryExpr binaryExpr && !binaryExpr.hasBeenReordered) {
       return ParserUtils.reorderBinaryExpr(binaryExpr);
     }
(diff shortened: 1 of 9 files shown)


AndreasKrall (changes requested):
Looks mostly fine to me. But I also took a look into ParserUtils and vadl.ATG. The lists seem similar, but are not in the same order. Please change ParserUtils and/or vadl.ATG that both are in the same order. Additionally you should add some more allowed keywords, there has been some tokens added recently.

flofriday:
I've just looked up git blame and I don't think any keywords were added in the last year.

AndreasKrall (approved):
Looks fine now.

PR #973, opened 2026-05-21
lcb: Adapt LLVM register - subregister hierarchy

Tbswss:
This PR adapts the register - subregister relation in the generated LLVM register-classes. Currently, all alias registers of a given base-register-class are direct subregisters of this base-register-class and therefore siblings. This causes problems, when one subregister-class wants to access another subregister-class of the same base-register-class.

With the new changes, the alias-registers are arranged in a more tree-like form, where smaller alias-registers are subregisters of wider alias-registers (when applicable).

Currently, following vadl specification

```
register        S : Index -> BitsX

alias register  R = S(*)(31..0)
alias register  X = S
alias register  W  = X(*)(31..0)
```

generates following register-class hierarchy for LLVM:

- S
  - X
  - W
  - R

One problem with this is, that there is no way for the `X` register-class to access the `W` register-class directly. Instructions, which generate `X` registers, which then need truncation to `W` registers, fail during clang-runtime.

With the new changes, following hierarchy is created:

- S
  - X
    - R
      - W

This way, `S`, as well as `X`, are able to access 32-bit subregisters, which are all defined on the same base-register `S`. Alias / Subregisters are not naively nested under each other, only when they are compatible / valid.

Following vadl specification (which is not yet supported by the viam-lowering):

```
register        S : Index -> BitsX

alias register  R = S(*)(31..0)
alias register  X = S
alias register W  = X(*)(31..0)
alias register W_HI  = X(*)(63..32)
```

would result in

- S
  - X
    - W_HI
    - R
      - W

Furthermore, llvm-truncation-lowering was removed and replaced by explicit `trunc` patterns for all register-classes and their respective (in)direct subregisters. This is similar to how the upstream aarch64 compiler handles truncation of general purpose registers.


kper:
I need to look into that. Give me a day or two

kper (approved):
I will approve to let you go on with that approach and see where it leads you.

I think it's ok that you rearranged the registers in a way that you can continue. However, you didn't fix the underlying issue with sub registers and their relationships to their parents. It's just not as arbitrary as VADL requires it to be.

Also I am quite surprised that you chose the TableGen way. It's fine but it might hurt you since TableGen is very unhandy and hard to debug.

Anyways, I will unblock you and let's see how it goes.

kper (approved):

PR #974, opened 2026-05-21
chore: Improve the visibility of generated Coco/R files

### LINKED ISSUE #749: Rename Parser.java and Scanner.java
New developers are constantly tricked by them as it's not obvious enought that they are autogenerated.

I propose something like "ParserAutoGenerated.java" or something like it.

flofriday:
By adjusting the gradle build we can tell IntelliJ that the autogenrated files are that. Which makes the IDE add a huge banner.

There are also now comments that better explain how the files are generated.

Closes: #749

**Screenshots:**

Screenshot 2026-05-21 at 18 20 50

Screenshot 2026-05-21 at 18 22 13


Jozott00 (approved):
Thanks!

PR #976, opened 2026-05-22
frontend: Group Array Type

rascmatt:
The group definition's name can be used in expressions (within group annotations) to access the matched instructions / operations at runtime. In addition it exposes a `length` and `bitLength` field.

```vadl
VLIW.length // number of instructions in the group
VLIW.bitLength // number of bits in the group

VLIW(1) // the second instruction in the group
VLIW(VLIW.length - 1) // the last instruction in the group
```

I introduced a 'GroupType' for such group references. Since we're able to upper-bound the `length` and `bitLength`, we can also give these properties a `UIntType` type with some maximal width.

### DIFF
--- a/vadl-frontend/main/vadl/ast/GroupDefUtils.java
+++ b/vadl-frontend/main/vadl/ast/GroupDefUtils.java
@@ -18,7 +18,9 @@ package vadl.ast;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import vadl.ast.nodes.Group;
+import vadl.ast.nodes.InstructionDefinition;
 import vadl.ast.nodes.OperationDefinition;
 import vadl.ast.nodes.RangeExpr;
 
@@ -31,6 +33,10 @@ public interface GroupDefUtils {
    * Collect all operations occurring in a group expression.
    */
   class OperationCollector implements Group.GroupVisitor<List<OperationDefinition>> {
+
+    private OperationCollector() {
+      // Direct instantiation not allowed
+    }
 
     /**
      * Collect all operations occurring in a group expression.
@@ -126,6 +132,68 @@ public interface GroupDefUtils {
       }
       final var constant = constantEvaluator.eval(range.to);
       return constant.value().intValueExact();
+    }
+
+  }
+
+  /**
+   * Determine the maximum bitlength of a group expression.
+   */
+  class GroupExprBitLengthCollector implements Group.GroupVisitor<Integer> {
+
+    private final ConstantEvaluator constantEvaluator;
+
+    private GroupExprBitLengthCollector(ConstantEvaluator constantEvaluator) {
+      this.constantEvaluator = constantEvaluator;
+    }
+
+    static int maxBitLength(ConstantEvaluator constantEvaluator, Group group) {
+      return group.accept(new GroupExprBitLengthCollector(constantEvaluator));
+    }
+
+    @Override
+    public Integer visit(Group.Sequence seq) {
+      int len = 0;
+      for (Group group : seq.groups) {
+        len += group.accept(this);
+      }
+      return len;
+    }
+
+    @Override
+    public Integer visit(Group.Alternative alt) {
+      int len = 0;
+      for (Group.Sequence seq : alt.sequences) {
+        len = Math.max(len, seq.accept(this));
+      }
+      return len;
+    }
+
+    @Override
+    public Integer visit(Group.Permutation perm) {
+      int len = 0;
+      for (Group.Sequence seq : perm.sequences) {
+        len += seq.accept(this);
+      }
+      return len;
+    }
+
+    @Override
+    public Integer visit(Group.Literal lit) {
+
+      int bitLen = 0;
+      for (InstructionDefinition insn : lit.getOperation().instructions) {
+        final int formatWidth =
+            Objects.requireNonNull(insn.formatNode).type().asDataType().bitWidth();
+        bitLen = Math.max(bitLen, formatWidth);
+      }
+
+      if (!(lit.size instanceof RangeExpr range)) {
+        return bitLen;
+      }
+
+      final var constant = constantEvaluator.eval(range.to);
+      return bitLen * constant.value().intValueExact();
     }
 
   }
--- a/vadl-frontend/main/vadl/ast/GroupDefUtils.java
+++ b/vadl-frontend/main/vadl/ast/GroupDefUtils.java
@@ -1,0 +1,132 @@
+// SPDX-FileCopyrightText : © 2026 TU Wien <vadl@tuwien.ac.at>
+// SPDX-License-Identifier: GPL-3.0-or-later
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package vadl.ast;
+
+import java.util.ArrayList;
+import java.util.List;
+import vadl.ast.nodes.Group;
+import vadl.ast.nodes.OperationDefinition;
+import vadl.ast.nodes.RangeExpr;
+
+/**
+ * Utilities for working with group definitions.
+ */
+public interface GroupDefUtils {
+
+  /**
+   * Collect all operations occurring in a group expression.
+   */
+  class OperationCollector implements Group.GroupVisitor<List<OperationDefinition>> {
+
+    /**
+     * Collect all operations occurring in a group expression.
+     *
+     * @param group the group expression
+     * @return the list of operations
+     */
+    public static List<OperationDefinition> operations(Group group) {
+      return group.accept(new OperationCollector());
+    }
+
+    @Override
+    public List<OperationDefinition> visit(Group.Sequence seq) {
+      final var operations = new ArrayList<OperationDefinition>();
+      for (Group group : seq.groups) {
+        operations.addAll(group.accept(this));
+      }
+      return operations;
+    }
+
+    @Override
+    public List<OperationDefinition> visit(Group.Alternative alt) {
+      final var operations = new ArrayList<OperationDefinition>();
+      for (Group.Sequence sequence : alt.sequences) {
+        operations.addAll(sequence.accept(this));
+      }
+      return operations;
+    }
+
+    @Override
+    public List<OperationDefinition> visit(Group.Permutation perm) {
+      final var operations = new ArrayList<OperationDefinition>();
+      for (Group.Sequence sequence : perm.sequences) {
+        operations.addAll(sequence.accept(this));
+      }
+      return operations;
+    }
+
+    @Override
+    public List<OperationDefinition> visit(Group.Literal lit) {
+      return List.of(lit.getOperation());
+    }
+
+  }
+
+  /**
+   * Determine the maximum length of a group expression, i.e., the maximum number of literals that
+   * can be matched by the group.
+   */
+  class GroupExprLengthCollector implements Group.GroupVisitor<Integer> {
+
+    private final ConstantEvaluator constantEvaluator;
+
+    private GroupExprLengthCollector(ConstantEvaluator constantEvaluator) {
+      this.constantEvaluator = constantEvaluator;
+    }
+
+    static int maxLength(ConstantEvaluator constantEvaluator, Group group) {
+      return group.accept(new GroupExprLengthCollector(constantEvaluator));
+    }
+
+    @Override
+    public Integer visit(Group.Sequence seq) {
+      int len = 0;
+      for (Group group : seq.groups) {
+        len += group.accept(this);
+      }
+      return len;
+    }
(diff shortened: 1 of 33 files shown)


Jozott00:
I'll wait with the review until #918 got merged, so it becomes easier to see this PR's changes.

Jozott00 (commented):
Since this is unrelated to VIAM or ISS, I’ll leave the approval to @flofriday.

rascmatt:
It will need to be added to the VIAM as well, but I didn't get to that yet. I switched it to a Draft PR, and amend it later.

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidGroupReference.vadl
```diff
@@ -0,0 +1,53 @@
...
+//       │
+//       Arrays only have the fields `length` or `bitLength`
+//
+//  error: Invalid annotation expression
+//       ╭── test/resources/frontend-snapshots/typechecker/invalidGroupReference.vadl:27:13
+//       │
+//    27 │   [assert : VLIW.a = 0b0]
+//       │             ^^^^^^^^^^^^ Unable to determine type
+//       │
+//       help: Check additional errors which may have caused the type to be missing
```
flofriday: **nitpick:** I don't like this error, I think we can completely remove it as it doesn't add any useful information.

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidGroupReference.vadl
```diff
@@ -0,0 +1,53 @@
...
+
+//  Reported Diagnostics:
+//
+//  error: Unknown array access `a`
+//       ╭── test/resources/frontend-snapshots/typechecker/invalidGroupReference.vadl:27:13
+//       │
+//    27 │   [assert : VLIW.a = 0b0]
+//       │             ^^^^^^
+//       │
+//       Arrays only have the fields `length` or `bitLength`
```
flofriday: **question:** Will _this_ also be called an "Array" in the docs? Or is it something more specific like a Group Array?

I'm mostly asking because it could be confusing for users to name them Array when they only exist in a subsection of the language when the more useful analog from other languages is a Tensor.

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidGroupElemField.vadl
```diff
@@ -0,0 +1,52 @@
...
+//  Reported Diagnostics:
+//
+//  error: Unknown format field `c`
+//       ╭── test/resources/frontend-snapshots/typechecker/invalidGroupElemField.vadl:27:13
+//       │
+//    27 │   [assert : VLIW(0).c = 0b0]
+//       │             ^^^^^^^^^
+//       │
+//       Intersection format `(AType ∩ BType)` doesn't have any field with this name
+//       help: Did you maybe mean one of: a, b
```
flofriday, line 41: **praise This is very readable, and a great error message. 🎉

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidGroupArrayMultiIndex.vadl
```diff
@@ -0,0 +1,51 @@
...
+
+//  Reported Diagnostics:
+//
+//  error: Type Mismatch
+//       ╭── test/resources/frontend-snapshots/typechecker/invalidGroupArrayMultiIndex.vadl:27:13
+//       │
+//    27 │   [assert : VLIW(1)(2).a = 0b0]
+//       │             ^^^^^^^
+//       │
+//       Type `a: Bits<20>, b: Bits<5>` cannot be indexed or sliced
```
flofriday: **minor:** I don't love the Type description, maybe `GroupType<a: Bits<20>, b: Bits<5>>` makes it clearer?
I think we have something similar for FormatTypes.

vadl/main/vadl/types/Type.java
```diff
@@ -379,6 +379,22 @@ public static <T extends Type> DataType constructDataType(Class<T> typeClass, in
...
+   *
+   * @param elementType   the type of the elements of the array.
+   * @param lengthType    the type of the length of the array. Must be an unsigned integer type.
+   * @param bitLengthType the type of the bit length of the array. Must be an unsigned integer type.
+   * @return the array type.
+   */
+  public static ArrayType array(Type elementType, UIntType lengthType, UIntType bitLengthType) {
+    var hash = Objects.hash(elementType, lengthType, bitLengthType);
+    return arrayTypes
+        .computeIfAbsent(hash, k -> new ArrayType(elementType, lengthType, bitLengthType));
```
flofriday: **issue:** This has hash-collisions. You need to wrap the contents in a datastructure, like a `Triple`.

vadl/main/vadl/types/Type.java
```diff
@@ -379,6 +379,22 @@ public static <T extends Type> DataType constructDataType(Class<T> typeClass, in
...
+
+  /**
+   * Returns an array type with the given element type.
+   *
+   * @param elementType   the type of the elements of the array.
+   * @param lengthType    the type of the length of the array. Must be an unsigned integer type.
+   * @param bitLengthType the type of the bit length of the array. Must be an unsigned integer type.
+   * @return the array type.
+   */
+  public static ArrayType array(Type elementType, UIntType lengthType, UIntType bitLengthType) {
```
flofriday: **Question:** Does this type ever land in the VIAM? 
If not we should consider moving it to the AST package.

vadl/main/vadl/types/GroupType.java
```diff
@@ -0,0 +1,62 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package vadl.types;
+
+/**
+ * An array type, as used in group annotations.
+ * E.g: <pre>
+ * [assert : VLIW.length <= 2]
+ * group VLIW = O1.O2
```
flofriday, line 23: **question:** Is that how we define a group? What does `O1.O2` mean?

vadl/main/vadl/types/ArrayType.java
```diff
@@ -0,0 +1,62 @@
...
+package vadl.types;
+
+/**
+ * An array type, as used in group annotations.
+ * E.g: <pre>
+ * [assert : VLIW.length <= 2]
+ * group VLIW = O1.O2
+ * </pre>
+ */
+public class ArrayType extends Type {
```
flofriday: **question:** How does a ArrayType differ from an TensorType?

vadl-frontend/main/vadl/ast/GroupDefUtils.java
```diff
@@ -0,0 +1,192 @@
...
+  class OperationCollector implements Group.GroupVisitor<List<OperationDefinition>> {
+
+    /**
+     * Collect all operations occurring in a group expression.
+     *
+     * @param group the group expression
+     * @return the list of operations
+     */
+    public static List<OperationDefinition> operations(Group group) {
+      return group.accept(new OperationCollector());
```
flofriday, line 48: **nitpick:** Since we don't want users to call the constructor and instead the static method here, let's make the constructor private.

vadl/main/vadl/ast/GroupDefUtils.java
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Determine the maximum length of a group expression, i.e., the maximum number of literals that
+   * can be matched by the group.
+   */
+  class GroupExprLengthCollector implements Group.GroupVisitor<Integer> {
+
+    private final ConstantEvaluator constantEvaluator;
+
+    public GroupExprLengthCollector(ConstantEvaluator constantEvaluator) {
```
flofriday: **nitpick:** Maybe this too should be a private constructor.

vadl/main/vadl/ast/GroupDefUtils.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * Determine the maximum bitlength of a group expression.
+   */
+  class GroupExprBitLengthCollector implements Group.GroupVisitor<Integer> {
+
+    private final ConstantEvaluator constantEvaluator;
+
+    public GroupExprBitLengthCollector(ConstantEvaluator constantEvaluator) {
```
flofriday: **nitpick:** Same private constructor here.

vadl-frontend/main/vadl/ast/TypeChecker.java
```diff
@@ -3629,17 +3654,18 @@ private void visitSliceIndexCall(CallIndexExpr expr, Type typeBeforeSlice,
...
-
-    if (!(typeBeforeSlice instanceof BitsType) && !(typeBeforeSlice instanceof TensorType)) {
-      var loc = expr.target.location().join(lastSlice.location);
-      addErrorAndStopChecking(error("Type Mismatch", loc)
-          .description("Only bit types can be sliced but the target was a `%s`", typeBeforeSlice)
-          .build());
-    }
-
     Type currType = typeBeforeSlice;
+    SourceLocation targetLoc = expr.location();
```
flofriday: **nitpick:** Calculating expression _can_ be expensive in hot loops so let's add it to the only branch where we actually need it when there is an issue we need to report.

In general we optimize for the case where no errors occur.

vadl-frontend/main/vadl/ast/TypeChecker.java
```diff
@@ -3629,17 +3654,18 @@ private void visitSliceIndexCall(CallIndexExpr expr, Type typeBeforeSlice,
...
     for (var slice : sliceGroups) {
+
+      if (!(currType instanceof BitsType) && !(currType instanceof TensorType)
+          && !(currType instanceof ArrayType)) {
+        var loc = expr.target.location().join(targetLoc.location());
```
flofriday: **minor:** The `.location()` call is redundant on `targetLoc.location()` as it's already a SourceLocation.

vadl-frontend/main/vadl/ast/TypeChecker.java
```diff
@@ -3903,8 +3966,28 @@ private void visitSubCall(CallIndexExpr expr, Type typeBeforeSubCall) {
...
+
+        expr.type = switch (fieldName) {
+          case "length" -> arrayType.lengthType();
+          case "bitLength" -> arrayType.bitLengthType();
+          default ->
+              throw new IllegalArgumentException("Unknown array field: `%s`".formatted(fieldName));
+        };
       } else {
-        addErrorAndStopChecking(error("Cannot resolve `%s`".formatted(fieldName), expr)
+        throw addErrorAndStopChecking(error("Cannot resolve `%s`".formatted(fieldName), expr)
```
flofriday, line 4270: **question:** Is there a reason you added the throw?

flofriday (commented):
Looks pretty good.
I obviously don't get why there is a need for the ArrayType 😅
Or how the groups work and I also couldn't find a lot in our documentation.

Maybe you can explain it in a comment and then disregard all the other comments.

vadl/main/vadl/ast/GroupDefUtils.java
rascmatt: done

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidGroupReference.vadl
rascmatt: I don't think so. It's quite Group-specific, and this was merely an attempt to introduce a more generic type, which may be reused for other structures in the future.

However, I now renamed it, since it's quite a specific use case anyways. Also the type is reused in the VIAM as type for a group reference.

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidGroupElemField.vadl
rascmatt: ty :)

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidGroupArrayMultiIndex.vadl
rascmatt: The actual type here is a `PseudoFormatType`, but I agree that this is not very informative. It now displays the type name, which would be `(AType ∩ BType)`.

vadl-frontend/main/vadl/ast/TypeChecker.java
rascmatt: Because I think it improves readability, given that the method always throws (so never completes normally), and also returns the error.

Also, IntelliJ gives a warning if you don't throw: `Result of 'addErrorAndStopChecking()' not thrown`, and imho rightly so.

vadl/main/vadl/types/Type.java
rascmatt: Yes it will also be used in the VIAM as type of a `GroupRef`

vadl/main/vadl/types/GroupType.java
rascmatt: Yes, `.` is the sequencing operator in the regular expression.

vadl/main/vadl/types/ArrayType.java
rascmatt: I would say the big difference is that a `TensorType` is a `DataType`, i.e. we know the dimensions statically. Array type (now renamed to `GroupType`) depends on the group expression, which may be unbounded (in theory, in practice it will be bounded by a `stop` annotation).

vadl-frontend/main/vadl/ast/GroupDefUtils.java
rascmatt: done

Jozott00 (approved):

PR #977, opened 2026-05-22
iss: Emit simple vector instructions as gvec TCG operations

Jozott00:
This PR improves the performance of the generated ISS by emitting gvec operations for certain `forall` constructs that were proved to be suitable for gvec lowering during vector analysis.

Additionally, it refactors some approaches introduced in #957. Previously an gvec suitable instructions were emitted in a separate emission path than TCG scalar and helper only instructions. Now, the analyzed forall loops are lowered to a new `TcgGvecOpNode` and handled by the normal TCG lowering and emission path.

In an follow-up PR, I will also refactor the vector analysis/planning to be applied on instruction behavior regions instead of the instruction as a whole. This should allow us to recognized multiple gvec operations as well as scalar and vector operation mixes.


BadGraphixD (approved):
Didn't read everything exactly but looks good to me.

PR #980, opened 2026-05-24
frontend/viam: Support partial (alias) register access

Jozott00:
This PR moves the partial (alias) register access expansion to the VIAM.
This means that the frontend generates normal (alias) register access nodes, even for accesses where not all indices are provided. 

In the common VIAM pipeline, two passes perform the expansion logic on the register accesses, so the VIAM stays the same for all backends.

This is crucial for the ISS, as it will later bypass the expansion passes and handle partial vector accesses independently, resulting in improved performance.

### DIFF
--- a/vadl/main/vadl/ast/BehaviorLowering.java
+++ b/vadl/main/vadl/ast/BehaviorLowering.java
@@ -68,7 +68,6 @@ import vadl.viam.Logic;
 import vadl.viam.Memory;
 import vadl.viam.Operation;
 import vadl.viam.Procedure;
-import vadl.viam.RegisterResource;
 import vadl.viam.RegisterTensor;
 import vadl.viam.Resource;
 import vadl.viam.StageOutput;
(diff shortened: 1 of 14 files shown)


vadl/main/vadl/ast/BehaviorLowering.java
```diff
@@ -788,197 +786,30 @@ private SubgraphContext fetch(Statement stmt) {
...
-      }
-    }
-
-    return requireNonNull(result);
+  private ExpressionNode readRegisterTensorDirect(RegisterTensor resource,
+                                                  List<ExpressionNode> indices,
+                                                  DataType type) {
+    // Preserve aggregate base-register accesses when fewer than all register-tensor indices are
+    // supplied. The VIAM and ISS node contracts already support these partial reads directly.
+    return new ReadRegTensorNode(resource, new NodeList<>(indices), type, null);
```
flofriday, line 794: **praise:** Awesome ❤️

vadl/main/vadl/viam/ArtificialResource.java
```diff
@@ -231,7 +231,15 @@ public DataType resultType() {
 
   @Override
   public DataType resultType(int providedDimensions) {
-    return resultType();
+    ensure(providedDimensions <= maxNumberOfAccessIndices(),
+        "Too many dimensions provided, max is %s, got %s.", maxNumberOfAccessIndices(),
```
flofriday, line 235: **question:** Will this now only be checked when the backend runs?

flofriday (approved):
LGTM

vadl/main/vadl/viam/ArtificialResource.java
Jozott00: Mostly. It will be checked when calling `resultType`, which is probably mostly called by backends. 

Note that if this check fails, this is emitted as an OpenVADL crash/internal error, not a diagnostic.

Issue #979, opened 2026-05-24
frontend/lsp: endless loop / out of memory exception with an faulty specification (modified huge.vadl)


AndreasKrall:
With a modified wrong variant of sys/huge/huge.vadl both the cli version of VADL and the lsp go into an endless loop resulting in a crash of the cli version with an out of memory error and the lsp to using 100% of a processor. To create this behavior it is just necessary to delete the first format definition. In the lsp I can go to 1800% cpu usage by cutting and inserting this format definition 18 times (I have 18 cores) resulting in heating up the processor and activating the fan.

Delete this complete format definition (once to crash the cli version, several times for the lsp to capture cpus)
```
  format ArithmeticRegFormat : Instr =  // arithmetic register +-&| shifted register format
    { opsize   : Bits<2>                // operation size
    , cond     : Bits<3>                // condition
    , ccsz     : Bits<2>                // condition size
    , cregl    : IndexC                 // condition register left
    , cregr    : IndexC                 // condition register right
    , op       : Bits<7>                // opcode
    , regd     : Index                  // destination  register
    , regl     : Index                  // left  source register
    , regr     : Index                  // right source register
    , regs     : Index                  // shift source register
    , opShift  : Bits<2>                // operation with shifted register
    , funct    : Bits<14>               // function code
    , shift    : Shift                  // shift amount
    }
```

flofriday:
That's really odd, haven't had a fault like this before. 🤔

flofriday:
Okay, I got it, it all seems to be from the error handling code.
OpenVADL tries to be really helpful to the programmer and so if it cannot find a variable name (or format like here) it tries to find all similar names. 
In huge specs like this that takes just waaaay to long so we need to set some upper bounds here.

To be fair this isn't an OOM issue because that only happens on the no-gc build of the compiler.

AndreasKrall:
I just noticed that it is related to generate error messages, but this will be around some million error messages. The same happens if you just change on letter of the name of the format used in the instruction definition:

```
  model ArithmeticRegInstr (size: Id, sizeStr: Str, c : CondRec, i : InstrRec, o : OpRec) : IsaDefs = {
    instruction AsId ($i.id, $size, $o.opId, $c.condStr) : ArithmeticRegFormat =
      if ($c.condEx) then
        AsId (R, $size)(regd) := VADL::$i.op (AsId (R, $size)(regl), $o.opEx) as AsId(Bits, $size)
[...]

PR #981, opened 2026-05-25
Frontend: Improve runtime errors huge spec

flofriday:
Related: #979

### DIFF
--- a/vadl/main/vadl/ast/SymbolTable.java
+++ b/vadl/main/vadl/ast/SymbolTable.java
@@ -39,6 +39,9 @@ import vadl.utils.SourceLocation;
 import vadl.utils.WithLocation;
 
 class SymbolTable {
+
+  static private final int MAX_SUGGESTIONS_SYMBOL_NAMES = 100;
+
   @Nullable
   SymbolTable parent;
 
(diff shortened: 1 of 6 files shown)


vadl/main/vadl/ast/SymbolTable.java
```diff
@@ -39,6 +39,9 @@
 import vadl.utils.WithLocation;
 
 class SymbolTable {
+
+  static private final int MAX_SUGGESTIONS_SYMBOL_NAMES = 100;
```
Jozott00: This is the maximum number of collected symbol names for suggestions right? Maybe we could add "COLLECTED" to the name, so it is clear that this is about symbol collection instead of symbols suggested to the user.

Jozott00 (approved):
LGTM

AndreasKrall (approved):
Looks good to me. The most important fix is the restriction to `MAX_PRINTED_BACKTRACES`.

Jozott00 (approved):

AndreasKrall (changes requested):
I believe it was not intentional that you want to push a wrong huge.vadl in sys/huge. Please revert the deletion of the format ArithmeticRegFormat so we have a working huge.vadl

AndreasKrall (approved):
LGTM

flofriday:
There was a OOM error, because the new huge spec allocated to much in the tests (which defaulted to 512MB). I adjusted the limit and everything should work now.

AndreasKrall (approved):
LGTM

PR #984, opened 2026-05-25
chore: Improve performance of decoder generation

rascmatt:
Implemented some improvements to the decoder generation. Some of them were long overdue, in particular the refactor of the representation of bit patterns using `BigInteger`.

* Refactor the `BitPattern` and `BitVector` utility implementations to use `BigInteger` as internal state, and implement utility methods using bitwise operations where possible.
* Refactor constraint synthesis pass to use an inverted index for lookup, rather than nested loops
* Applied a pre-filter during the split-pattern generation, to avoid large nested loops when filtering for matching instructions for those patterns later.
* Refactor the utility methods used in input preparation, to avoid bitwise changes to `BigInteger`, which result in unecessary allocations and loop iterations

edit: I removed the fast path for `PatternUtils#compatible` again, which did not result in a measurable performance difference anymore.

---

The timings resulted from the following command, with three iterations per version, on an Apple M5 with 32GB RAM:

```bash
openvadl-native check sys/huge/huge.vadl --timings --decoder skip=encoding-verification --decoder skip=decoder-verification
```

VDT Constraint Synthesis went from ~14.6s to ~60ms, and VDT Lowering went from ~67.1s to ~137ms. Combined relevant VDT time dropped from ~81.7s to <1s.

---

Interestingly, it seems like the timings are not necessarily consistent across compilations (`./gradlew nativeCompile`). After trying out some changes and then later restoring the branch, I was not able to reproduce the timings from before exactly, by about ~0.8s. Even though the variance between iterations was very low.


AndreasKrall (approved):
I didn't understand everything, but it looks good to me.

PR #985, opened 2026-05-25
rtl: Iterative synthesis finding max. clock frequency

linushdot:
Iterative Synthesis with OpenROAD for benchmarks

CSV columns now: `spec,tag,chip area,data arrival time`

New CSV columns: `spec,tag,decode area,chip area,clock period,timing delta`

Decode area is taken from Yosys output (pre-placement)


rascmatt (approved):
LGTM

PR #986, opened 2026-05-26
iss: Implement VADL::rrx, add VADL::rrxs

BadGraphixD:
- Changed VADL::rrx spec
  - Removed second operand, it always rotates by 1
  - Added VADL::rrxs
- Implemented VADL::rrx for ISS
  - Since there is no tcg equivalent, it is converted in IssNormalizationPass (similar to how upstream does it)
- Changed aarch32 assembly to print `rrx` instead of `rrx #1` for rotating input operands

### DIFF
--- a/vadl/main/vadl/types/BuiltInTable.java
+++ b/vadl/main/vadl/types/BuiltInTable.java
@@ -945,6 +945,8 @@ public class BuiltInTable {
   public static final BuiltIn RRX =
       func("VADL::rrx",
           Type.relation(List.of(BitsType.class, BoolType.class), BitsType.class))
+          .compute((Constant.Value a, Constant.Value c) -> a.rrx(c.bool())
+              .get(BUILTIN_RESULT, Constant.Value.class))
           .takesDefault()
           .returnsFirstBitWidth(BitsType.class)
           .build();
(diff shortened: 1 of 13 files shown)


AndreasKrall (approved):
Looks good to me, but there are failing checks.

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -940,17 +940,27 @@ public class BuiltInTable {
...
+   * {@code function rrxs( a : Bits<N>, c : Bool ) -> ( Bits<N>, Status ) }
+   */
+  public static final BuiltIn RRXS =
+      func("VADL::rrxs",
+          Type.relation(List.of(BitsType.class, BoolType.class), BitsType.class))
+          .takesDefault()
+          .returnsFirstBitWidthAndStatus(BitsType.class)
+          .build();
+
+
```
Jozott00, line 966: Would be nice if you could add a `compute` function (implementation in `Constant.Value`), so the built-in can be constant evaluated.

If you do it, please also add constant evaluation tests for the built-in.

Jozott00 (approved):
LGTM

Jozott00 (approved):

PR #988, opened 2026-05-27
frontend: Add named forall fold actions

flofriday:
Previously only operators worked.


AndreasKrall (commented):
Looks fine to me, but it is necessary to add test cases which check the new functions s/umin, s/umax.

Regarding the FIXME I do not think we need further fold operators beside subtract. Subtract is a special case as it is not commutative, but if we define the fold as 0 - elements then it is also well defined.

flofriday:
Yes, I know. But that's why this PR is still a draft and I didn't assign any reviewers. 😅

Jozott00 (approved):
LGTM

PR #990, opened 2026-05-27
frontend: Further improve performance on many name errors

flofriday:
Further improve the performance of the error messages, if there are reported a lot.

Related: #979


AndreasKrall (approved):
Looks good to me, but I believe a value of 1000 for `MAX_DIAGNOSTICS_WITH_NAME_SUGGESTIONS` is still to high.

PR #993, opened 2026-05-28
iss: Support gvec lowering for `forall tensor` assignments

Jozott00:
This PR enables gvec lowering for forall tensor assignments by skipping the (partial) register access expansion passes.


BadGraphixD (approved):
Looks good to me

PR #994, opened 2026-05-28
frontend: Allow integer, boolean and binary literals in AsId and AsStr expressions

### LINKED ISSUE #992: language: Allow Integers in AsId and AsStr
Currently the following code isn't allowed:

```
constant AsId("C", 1) = 1
```

Obviously this example can be solved by changing the one also to integer literal but when writing models this means we have to pass two parameters to it as there is no way to convert integers to strings.

flofriday:
Closes: #992


AndreasKrall (approved):
Looks fine to me.

PR #996, opened 2026-05-28
cli: Add new version format

### LINKED ISSUE #676: cli: Add version flag
It often could make sense to discover the current version of the openvadl compiler.

For example editor plugins could implement different behaivor based on the version.

My preferred output would be something like:
```
openvadl --version
v0.0.3 commit 5eb1ac837 build at Jan 14 2025 14:13
```

But probably just the version number is good enough.

flofriday:
Closes: #676

### DIFF
--- a/vadl-cli/main/vadl/cli/Helpers.java
+++ b/vadl-cli/main/vadl/cli/Helpers.java
@@ -35,7 +35,12 @@ class VersionProvider implements CommandLine.IVersionProvider {
 
   @Override
   public String[] getVersion() {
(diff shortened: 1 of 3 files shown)


Jozott00 (commented):
Please checkout the root `build.gradle.kts` where the current versioning logic lives. We are already using a git versioning plugin, so we could probably just tweak it a bit.

flofriday:
Yes, we can!
It was much easier than my original implementation, thanks 🤗

Jozott00 (approved):
LGTM

PR #1000, opened 2026-05-29
iss: Refactor vector analysis

Jozott00:
Instead of using per-instruction classification, we now operate on regions within instructions. This allows us to analyze and classify forall loops in the same independently of each other.

Note that most of the added lines are snapshot tests that allow us to quickly spot differences in instruction analysis, so notice regression before benchmarking.


BadGraphixD (approved):
Looks good.

PR #1001, opened 2026-05-30
frontend: Fix errors in currently visiting context

flofriday:
There was an issue in the implmeentation with the currently visiting stack.
I've also added comments explaining the methods and why it's so easy to misuse.


flofriday:
@rascmatt I've updated the PR with the stack fix and the new example.

rascmatt (approved):
Nice, looks good to me!

rascmatt (approved):

PR #1004, opened 2026-05-30
chore: Add memory big-/bi-endian tests

BadGraphixD:
- Adds frontend snapshot tests and tests for `IssApplyMemoryEndiannessPassTest`
- Bugfix in `IssApplyMemoryEndiannessPass` concerning endianness of memory writes in memory region init procedure
  - before: writes were lowered to `IssStaticEndianConditionNode`
  - now: endian condition is constant-evaluated


Jozott00 (approved):
LGTM!

PR #1005, opened 2026-05-31
chore: Add execution-state register tests

BadGraphixD:
Adds frontend snapshot tests for the `[ execution state ]` annotation, tests for `IssRegisterAccessLoweringPass` and snapshot tests for the generated c-code responsible for packing/extracting tb-state registers.

What remains to be tested:
- Translation block ending and chaning when tb-state writes are involved (though that should probably wait until it is rewritten, see #865)


BadGraphixD:
@Jozott00 I tried adding some form of snapshot tests for the generated c code (`translate.c` and `cpu.h`). But the output seems to be too inconsistent (especially the clang code formatting behaves differently on my machine and during github actions). My solution now is to simply test if certain lines are present. Do you already have a solution for testing generated c code like that or should I avoid writing tests like that?

Jozott00:
Until now I avoided snapshot tests, because the ISS behavior output changed a lot. However, in the meantime the ISS became quite stable, so adding snapshot tests would make sense now.

Also I think having snapshot tests like the frontend is the better approach than checking for certain lines. Regarding the clang formatting: Maybe we should just add a config flag to the `IssConfiguration` to turn off formatting. What do you think?

Jozott00 (approved):
LGTM

BadGraphixD:
@Jozott00 I think disabling the clang formatting for the tests will work.

Jozott00 (approved):
LGTM

PR #1009, opened 2026-06-04
iss: Optimize IssSafeResourceReadPass

BadGraphixD:
This pass previously recursively traversed the data dependency graph of some nodes to find usages. In certain graphs this lead to huge amounts of duplicates, because some dependency nodes were visited multiple times. Now, the usages are cached per node during the traversal.


Jozott00 (approved):
Thanks! LGTM

PR #1011, opened 2026-06-05
asm: Strengthen immediate checking in assembly parser

### LINKED ISSUE #982: asmparser: Decode of immediate operands truncates values, which makes predicates always true
To check if immediate operands are within the valid range, the asm parser uses the predicates generated by the LCB. Predicates generally operate on the 'field access view' of values, but the asm grammar allows specifying operands (as attributes) with both the field name or the field access name.

Therefore, to check if an operand specified with its field name (=parsed value is in 'field view') is in its valid range, the value needs to be transformed into its 'field access view' by calling the corresponding `decode` function. Only then can the predicate be called to check if the value is within its range.

But there is an issue with some generated `decode` functions, where the value is truncated, which makes the call to the predicate always true. An example of this is the `BAL` instruction of the VarRisc spec.

This is the asm grammar specification of `BAL`:
```java
BalInstruction @instruction:
  mnemonic = "BAL" @operand
  rd = Register @operand ","
  imm = ImmediateOperand      // immediate specified with field name 'imm'
;
```

In the asm parser the immediate value is checked by first calling `decode` and then the predicate:
```cpp
opImm64 = VarRISC32_BAL_offsetS_decode(opImm64);
            
if(!VarRISC32_TYPE_C_offsetS_predicate(opImm64)) {
  std::string error = "Invalid immediate operand for imm. The predicate does not hold.";
  Parser.Error(Op.getStartLoc(), error);
  return true;
}
```
where this is the used `decode` function, which truncates the input value:
```cpp
static int32_t VarRISC32_BAL_offsetS_decode(uint16_t imm) {
   return VADL_lsl(VADL_sextract(imm, 16), 32, ((uint8_t) 0x1 ), 1);
}
```

This truncation leads to the following incorrect behavior of the assembler:
```
bal r1, 0x7fff // works with 32767
bal r1, 0x8000 // works with -32768
bal r1, 0xffff // works with -1
bal r1, 0x10000 // should be invalid imm, but truncated to 0
bal r1, 0x10001 // should be invalid imm, but truncated to 1
bal r1, 0x40001 // should be invalid imm, but truncated to 1
bal r1, 0xfffff // should be invalid imm, but truncated to -1
```

Is the generated `decode` function correct? Is it supposed to truncate the value?
If this `decode` is correct, then I would suggest to generate special predicates which operate on the 'field view'.
If this `decode` is incorrect, obviously decode function generation in the LCB needs to be fixed.

(This bug also affects semantic rule merging in rule generation, since there predicates are used to guide the parser.)

benjaminkasper99:
This PR fixes #982 by adding a check for the bitwidth of field operands, before `decode` functions are called. This ensures that passed values meet `decode` functions expectation regarding bit width.

Further included in this PR:
- Checks for value ranges of immediate operands of pseudo instructions
- Fixed location reporting regarding immediate errors in the parser
- Slightly refactored the `EmitAsmParserCppFilePass`

### DIFF
--- a/sys/risc-v/rv3264im.vadl
+++ b/sys/risc-v/rv3264im.vadl
@@ -443,10 +443,10 @@ $Arch3264 ( // ((((((((((((((((((((((((((((((((((((((((
   }
   assembly LLA = (mnemonic, " ", register(rd), ",", hex( symbol ))
 
-  pseudo instruction LI( rd : Index, symbol : Bits<32> ) =
+  pseudo instruction LI( rd : Index, symbol : SIntW ) =
   {
-      LUI{ rd = rd, imm = hi( symbol ) }
-      ADDI{ rd = rd, rs1 = rd, imm = lo( symbol ) }
+      LUI{ rd = rd, imm = hi( symbol as Bits<32> ) }
+      ADDI{ rd = rd, rs1 = rd, imm = lo( symbol as Bits<32> ) }
   }
   assembly LI = (mnemonic, " ", register( rd ), ",", decimal( symbol ))
 
--- a/sys/v-risc/ABI.vadl
+++ b/sys/v-risc/ABI.vadl
@@ -196,6 +196,12 @@ assembly description AD for ABI = {
             imm = ImmediateOperand
         ;
 
+        BalLInstruction @instruction:
+            mnemonic = "BAL_L" @operand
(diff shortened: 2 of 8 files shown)


Tbswss (approved):
I do not have any comments, looks fine to me! Master needs to be merged / rebased into the branch.

Issue #995, opened 2026-05-28
lsp: Restrict useless concurrency


flofriday:
In case the frontend runs forever, the LSP should only be able to pin one core to 100% instead of all cores.

Reasoning: https://github.com/OpenVADL/openvadl/issues/979#issuecomment-4529612739

jaraonthe-dot-net:
We may want to re-use an AST as long as the underlying files aren't modified, to improve responsiveness in the UI.
I.e. one thread may use/wait for an AST that has been/is being computed in another thread (e.g. publishDiagnostics has no errors -> use this AST for Token Hover (#1012)). This should be taken into account here or tackled in a follow-up ticket.

PR #1012, opened 2026-06-05
lsp: Type Hover

jaraonthe-dot-net:
Closes #841


jaraonthe-dot-net:
#841 using
#841 using (Monokai)
#841 register

Implementation notes:
- Attempting to produce Hover for any `TypeeNode` entity (see Screenshots: Hover for entiry `register` definition, but this wouldn't work on `using` definition). Is this okay, or should we constrain this hover to `Expr` nodes?
- LSP clients may apply TextMate grammar to the hover text (due to it being markup with a vadl code block). As you can see in the screenshot with the other theme, it is not exactly as expected, as the TextMate grammar doesn't expect snippets like this (here: the angle brackets aren't colored in red). We can accept this, or adapt the TextMate grammar, or not have the hover text highlighted at all.
- I had to do a small modification of the AST datatypes, as otherwise the class hierarchy wouldn't work in `AstFinderByPosition`.
- I've noted that, after the `TypeChecker`, some nodes don't have a type set after all. E.g. in my example code, the `baseType` of "Reg". I had to work around this in `AstFinderByPosition`. See screenshot:

image

vadl-lsp/test/vadl/lsp/TestSnapshot.java
```diff
@@ -138,7 +139,7 @@ public TestSnapshot(String testCase, boolean shortenFilePath) {
   public void add(String name, Object value) {
     data.append("- ").append(name).append(": ");
 
-    var valueString = value.toString();
+    var valueString = Objects.toString(value);
```
flofriday, line 142: **question:** What's the reason for this change? The `Objects.toString)` should only be necessary if the value is nullable but in the signature it looks like it's never null.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -206,6 +209,65 @@ private Either<List<? extends Location>, List<? extends LocationLink>> definitio
...
+  @Override
+  public CompletableFuture<Hover> hover(HoverParams params) {
+    log.debug(">> hover: {}", params);
+
+    LspSnapshotFileSystem snapshots = createSnapshotFileSystem();
+    return CompletableFuture.supplyAsync(() -> {
+      Document document = getDocumentForParams(params, snapshots, "hover");
+      Ast ast;
+      try {
+        // Note: Would like to avoid applying ModelRemover, but the TypeChecker depends on that
```
flofriday: **question:** Why would you like to drop the ModelRemover?

Because even if we persist models, we probably couldn't typecheck them because the types depend on the scope in which the model is invoked.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -206,6 +209,65 @@ private Either<List<? extends Location>, List<? extends LocationLink>> definitio
...
+        // step.
+        ast = Frontend.compileToAst(toPath(document.uri), snapshots);
```
flofriday, line 220: **question:** Does this mean we recompile the spec on every hover event?

Does that run reasonable fast on large specs like aarch32 or huge.vadl?

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
```diff
@@ -59,7 +46,7 @@ private AstFinderByPosition(Path path, SourceLocation.Position position) {
   //        identifier nor a target
 
   /**
-   * Finds an Identifier or IdentifierPath at the given source code position, and returns it's
+   * Finds an Identifier or IdentifierPath at the given source code position, and returns its
    * target's location.
```
flofriday, line 49: **chore:** If we are already adjusting the docs than let's say it returns the identifier and not the target location.

vadl-lsp/test/vadl/lsp/integration/HoverTest.java
```diff
@@ -0,0 +1,83 @@
...
+    var position = new Position(line - 1, character - 1);
+    snapshot.add("Input with position marked", TestUtils.showPositionInFile(position, input));
+
+    // Doing this to avoid race condition in the TypeChecker with publishing diagnostics:
+    // Type.bitsTypes may be modified by both at the same time (ConcurrentModificationException).
+    // Unlikely for this to occur in normal operation, but this test would be flaky otherwise.
+    client.doThenWaitForPublishDiagnostics(() -> openDocument(inputUri, input));
+    var result = textService.hover(new HoverParams(
+        new TextDocumentIdentifier(inputUri), position
+    )).get();
```
flofriday, line 69: **question:** First, is that because we run the typechecker once when opening a document and once when hovering?

And second if this is an issue, doesn't that also mean that during normal handling we could also get a racecondition and ConcurrentModificationException in the LSP, if I hover over something shortly after I opened the file?

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
```diff
@@ -87,27 +74,171 @@ private AstFinderByPosition(Path path, SourceLocation.Position position) {
...
+  /**
+   * Finds a TypedNode at the given source code position.
+   *
+   * @param ast Should have gone through the TypeChecker
+   * @param path The source code file to search in
+   * @param position The position to search for (within the file identified by {@code path})
+   * @return Null if no TypedNode found at {@code position}
+   */
+  static @Nullable TypedNode findTypedNode(Ast ast, Path path, SourceLocation.Position position) {
+    var visitor = new AstFinderByPosition.AfterTravel<TypedNode>(ast, path, position, (n) -> {
```
flofriday: **suggestion:** By changing the these two lines to the following you no longer need to modify the `TypedNode` and is how I intended TypedNode to be used.

```java
  static @Nullable <T extends Node & TypedNode> T findTypedNode(Ast ast, Path path, SourceLocation.Position position) {
    var visitor = new AstFinderByPosition.AfterTravel<T>(ast, path, position, (n) -> {
 ```

To be fair I don't _love_ how we use TypedNode and I'd much rather have a language with multiple inheritance but in Java I think this is the best tradeof.

File: vadl/main/vadl/ast/Ast.java [hunk 7]
```diff
@@ -248,7 +248,7 @@ interface IdentifiableNode {
   Identifier identifier();
 }
 
-interface TypedNode {
+interface TypedNode extends WithLocation {
```
flofriday: **nitpick:** I don't like this change because in general TypedNode should only be used in tandem on elements that are already a `Node` so it should always be possible to cast the value to a `Node` which already implements WithLocation.

If you apply the suggestion in AstFinderByPosition class than this change is also no longer needed.

vadl/main/vadl/ast/Expr.java
```diff
@@ -1971,7 +1971,10 @@ public int hashCode() {
       return Objects.hash(values);
     }
 
-
+    @Override
+    public SourceLocation location() {
+      return location;
+    }
```
flofriday: **chore:** If we revert the TypedNode modification then this is also no longer needed.

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
```diff
@@ -29,23 +32,7 @@
...
-    public FoundSignal(IsId identifier) {
-      this.identifier = identifier;
-    }
-  }
-
-  private AstFinderByPosition(Path path, SourceLocation.Position position) {
-    searchPath = path;
-    searchPosition = position;
-  }
+public abstract class AstFinderByPosition<N extends WithLocation> extends RecursiveAstVisitor {
```
flofriday: **nitpick:** Is there any good reason not to replace this with `<N extends Node>` since it should always return a Node right?

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
```diff
@@ -87,27 +74,171 @@ private AstFinderByPosition(Path path, SourceLocation.Position position) {
    * @return Null if no Identifier or IdentifierPath found at {@code position}
    */
   static @Nullable IsId findIdentifier(Ast ast, Path path, SourceLocation.Position position) {
-    var visitor = new AstFinderByPosition(path, position);
+    var visitor = new AstFinderByPosition.BeforeTravel<IsId>(ast, path, position,
+        (n) -> n instanceof IdentifierPath || n instanceof Identifier);
+    return visitor.find();
```
flofriday: **praise:** This looks very clean ✨

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
```diff
@@ -87,27 +74,171 @@ private AstFinderByPosition(Path path, SourceLocation.Position position) {
...
+    if (typedNode == null) {
+      return null;
+    }
+    return new TypeAndLocation(typedNode.type(), typedNode.location());
+  }
+
+  /**
+   * Return value for {@link #findTypedNodeType(Ast, Path, SourceLocation.Position)}.
+   */
+  public record TypeAndLocation(Type type, SourceLocation location) {}
```
flofriday, line 103: **suggestion:** With the change below such that findTypedNode returns a `T extends Node & TypedNode` the caller would receive a Node on which they can call `type()` and `location` directly kinda making this method record pair redundant.

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
```diff
@@ -87,27 +74,171 @@ private AstFinderByPosition(Path path, SourceLocation.Position position) {
...
+
+  /**
+   * If this yields true in {@code beforeTravel()} or {@code afterTravel()}, the current node is a
+   * candidate for what we are looking for. Must NOT yield true if {@code testedObject instanceof N}
+   * is false!
+   */
+  protected final Predicate<WithLocation> condition;
+
+  @Nullable
+  private N foundNode = null;
```
flofriday, line 144: **very nitpicky:** But I think I like fields before static methods, but that's definitly taste so disregard it if you disagree.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -206,6 +209,65 @@ private Either<List<? extends Location>, List<? extends LocationLink>> definitio
...
+    Hover result = null;
+    if (text != null) {
+      // For now, we assume that all hover texts are snippets of valid OpenVADL source code, so
+      // let's use Markdown to mark them as such
+      if (clientSupportsMarkdownInHoverMarkupContent()) {
+        result = new Hover(new MarkupContent(MarkupKind.MARKDOWN,
+            "```" + LANGUAGE_IDENTIFIER + "\n" + text + "\n```"));
+      } else {
+        // Fallback to deprecated MarkedString
+        result = new Hover(Either.forRight(new MarkedString(LANGUAGE_IDENTIFIER, text)));
```
flofriday: **minor:** `MarkedString` is deprecated since 2017, if you use an editor this old, your're on your own.

I think the better fallback would be to use `MarkupContent(MarkupKind.PLAINTEXT` which would be plaintext.

This also let's us get rid of the deprecation suppression.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -206,6 +209,65 @@ private Either<List<? extends Location>, List<? extends LocationLink>> definitio
...
+      if (info == null) {
+        return hoverResult(null, null);
+      }
+
+      return hoverResult(info.type().name(), document.calculateUtf16Range(info.location()));
+    });
+  }
+
+  @SuppressWarnings("deprecation")
+  private @Nullable Hover hoverResult(@Nullable String text, @Nullable Range range) {
```
flofriday, line 238: **suggestion:** I would remove all `Nullable` from this signature because if the inputs are null
then the outputs are null.

And so we can simply skip calling this method all together 
`return hoverResult(null, null);` -> `return null;`

flofriday (commented):
That was really fast, thank you. 🚀

Also well tested, I'm sure users will love it.
There are just a couple of notes and questions.

flofriday:
I saw you also had some questions.

> [...] Hover for entiry register definition, but this wouldn't work on using definition). Is this okay, or should we constrain this hover to Expr nodes?

I think this is actually quite good. Eventually we do
- Only type hover content for Expr and Statement nodes
- Some kind of signature preview for Definition Nodes, issue: #1015

> LSP clients may apply TextMate grammar to the hover text

I'm split on this. I think for the signature preview we will do later, this will be helpful but for the simple type, it isn't quite correct because loose type literals are not valid VADL code and also highlighting them doesn't provide much help.
However, it also doesn't distract that much so we can do that later.

> I've noted that, after the TypeChecker, some nodes don't have a type set after all. E.g. in my example code, the baseType of "Reg".

Yes that is true. In the frontend we use Identifiers sometimes as expressions `a + b` in which case they should have received a type. In other cases when they are used as names like in typeliterals `Bits<32>`or in definitions `function hello()` they don't get a type.

That might be confusing but every time when I tried adding another identifierWithoutType the code ended up worse than before.

vadl-lsp/test/vadl/lsp/TestSnapshot.java
jaraonthe-dot-net: One of the new test cases records a `null`. I'll add the annotation.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
jaraonthe-dot-net: The ModelRemover makes it impossible to provide Type Hover within a model definition. Maybe providing type information within a model is not possible anyway, but then again what if we want to provide some other information in the future (#1015)?

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
jaraonthe-dot-net: It takes a few seconds for huge.vadl.
I think it makes sense to re-use a perfectly good AST if the file hasn't been modified. That involves coordination across different threads, so this should be looked into in #995.

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
jaraonthe-dot-net: The adjustment was just a grammar fix (it's -> its).
More to the point: This is the javadoc for `findIdentifierTargetLocation()`, which does indeed return a location. Are you referring to `findIdentifier()`? (But there the javadoc should also be correct.)

vadl-lsp/test/vadl/lsp/integration/HoverTest.java
jaraonthe-dot-net: Yes, opening the document triggers a `publishDiagnostics()`, which, just like Hover, does a full `Frontend.compileToAst()`.
Without this safeguard, a race condition ensued every couple of times of running this test.

In regular operation this shouldn't be an issue because:
- A human user isn't as fast as the test suite; it is unlikely for the timings to line up that perfectly (both compilation runs must be at the same point in the TypeChecker at the same time).
- AFAIK, the static `Type.bitsTypes` is an object cache that is built up over time - once a particular entry has been computed, it should be fine going forward (as it is a `ConcurrentModificationException`). There are only very short windows (in the lifetime of the language server process) in which this race condition can be triggered at all.
- Finally, even if it occurs, the language server returns an error to the client but should be fine going forward (i.e. it's just a small fluke in the UX).

So, a very unlikely event with rather mild outcome. I don't think it is worth replacing the `Type.bitsTypes` data structure with a concurrency-safe one.

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
jaraonthe-dot-net: Didn't know Java had this feature. Definitely gonna use it. Thank you.

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
jaraonthe-dot-net: I would usually agree, but this class should only be used through its static methods; that's why I put them at the top where they are easy to find.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
jaraonthe-dot-net: But this method also logs a debug message, which I would like to keep in one place (and I want to see in the log that the result of this particular hover request is null).

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
jaraonthe-dot-net: That is not possible because `Node` and `TypedNode` are not declared `public`, i.e. I cannot access them in the `vadl.lsp` namespace.

vadl-lsp/test/vadl/lsp/TestSnapshot.java
flofriday: How were you even able to compile with the annotation missing? Normally we the null checks abort the compilation task if they fail.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
flofriday: Well as far as I can tell the language (and with that the compiler) never treats the code in models as valid because it can never reason about the context they are used in.

If that holds it will always be impossible in models to do:
- typechecking
- symbol resolution (goto declaration, usages, signatures)

Let's discuss that again with Andi on Monday in the meeting but right now I cannot imagine how that should work.

**Update: ** I've created an issue with all my thoughts: #1016

vadl-lsp/test/vadl/lsp/TestSnapshot.java
jaraonthe-dot-net: It seems that nullability isn't checked in test sources. I experimented a bit with a random unit test in the vadl subproject, and it didn't complain here neither.

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
flofriday: you are right, I must have gotten confused by the diff UI

vadl-lsp/test/vadl/lsp/integration/HoverTest.java
flofriday: Even if it's not common, ignoring race conditions is quite hacky. 
In my testing I couldn't trigger it by hand so I'm fine going forwart with it for now but I've openend #1017  to tackle this.

vadl-lsp/main/vadl/ast/AstFinderByPosition.java
flofriday: God that is stupid.
Ok that's the final straw, I'll rewrite the AST package and make all the nodes public so that you don't have to handle two packages just to get access to the AST.
#1018

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
flofriday: I still think that copying the log message would be cleaner but if you prefer it this way that's ok.

vadl-lsp/test/vadl/lsp/TestSnapshot.java
flofriday: Ah fair, ok that's fine.

flofriday (approved):
LGTM

Issue #989, opened 2026-05-27
frontend: Improve levenshtein performance


flofriday:
I've just run into a problem where the performance isn't fast enough when deleting the first format of `huge.vadl`.

AndreasKrall:
It is not necessary to delete the first format, it is enough to change one letter of the format name which result in the same problem.

flofriday:
This is fixed with #1013

PR #1013, opened 2026-06-05
frontend: Improve symbol error performance

flofriday:
This commit fixes the hours of compile time with deleting the format in the expanded huge spec.

It does this by:
- Improving the levenshtein algorithm (found after profiling)
- Optimizing the diagnostic compression from O(n^2) to O(n).
- Limiting the amount of printed diagnostics to 10k

This means that the huge spec without the error now takes 6sec (unchanged) and with the errors it also only takes 6sec.

Related: #989


AndreasKrall (approved):
Looks good to me.

PR #1014, opened 2026-06-06
iss: Perform ISS liveness analysis in topological order

BadGraphixD:
Under the assumption that the CFG during TCG variable allocation is acyclic, we can optimize the liveness analysis by visiting each node only once and in topological order. This reduces the runtime from exponential to linear in the number of nodes if we assume everything else constant.

I have also kept the original algorithm and left some notes. As far as I can tell, the previous algorithm does not terminate on cyclic CFGs.


Jozott00 (approved):
LGTM

PR #1020, opened 2026-06-07
frontend: Change visibility of AST nodes to public

### LINKED ISSUE #1018: frontend: Make all AST nodes public
Currenlty the AST is package private which works great in a traditional compilation pipeline but since the LSP also needs acces to the nodes it has become quite annoying.

flofriday:
Closes: #1018

**For the reviewers:**
- @jaraonthe-dot-net  this code touches part of the LSP, can you verify that the new changes help the ergonomics in the LSP?
- @benjaminkasper99 can you look over the changes that affect your parts?
- @rascmatt I had to refactor parts of the recently added pseudoformattypes and add another FrontendBuiltin Table to avoid cyclic dependencies, can you review these parts?
- @Jozott00 can you review the general structure?


Jozott00:
Actually, maybe this would be a great opportunity to move the frontend in a separate module. The ast package is already very separate, so it should be easy to create a new module for it and move it there.

vadl-frontend/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+    testCompileOnly(libs.jsr305)
+    testAnnotationProcessor(project(":java-annotations"))
+    testImplementation(libs.buildkitcli)
+    testImplementation(libs.assertj.core)
+    testImplementation(libs.awaitility)
+    testImplementation(libs.testcontainers)
+    testImplementation(libs.archunit.junit5)
+    testImplementation(libs.velocity.engine.core)
+    testImplementation(libs.jqwik)
+    testImplementation(libs.snakeyaml)
```
Jozott00: minor: I think some of them are actually not required (like thymeleaf?)

Would be nice if you could check which one are actually needed by the module.

vadl-frontend/build.gradle.kts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+kotlin {
+    jvmToolchain(25)
+}
```
Jozott00: minor: I think the frontend doesn't has any Kotlin files no?

vadl/test-weird/dump/0_Single_Resource_Write_Validation_Pass.html
Jozott00, line 1: major: I think this should not be committed (although I don't know were it comes from)

vadl/test-weird/dump/1_Single_Resource_Write_Validation_Pass.html
Jozott00, line 1: major: Same here

Jozott00 (commented):
Nice, thank you! Just some notes

vadl-frontend/main/vadl/ast/PseudoFormatType.java
rascmatt, line 51: @flofriday 
I think the reason for the `PseudoFormatType` to be in the `ast` package in the first place, was due to the fact that the ast nodes (`InstructionDefinition` etc.) were not public.

So if that's no longer a restriction, I'd personally see/prefer it in the `types` package. This should also eliminate the separate handling in the BuiltIn table.

What do you think?

flofriday: sounds good to me, I'll try that :)

Jozott00: I think that won't work as it introduces a cyclic dependency between the `vadl` and the `vadl-frontend` module.

Only `vadl-frontend` may depend on `vadl` but not vice versa.

vadl-frontend/main/vadl/ast/nodes/AsmDescriptionDefinition.java
```diff
@@ -0,0 +1,148 @@
...
+package vadl.ast.nodes;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import vadl.javaannotations.ast.Child;
+import vadl.utils.SourceLocation;
+
+@SuppressWarnings({"MissingJavadocType", "MissingJavadocMethod"})
+public class AsmDescriptionDefinition extends Definition implements IdentifiableNode {
```
benjaminkasper99: The Asm... definitions all have a Javadoc comment, we can use them here instead of the `@SupressWarnings`.

vadl-frontend/main/vadl/ast/PseudoFormatType.java
flofriday: Yes exactly, and I couldn't think of a good way to quickly solve that :/

rascmatt: Hmm, true.

The refactor looks good to me, so I'm happy to keep it as is for this PR. 

To no longer depend on the AST nodes directly, the `PseudoFormatType` would have to either reuse the VIAM nodes, or introduce it's own data structures. Reusing the VIAM nodes (`Instruction`, `Operation`) doesn't really work, because the type is already used in the type checker before the VIAM lowering, and I don't particularly like to introduce yet another representation of Instructions, Operations, Formats etc. just for the PseudoFormatType ...

jaraonthe-dot-net:
Looks fine from LSP perspective.

vadl-frontend/main/vadl/ast/nodes/AsmDescriptionDefinition.java
flofriday: During the splitting some docstrings were lost, I recovered them again.

Jozott00 (approved):

benjaminkasper99 (approved):

PR #1027, opened 2026-06-20
iss: Optimize TB chaining and exiting

BadGraphixD:
Sometimes translation blocks must be ended (e.g. after jumps or tb-state changes). In some cases we can chain translation blocks (when the next tb-state and pc are always the same in the next block). The decision when to do that may depend on facts that are only known at translation time (e.g. instruction operands).

This PR defers the decision to translation time in certain cases.

Tests are included, but benchmarks need to be added to test for possible performance regressions.

Closes #865


vadl/main/vadl/iss/passes/tcg/lowering/TcgOpLoweringPass.java
```diff
@@ -443,32 +439,20 @@ private void addBeforeCurrent(TcgNode node) {
...
-      // we must do a TCG lookup and jump to the current PC
-      node.replaceAndLink(
-          new TcgLookupAndGotoPtr()
-      );
+    if (isDirectJmp(node)) {
+      // TODO: Currently we decide during translation time if a jump slot is free. We also
+      //  always use it as soon as possible.
+      // TODO: We could use heuristic to find the best slot assignment.
+      //  (other than just the first one)
+      node.replaceAndLink(new TcgIsJmpDirectJmp(node.cause().value(), optJumpSlot));
```
Jozott00, line 453: Lets create an issue for this (the heuristic based jump slot assignment)

Jozott00 (approved):
Really nice! And thanks for the nice tests!

vadl/main/vadl/iss/passes/tcg/lowering/TcgOpLoweringPass.java
BadGraphixD: see #1050

PR #1030, opened 2026-06-22
ci: Switch avoid artifact upload for frontend benchmarks

flofriday:
This PR also bumps some versions in the benchmark ci setup


Jozott00 (approved):
LGTM

PR #1032, opened 2026-06-22
typechecker: Add caches and general speed improvements

flofriday:
This PR includes:
- Caching explicit casting results
- Caching builtin checking results
- Skip generation of diagnostics that are never used

On my machine this speeds up all benchmarks, and for aarch32-expanded.vald it reduces the time **from 280ms to 262ms**.
On large specs this has a larger affect because they are more likely to keep checking the same types  and functions over and over again especially if they are generated with macros.

### Results
Above before, below after

Screenshot 2026-06-22 at 17 30 01

Screenshot 2026-06-22 at 17 30 04

### Note
~~I'll have to look into why the performance increased drastically for huge but not for huge-expanded. It kinda smells like GC cycle.~~

The reason why in the screenshots huge.vadl makes such a big jump is because of a logical error in my benchmarking setup, please ignore the huge benchmarks, all other specs are fine though.

### DIFF
--- a/vadl-frontend/main/vadl/ast/InternalErrorType.java
+++ b/vadl-frontend/main/vadl/ast/InternalErrorType.java
@@ -23,6 +23,11 @@ import vadl.types.Type;
  * expression. This differes from null, which means we haven't visited the expression yet.
  */
 public class InternalErrorType extends Type {
+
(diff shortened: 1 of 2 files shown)


Jozott00 (approved):
LGTM

PR #1036, opened 2026-06-24
frontend: Reduce memory footprint

flofriday:
This PR is an odd collection resulting from lots of profiling and eliminating memory heavy sections wherever possible.

This includes stuff like:
- Sizing ArrayLists and StringBuilder
- Avoiding Iterators for tiny lists on hot paths

As a result in my benchmarks huge.vadl now runs in **601ms down from 713ms**.


Jozott00 (approved):
I'm surprised that switching from iterators to array list access has any significant impact on the performance.

PR #1038, opened 2026-06-29
dump: Improve performance of VIAM dumps

flofriday:
This PR implements two optimizations for emitting the VIAM dumps:

## Enabling thymeleaf caches

Reenabling the caches, improves generation performance by an order of magnitude.
@Jozott00  do you remember why this was disabled in the first place.

**Before:**
	- HTML Dump at phase VIAM Creation:         6380.560ms
	- HTML Dump at phase VDT Creation:          5967.223ms

**After:**
	- HTML Dump at phase VIAM Creation:          966.983ms
	- HTML Dump at phase VDT Creation:           435.627ms
	
## Optimizing CollectBehaviorDotGraphPass.Result.map
Which first mapped over a dict to later evict all entries except for a special key.
Instead we are now just using the dictionary to get the values for a key.

With that dumps are now almost instant and even for huge you only have to wait for ~12s per dump, which previously didn't finish in any reasonable time (after 20m I gave up).

### DIFF
--- a/vadl/main/vadl/template/AbstractMultiTemplateRenderingPass.java
+++ b/vadl/main/vadl/template/AbstractMultiTemplateRenderingPass.java
@@ -18,6 +18,7 @@ package vadl.template;
 
 import static vadl.viam.ViamError.ensure;
 
+import java.io.BufferedWriter;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStreamReader;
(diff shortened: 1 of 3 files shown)


vadl/main/vadl/template/AbstractMultiTemplateRenderingPass.java
```diff
@@ -228,6 +229,7 @@ private void renderTemplate(Map<String, Object> vars,
     var actualWriter =
         enableCopyright() ? new PrependingWriter(writer, getCopyrightNotice()) : writer;
     templateEngine.process(getTemplatePath(), ctx, actualWriter);
+    System.out.println("renderTemplate " +  getTemplatePath());
```
Jozott00: minor: Should be removed

Jozott00 (changes requested):
Not sure why we disabled it.
Lets see if all tests pass when enabling caching.

Please change the PR title

Jozott00 (approved):
Thanks for optimizing it!

PR #1041, opened 2026-07-11
viam: Add InstructionWidthNode

### LINKED ISSUE #1003: language: PC manipulation in exceptions
We are currently in the progress of fixing some hardcoded values (8bit byte-size and 32bit instruction witdth) in the lowering of the program counters.

However, while I can get the current instruction with for almost all specs I don't know what to do with this code in ppc64
```
  // triggered by a system call instruction
  exception SysCall (lev : Bits<7>) = {
    SRR0 := PC.next
    SRR1 := (MSR(63..31), 0 as Nibble, MSR(26..22), srrlev(lev), 0 as Nibble, MSR(15..0))
    PC   := 0x0000'0000'0000'0C00
  }
```

To lower `PC.next` we need to figure out in which instruction we are in but we aren't in an instruction, we are in an exception.

BadGraphixD:
Add `InstructionWidthNode`, which is a new VIAM `ExpressionNode`, which represents the width of the current instruction in bytes. The necessity of this new node was discussed in #1003. It simplifies the frontend.

The node is replaced with a `ConstantNode` by the `PcOffsetPass`. Outside instruction behaviours, the instruction width is not known (e.g. in exceptions), so we will have to find custom solutions for some artefacts. For now, the pass replaces such occurrences with the width of the first instruction found in the ISA.

Closes #1003

### DIFF
--- a/vadl-frontend/main/vadl/ast/BehaviorLowering.java
+++ b/vadl-frontend/main/vadl/ast/BehaviorLowering.java
@@ -22,7 +22,7 @@ import static vadl.error.Diagnostic.ensure;
 import static vadl.error.Diagnostic.error;
 import static vadl.error.Diagnostic.warning;
 import static vadl.utils.GraphUtils.ifElseSideEffect;
-import static vadl.utils.GraphUtils.intS;
+import static vadl.utils.GraphUtils.intSNode;
 import static vadl.utils.GraphUtils.intU;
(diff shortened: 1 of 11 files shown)


Jozott00 (approved):
LGTM! 

Maybe create a separate issue for the missing dynamic instruction width support in exceptions.

PR #1042, opened 2026-07-19
 Implement operation element-of expression

rascmatt:
Implements type checking and behavior lowering for the element-of (`∈`) and no-element-of (`∉`) binary operators within the `[assert : ...]` and `[stop : ...]` annotations of group expressions.

Alternatively, `in` and `!in` can be used as aliases to `∈` and `∉` respectively.

Since operation definitions are sets of instructions, the semantics of the element-of operators behaves as expected like the element-of operator of a set. To the left side of the expression we allow a reference to some captured instruction, either bound via an `exist .. then` or `forall .. then` expression index, or via a group reference.

Examples:
```vadl
VLIW(0) ∈ ALU_FP
i ∈ ALU_FP
```

This is the last remaining piece for the group expression/annotations required for VLIW support, so I re-enabled those in the `hexagon.vadl` spec. This required some slight adaptions, to use the logical or/and operators (`||`, `&&`) rather than the bitwise operators.


rascmatt:
As discussed with @AndreasKrall in our weekly meeting today, the `in` and `!in` operators are expected to be alternatives to the `∈` and `∉`, so I will extend the frontend with that.

rascmatt:
Done, added the operator aliases `in` and `!in`, as discussed above.

sys/hexagon/hexagon.vadl
flofriday, line 1: **praise:** Nice to finally being able to reenable them ☺️

vadl-frontend/main/vadl/ast/AstUtils.java
```diff
@@ -90,7 +91,7 @@ static BuiltInTable.BuiltIn getOperatorBuiltIn(Operator operator, List<Type> arg
     }
 
     String finalOperatorSymbol = symbol;
-    var builtIns = operatorLookupTable.getOrDefault(finalOperatorSymbol, List.of());
```
flofriday, line 91: **question:** Why is it necessary to return a mutable list here?

vadl-frontend/test/resources/frontend-snapshots/typechecker/operationElementOfTrivialWarning.vadl
```diff
@@ -0,0 +1,52 @@
...
+}
+
+
+//  Reported Diagnostics:
+//
+//  warning: This expression is always false
+//       ╭── test/resources/frontend-snapshots/typechecker/operationElementOfTrivialWarning.vadl:27:13
+//       │
+//    27 │   [assert : VLIW(0) ∈ OpB]
+//       │             ^^^^^^^^^^^^^^^
```
flofriday, line 39: **minor:** There seems to be a problem with the source location because the last thing highlighted is still the bracket of the annotation. 🤔

This is probably an issue in the grammar because the in-expressions probably uses the next location of the next token that we're gonna park.

vadl-frontend/main/vadl/ast/FrontendBuiltIns.java
```diff
@@ -46,16 +46,78 @@ final class FrontendBuiltIns {
...
+              Type.relation(PseudoFormatType.class, PseudoFormatType.class, BoolType.class))
+          .takesDefault()
+          .returns(Type.bool())
+          .build();
+
+  static final List<BuiltInTable.BuiltIn> operationElementOfPredicates =
+      List.of(OP_ELEM_OF, OP_NOT_ELEM_OF, OP_IN, OP_NOT_IN);
+
+  private FrontendBuiltIns() {
+  }
```
flofriday: **question:** Is there a good reason that we add those functions into the front-end built-ins and not the global built-ins list?

The front end built-ins list is mostly a workaround or hack because we depend on some types that only exist in the front end, and usually even if a function only makes sense in the front end, we'll still add it to the global built-ins list. Maybe we should have documented that better.

vadl-frontend/main/vadl/ast/TypeChecker.java
```diff
@@ -1042,6 +1044,28 @@ private BuiltInCheckResult unCachedCheckBuiltin(BuiltInTable.BuiltIn builtIn, Li
       }
     }
 
+    if (args.size() == 2 && FrontendBuiltIns.operationElementOfPredicates.contains(builtIn)) {
+      // Static checks for operation element predicates
+      final PseudoFormatType l = (PseudoFormatType) args.getFirst().type();
+      final PseudoFormatType r = (PseudoFormatType) args.getLast().type();
```
flofriday: **question:** Can we really guarantee here that the types are always pseudoformattypes?

vadl-frontend/main/vadl/ast/TypeChecker.java
```diff
@@ -1042,6 +1044,28 @@ private BuiltInCheckResult unCachedCheckBuiltin(BuiltInTable.BuiltIn builtIn, Li
...
+        // always false, and for ∉ and `!in` it's always true.
+        final boolean constVal = builtIn == OP_NOT_ELEM_OF || builtIn == OP_NOT_IN;
+        final var op = Objects.requireNonNull(r.operations().stream().findFirst().orElse(null));
+        DeferredDiagnosticStore.add(
+            warning("This expression is always %s".formatted(constVal), location)
+                .description(
+                    "None of the possible concrete instructions matched by the left side "
+                        + "are part of operation `%s`.", op.identifier().name)
+                .build());
+      }
```
flofriday, line 1087: **praise:** Nice warning

flofriday (commented):
Well done 🎉

Code looks quite nice, mostly some questions for me I think we can merge this soon 🚀

vadl-frontend/main/vadl/ast/AstUtils.java
rascmatt: An immutable collection will throw an `UnsupportedOperationException` on the next line when calling `removeIf` (regardless of the filter expression).

Of course if we cannot find the builtin, it does constitute an error, but this is handled gracefully in the switch, where we throw a meaningful error message.

vadl-frontend/main/vadl/ast/TypeChecker.java
rascmatt: Right now, yes. These operators are only used in combination with PseudoFormatTypes (or e.g. `in` is already resolved by the parser in case of `let`, `forall-then` and `exists-then` expressions directly via the grammar).

However I think it still makes sense to add a safety net to the cast, and throw a meaningful error message, in case this ever changes.

vadl-frontend/main/vadl/ast/FrontendBuiltIns.java
rascmatt: Yes, the `PseudoFormatType` is one of those types that only exist in the frontend. It iteslf references some Ast Nodes (OperationDefinition, FormatDefinition), which are also part of the frontend.

However, thinking about it a bit more closely, I think it is anyways an issue that all those builtins are implemented over frontend- exclusive types. In fact, currently those types are silently passed to the backend, where they shouldn't (and can't) be used.

However, since the PseudoFormatType is so closely coupled to Ast nodes, and must be, since quite a bit of validation and type checking takes place on it, I'll introduce a similar type to be used in the VIAM, referencing other viam definitions/types. The behavior/viam lowering then converts those types.

This way we can also get rid of the frontend-exclusive built-in table.

vadl-frontend/test/resources/frontend-snapshots/typechecker/operationElementOfTrivialWarning.vadl
rascmatt: hmm, strange. I'll have a look at it. But don't think it's related to this PR, so can we create an issue for it and investigate separately?

Jozott00 (approved):
LGTM

vadl-frontend/test/resources/frontend-snapshots/typechecker/operationElementOfTrivialWarning.vadl
flofriday: I guess so 👍

rascmatt: I created #1061

vadl-frontend/main/vadl/ast/FrontendBuiltIns.java
flofriday: > In fact, currently those types are silently passed to the backend, where they shouldn't (and can't) be used.

There _should_ be a step in the lowering preventing this in `ViamLowering::getViamType` that checks that frontend types never reach the backend. However, I know that it's easy to miss and that there are some wholes.

In general yes, in that case we should probably use a different structure. Maybe we can even get away with types that don't link to definitions at all and only store the name of them or something.

vadl-frontend/main/vadl/ast/BehaviorLowering.java
```diff
@@ -845,13 +846,8 @@ private ExpressionNode fetch(Expr expr) {
...
     expressionCache.put(expr, result);
-    // FIXME: Should this really be here?
-    result.ensure(!(result.type() instanceof ConstantType),
-        "Constant types must not exist in the VIAM");
-    result.ensure(!(result.type() instanceof FormatType),
-        "Format types must not exist in the VIAM");
-    result.ensure(!(result.type() instanceof TensorType),
-        "Tensor types must not exist in the VIAM");
+
+    result.setType(getViamType(result.type()));
```
flofriday, line 850: **praise** Great, catch, because, no we shouldn't do this here 🎉

vadl-frontend/main/vadl/ast/FrontendBuiltIns.java
flofriday, line 1: **praise:** Great to remove this hack :)

flofriday (approved):
Looks good now :)

PR #1043, opened 2026-07-21
asm: Add semantic predicates to generated rules

benjaminkasper99:
This PR adds semantic predicate generation to asm rule generation. Semantic predicates are used to differentiate instructions with the same syntax.

Below is the detailed explanation copied from documentation:

> If there are multiple instructions of a single syntax kind, they are merged into a single rule by differentiating according to the first immediate operand.
> As example say there are three versions of an ADDI instruction ADDI_S, ADDI and ADDI_L, but all with the same syntax: `addi rd, rs1, imm`. Upon parsing the assembler needs to decide which of version of ADDI should be emitted.
> To do this we create a merged rule where each version of the instruction is an alternative guarded by a semantic predicate, calling the predicate of the respective immediate operand. The alternatives are ordered by instruction bit size ascending, such that the assembler chooses the smallest instruction for which the predicate holds. In the case that the immediate operand is a label, we fall back to the longest version of the instruction.
> Conceptually the resulting merged rule then looks like this:
> ```java
> ADDI_MERGED_RULE @instruction :
>   // parse up until immediate operand in question
>   "ADDI"
>   // parse other operands up until immediate into local vars
>   // default case first
>   inst = (
>     // local var usages up to immediate
>     mnemonic = addi_l_mnemonic<> @operand
>     offset = Label @operand
>     // parse operands after immediate
>   ) @instrution
>   |(
>   var constant_var = Integer
>   inst = ( ?(addi_s_imm_predicate(constant_var))
>     // local var usages up to immediate
>     mnemonic = addi_s_mnemonic<> @operand
>     imm = constant_var @operand
>     // parse operands after immediate
>   ) @instruction
>   | inst = ( ?(addi_imm_predicate(constant_var))
>     // local var usages up to immediate
>     mnemonic = addi_mnemonic<> @operand
>     imm = constant_var @operand
>     // parse operands after immediate
>   ) @instruction
>   | inst = (
>     // local var usages up to immediate
>     mnemonic = addi_l_mnemonic<> @operand
>     imm = constant_var @operand // if val does not fit, error gets thrown in parser
>     // parse operands after immediate
>   ) @instruction
>   )
> ;
> ```

### DIFF
--- a/sys/v-risc/ABI_no_suffix.vadl
+++ b/sys/v-risc/ABI_no_suffix.vadl
@@ -80,45 +80,4 @@ application binary interface ABI for VarRISC32 = {
 }
 
 [ generate rules ]
-assembly description AD for ABI = {
-    function l_suffix(baseMnemonic: String) -> String = baseMnemonic + "_L"
-
-    function l_suffix_if_large_val(baseMnemonic: String, val: SInt<64>) -> String = if val > -1025 && val < 1024
-                                                                                    then baseMnemonic
-                                                                                    else ( baseMnemonic, "_L" )
-    grammar = {
-
-      Reg @register:
-        reg = Register ","
-      ;
-
-      BranchPseudoIds: "BEQZ" | "BNEZ" | "BLTZ" | "BLEZ" | "BGEZ" | "BGTZ" ;
-
-      BranchPseudoInstruction @instruction:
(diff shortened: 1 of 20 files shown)


Jozott00 (commented):
LGTM

Jozott00 (approved):

PR #1045, opened 2026-07-26
cosim: Remove unnecessary allocations in get_cpu_state

Giftzwerg02:
I managed to find a trick to further reduce unnecessary allocations / copying in the `get_cpu_state()` function, which is essentially the hot-loop of the cosimulation plugin.

Previously, the plugin
- **(a)** allocated a new GByteArray buffer which reads the data for each register (since a `GByteArray *` is required by the QEMU API)
- **(b)** and then copy the read bytes into the actual shared-memory buffer.

**(a)** is easily solved by simply only creating a GByteArray once and reusing it without freeing it
**(b)** is solved by abusing the fact that `qemu_plugin_read_register` internally never seems to be calling anything else but `g_byte_array_append` -> as long as the initial capacity of the `GByteArray` is large enough, this is essentially the same as `buf[len++] = data`.

Combining both gives a nice 1.33x - 2x performance increase (at least on the local tests that I ran).

### DIFF
--- a/vadl/main/resources/templates/iss/contrib/plugins/cosimulation.c
+++ b/vadl/main/resources/templates/iss/contrib/plugins/cosimulation.c
@@ -34,7 +34,10 @@
 #include <time.h>
 #include <unistd.h>
 
+
 QEMU_PLUGIN_EXPORT int qemu_plugin_version = QEMU_PLUGIN_VERSION;
+
+// TODO(perf): generate the defines in the iis generator to properly match the sizes
 
 // adjust as needed
 // NOTE: try to keep these values as small as possible to minimize memory usage
@@ -83,6 +86,7 @@ static qemu_plugin_id_t plugin_id;
     }                                                                          \
   } while (0)
 
+
 typedef struct {
   struct qemu_plugin_register *handle;
   const char *name;
@@ -97,6 +101,8 @@ typedef struct {
   size_t len;
   char value[SHMSTRING_MAX_LEN];
 } SHMString;
+
+// TODO(perf): try SOA representation
 
 typedef struct {
   int size;
@@ -231,6 +237,7 @@ static GPtrArray *registers_init(int vcpu_index) {
     reg->handle = rd->handle;
     reg->feature = rd->feature;
     reg->name = rd->name;
+
     g_ptr_array_add(registers, (gpointer)reg);
   }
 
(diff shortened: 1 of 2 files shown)


vadl/main/resources/templates/iss/contrib/plugins/cosimulation.c
```diff
@@ -242,7 +249,12 @@ static GPtrArray *registers_init(int vcpu_index) {
   return registers;
 }
 
+static GByteArray *buf;
 static void get_cpu_state(unsigned int cpu_index, SHMCPU *cpu) {
+  if (buf == NULL) {
+    buf = g_byte_array_sized_new(MAX_REGISTER_DATA_SIZE);
```
Jozott00, line 255: Does this allocate an actual byte array? If yes, why do we need it when we actually override the internals with the shm pointers anyway?

Jozott00 (approved):
Nice! LGTM to me in general, just the question why the buffer allocation required.

vadl/main/resources/templates/iss/contrib/plugins/cosimulation.c
Jozott00: I guess it is because of the capacity right? Does the memory allocation happen lazily? So that this constructor doesn't actually perform any large memory allocation?

Giftzwerg02: So you mean something like just a stack-allocated `static GByteArray buf`?
Afaik this doesn't work because glib uses a `GRealArray` internally (with further properties, e.g. the capacity) which is not exposed by the header file. *Technically* I could try to mimic this as well, but I think there is no real reason performance-wise to do this.

Giftzwerg02: > I guess it is because of the capacity right? Does the memory allocation happen lazily? So that this constructor doesn't actually perform any large memory allocation?

`g_byte_array_sized_new` afaik essentially just allocates the amount of bytes provided in the argument and `MAX_REGISTER_DATA_SIZE` is currently set 256 (bytes). So this definitely isn't a problem w.r.t allocation size.

Jozott00: Ok, then this is fine

PR #1048, opened 2026-08-03
lsp: Refine jump to definition position

### LINKED ISSUE #904: lsp: Refine jump to definition position
When jumping to a definition position IDE's usually jump not to the start of the definition but the point within the position where this identifier is defined.
In OpenVADL this should be easy as definitions that define new names usually implement the interface `IdentifyableNode` from which you can get the identifier and it's location.

**IntelliJ**
https://github.com/user-attachments/assets/c3e37fdb-c645-4470-bf36-f9a9e80c139e

**OpenVALD LSP**
https://github.com/user-attachments/assets/2a356360-fe83-4174-9886-21e4a3af46ad

jaraonthe-dot-net:
Closes #904

### DIFF
--- a/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
+++ b/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
@@ -152,7 +152,7 @@ public class VadlTextDocumentService implements TextDocumentService {
         return emptyDefinitionResult();
       }
       var target = identifier.target();
-      if (target == null || target.location().path() == null) {
+      if (target == null || !target.location().isValid()) {
         return emptyDefinitionResult();
       }
       var targetUri = toUri(Objects.requireNonNull(target.location().path()));
@@ -162,8 +162,9 @@ public class VadlTextDocumentService implements TextDocumentService {
         return emptyDefinitionResult();
       }
 
-      // targetSelectionRange is the location to navigate to, whereas targetRange refers to the
-      // entire definition code (i.e. less important information)
+      // targetSelectionRange is the location the cursor jumps to, whereas targetRange refers to the
+      // whole definition we jump to.
+      // See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink
       var targetRange = targetDocument.calculateUtf16Range(target.location());
       Range targetSelectionRange = targetRange; // Fallback
       if (target instanceof IdentifiableNode identifiableTarget) {
--- a/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
+++ b/vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
@@ -55,6 +55,8 @@ import org.slf4j.LoggerFactory;
 import vadl.ast.Ast;
 import vadl.ast.Frontend;
 import vadl.ast.VadlParser;
+import vadl.ast.nodes.IdentifiableNode;
+import vadl.ast.nodes.IsId;
 import vadl.error.Diagnostic.MsgType;
 import vadl.error.DiagnosticList;
 import vadl.utils.DiskVirtualFileSystem;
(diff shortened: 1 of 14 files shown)


flofriday:
I'm on holiday till Saturday enjoying the woods without a computer, review might take some time.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -137,35 +139,62 @@ public void didSave(DidSaveTextDocumentParams params) {
...
           position
       );
-
-      if (location == null || location.path() == null) {
-        return definitionResult(null);
+      if (identifier == null) {
+        return emptyDefinitionResult();
+      }
+      var target = identifier.target();
+      if (target == null || target.location().path() == null) {
```
flofriday, line 155: **suggestion:** You can use target.location().isValid() instead of checking the path here.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -137,35 +139,62 @@ public void didSave(DidSaveTextDocumentParams params) {
...
-        return definitionResult(null);
+        log.debug("Unexpected: Definition target file {} does not exist", targetUri);
+        return emptyDefinitionResult();
       }
 
-      var lspLocation = new Location(targetDocument.uri,
-          targetDocument.calculateUtf16Range(location));
-      return definitionResult(lspLocation);
+      // targetSelectionRange is the location to navigate to, whereas targetRange refers to the
+      // entire definition code (i.e. less important information)
```
flofriday, line 166: **minor:** I don't like the "less important information" here, it doesn't really describe what it means. How about we say target selection range is the position the cursor jumps to and the target range is the whole definition we jump to. Since we use the same words as a specification, we can also link to the specification here: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#locationLink

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/isa.snapshot
```diff
@@ -47,4 +78,15 @@ Test method: vadl.lsp.integration.GotoDefinitionTest.mainTest, Test case: isa
...
+    // GOTO POSITION 5:40 - "TEST" points to isa definition
+    instruction set architecture TEST = {
+      register X : Bits<5> -> Bits<32>
+    }
+    application binary interface ABI for <ORIGIN SELECTION RANGE>TEST</> = {
+      alias register hello = X(0)
+    
+      return address = hello
+    }
+  ================
```
flofriday, line 92: **nitpick:** I think we can merge these three into printing the source code only one time and having the different text into the s in one unified code snipped.

vadl-lsp/test/vadl/lsp/TestUtils.java
flofriday, line 96: **nitpick:** Now that the start marker has more different names the end marker should probably also contain the name, like html.

`+ "</" + name + ">" +`

vadl-lsp/test/vadl/lsp/TestUtils.java
```diff
@@ -62,6 +62,19 @@ public static String showPositionInFile(Position position, String fileContent) {
...
+  /**
+   * Displays a range marker within the given file content.
+   *
+   * @param range the lsp range (with start and end position)
+   * @param fileContent the entire content of the file which should contain range
+   * @param name the name of this range, used as part of the marker. Should be uppercase for
+   *             aesthetic reasons.
+   * @return given fileContent with range marked, formatted for use in a snapshot
+   */
+  public static String showRangeInFile(Range range, String fileContent, String name) {
```
flofriday, line 77: **thought:** Now that we have more commonly multiple ranges inside a code snippet we want to highlight it might be usefull to simply print the snippet once and just insert all the different ranges into it, which makes it much easier to read.

```java
record NamedRange(String name, Range range) { }

public static String showRangesInFile(List<NamedRange> ranges, String fileContent) {
```

flofriday (commented):
There are some minor things we can improve here, but overall a really good implementation :tada:.

jaraonthe-dot-net:
@flofriday I've incorporated all your notes. Please review again.

flofriday (approved):
Looks good

PR #1049, opened 2026-08-03
frontend: Add float support

BadGraphixD:
- Adds float-type definition
- Adds annotation `[ IEEE : <size> ]` for float-type
- Adds annotation `[ [sticky] fe flag <flag-name> : <format-field> ]` for registers
- Adds new syntax for SymbolExpr: `id::path::<const params>(params)`
- Adds new VIAM nodes and built-in types
  - FloatFormat node
  - Constant.FloatType
  - VIAM annotation for float exception flags
  - The built-in type FloatType, which is the type of FloatFormat
- Adds constant-param-based type inference and checking for built-ins

### DIFF
--- a/vadl-frontend/main/vadl/ast/AnnotationTable.java
+++ b/vadl-frontend/main/vadl/ast/AnnotationTable.java
@@ -24,6 +24,7 @@ import static vadl.viam.ViamError.ensurePresent;
 import com.google.errorprone.annotations.concurrent.LazyInit;
 import java.math.BigInteger;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -47,6 +48,7 @@ import vadl.ast.nodes.Definition;
 import vadl.ast.nodes.DerivedFormatField;
 import vadl.ast.nodes.EncodingDefinition;
 import vadl.ast.nodes.Expr;
+import vadl.ast.nodes.FloatTypeDefinition;
 import vadl.ast.nodes.FormatField;
 import vadl.ast.nodes.GroupDefinition;
 import vadl.ast.nodes.Identifier;
@@ -69,6 +71,7 @@ import vadl.gcb.annotations.StatusRegisterAnnotation;
 import vadl.types.BitsType;
 import vadl.types.Type;
 import vadl.utils.Pair;
+import vadl.utils.functionInterfaces.QuadConsumer;
 import vadl.utils.functionInterfaces.TriConsumer;
 import vadl.viam.Abi;
 import vadl.viam.ArtificialResource;
@@ -77,6 +80,8 @@ import vadl.viam.Constant;
 import vadl.viam.Counter;
 import vadl.viam.Encoding;
 import vadl.viam.Endianness;
+import vadl.viam.FloatExceptionFlag;
+import vadl.viam.FloatFormat;
 import vadl.viam.Format;
 import vadl.viam.Group;
 import vadl.viam.Instruction;
@@ -204,7 +209,7 @@ public class AnnotationTable {
         // this handled in the VIAM lowering when constructing the ArtificialResource
         .build();
 
-    annotationOn(RegisterDefinition.class, "execution state", FormatFieldAnnotation::new)
+    annotationOn(RegisterDefinition.class, "execution state", ExecutionStateAnnotation::new)
         .check((def, annotation, lowering) -> annotation.typeCheckTarget(def))
         .applyAst((def, annotation) -> annotation.calcBitSlice(def))
         .applyViam((def, annotation, lowering) -> {
@@ -329,6 +334,56 @@ public class AnnotationTable {
           group.addAnnotation(new StopAnnotation(graph));
         })
         .build();
+
+    /// FLOAT RELATED ///
+
+    annotationOn(FloatTypeDefinition.class, "IEEE", ConstantAnnotation::new)
+        .applyAst((def, annotation) -> def.size = annotation.constant.value().intValue())
+        .applyViam((def, annotation, lowering) -> {
+          var encoding = FloatFormat.Encoding.ieee(annotation.constant.value().intValue());
+          ensure(encoding != null,
+              () -> error("Invalid IEEE encoding size", annotation)
+                  .description("The following sizes are supported: %s",
+                      Arrays.stream(FloatFormat.Encoding.values())
+                          .map(e -> Integer.toString(e.size)).collect(Collectors.joining(", ")))
+          );
+          ((FloatFormat) def).setEncoding(encoding);
+        }).build();
+
+    QuadConsumer<RegisterTensor, FloatFlagAnnotation, Boolean, FloatExceptionFlag>
+        applyViamFloatFlag;
+    applyViamFloatFlag = (reg, annotation, sticky, flag) -> {
+      var idx = annotation.index;
+      if (reg.hasAnnotation(vadl.viam.annotations.FloatFlagAnnotation.class)) {
+        var ann = reg.expectAnnotation(vadl.viam.annotations.FloatFlagAnnotation.class);
+        var setFlag = ann.get(idx);
+        ensure(setFlag == null, () -> error(
+            "Bit already mapped as " + (ann.isSticky(idx) ? "" : "non ")
+                + "sticky " + requireNonNull(setFlag).name + " flag",
+            annotation
+        ));
+        ann.set(idx, sticky, flag);
+      } else {
+        var ann = new vadl.viam.annotations.FloatFlagAnnotation();
+        ann.set(idx, sticky, flag);
+        reg.addAnnotation(ann);
+      }
+    };
+
+    for (var flag : FloatExceptionFlag.values()) {
+      annotationOn(RegisterDefinition.class, "fe flag " + flag.name, FloatFlagAnnotation::new)
+          .check((def, annotation, lowering) -> annotation.typeCheckTarget(def))
+          .applyViam((def, annotation, lowering) ->
+              applyViamFloatFlag.accept((RegisterTensor) def, annotation, false, flag))
+          .build();
+
+      annotationOn(RegisterDefinition.class, "sticky fe flag " + flag.name,
+          FloatFlagAnnotation::new)
+          .check((def, annotation, lowering) -> annotation.typeCheckTarget(def))
+          .applyViam((def, annotation, lowering) ->
+              applyViamFloatFlag.accept((RegisterTensor) def, annotation, true, flag))
+          .build();
+    }
 
     /// PROCESSOR RELATED ///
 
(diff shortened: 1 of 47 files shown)


vadl/main/vadl/viam/FloatFormat.java
```diff
@@ -0,0 +1,102 @@
...
+  @Override
+  public void accept(DefinitionVisitor visitor) {
+    visitor.visit(this);
+  }
+
+  @Override
+  public void verify() {
+    super.verify();
+    // FIXME: for now this is checked here, but this should create a diagnostic instead of ViamError
+    ensure(encoding != null, "Encoding not specified");
```
Jozott00: This is checked by the annotation right? So it should be fine to have it as a safe guard ViamError here.

vadl/main/vadl/viam/annotations/FloatFlagAnnotation.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.annotation.Nullable;
+import vadl.viam.Annotation;
+import vadl.viam.Constant;
+import vadl.viam.FloatExceptionFlag;
+import vadl.viam.RegisterTensor;
+
+/**
+ * Annotation for registers that are saved in the translation block state. Contains a
+ * {@link Constant.BitSlice} that specifies what bits of the register are saved.
+ */
```
Jozott00, line 31: I guess this comment is a copy/paste error

Jozott00 (dismissed):
VIAM side looks good! (just the copy/paste error)

I would like to leave the frontend review to @flofriday

flofriday:
I'm still on vacation till Saturday (in the woods without any laptop) so this review will take until the end of the week.

Thanks for including me @Jozott00 😊

vadl/main/vadl/viam/FloatFormat.java
BadGraphixD: If no annotation is present, then the encoding is missing, which should lead to a diagnostic error. I am not sure how the absence of an annotation should be checked.

BadGraphixD:
@flofriday please take a look at https://github.com/OpenVADL/openvadl/discussions/578#discussioncomment-17872831 before the review, as it explains the intention behind this PR and the open problem with the new syntax. Thanks!

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidBuiltitinCalls.vadl
```diff
@@ -28,7 +28,7 @@ function abc(x: Bits<8>) -> Bits<8> = VADL::adds(1, x)
 //       ╭── test/resources/frontend-snapshots/typechecker/invalidBuiltitinCalls.vadl:8:39
 //       │
 //     8 │ function abc(x: Bits<8>) -> Bits<8> = VADL::adds(1, x)
-//       │                                       ^^^^^^^^^^^^^^^^ The builtin has the signature `(BitsType, BitsType) -> StructType` but got `Bits<1>, Bits<8>`.
+//       │                                       ^^^^^^^^^^^^^^^^ The builtin has the signature `(BitsType, BitsType) -> StructType` but got `(Bits<1>, Bits<8>)`.
```
flofriday, line 31: **issue:** I don't think that change was intentional.

flofriday:
@BadGraphixD can you add some tests to show the new capabilities in the frontend?
You will find most of them in: vadl-frontend/test/resources/frontend-snapshots

vadl-frontend/main/vadl/ast/nodes/FloatTypeDefinition.java
```diff
@@ -0,0 +1,96 @@
...
+public class FloatTypeDefinition extends Definition implements IdentifiableNode, TypedNode {
+  public IdentifierOrPlaceholder identifier;
+
+  /**
+   * Represents the bit-size of the represented float format. This is used by the type-checker
+   * to infer types when the float-type is used as a constant parameter in built-ins. This is not
+   * set during parsing, and must be set by an annotation such as {@code [ IEEE : 32 ]}.
+   */
+  @Nullable
+  public Integer size;
```
flofriday, line 40: **nitpick:** We usually only use annotations for optional things, and not for required fields.

Maybe a better syntax would be: `float-type binary32: Bits<32>` which always has the bit size.

Then we only would use the annotation to specify how the format of the flow type would look like. However, even there I'm on the fence, because it needs to have some format, right?

vadl-frontend/main/vadl/ast/nodes/FloatTypeDefinition.java
```diff
@@ -0,0 +1,96 @@
...
+import vadl.utils.SourceLocation;
+
+/**
+ * Represents a float-type definition, which is used to specify float formats.
+ * <pre>{@code
+ * [ IEEE : 32 ]
+ * float-type binary32
+ * }</pre>
+ */
+@SuppressWarnings({"MissingJavadocType", "MissingJavadocMethod"})
```
flofriday: **nitpick:** I know that a lot of other classes use this suppress warning, but it's probably best practice for new classes to not do that.

vadl-frontend/main/vadl/ast/nodes/SymbolExpr.java
flofriday, line 1: **minor:** I don't love the changes in this file, but to be honest, I didn't love this file in the first place. I tried to get rid of it multiple times, but so far we kept it because of compatibility and also because it's enshrined into the macrosystem. However, with the new changes, I'm not sure it even makes sense to keep any more also symbol expression was never really descriptive of a name and quite confusing. So maybe we can come up with something better here.

vadl-frontend/main/vadl/ast/AnnotationTable.java
```diff
@@ -204,7 +209,7 @@ public class AnnotationTable {
         // this handled in the VIAM lowering when constructing the ArtificialResource
         .build();
 
-    annotationOn(RegisterDefinition.class, "execution state", FormatFieldAnnotation::new)
+    annotationOn(RegisterDefinition.class, "execution state", ExecutionStateAnnotation::new)
```
flofriday, line 212: **question:** Was this change intentional?

vadl-frontend/main/vadl/ast/ConstantEvaluator.java
```diff
@@ -164,10 +164,18 @@ public ConstantValue evalBuiltin(BuiltInTable.BuiltIn builtin, List<ConstantValu
         }
       }
 
+      if (!builtin.signature().constArgTypeClass().isEmpty()) {
```
flofriday, line 170: **nitpick:** Let's add a comment here to explain that.

vadl-frontend/main/vadl/ast/MacroExpander.java
```diff
@@ -707,6 +713,14 @@ public Definition visit(ConstantDefinition definition) {
...
   }
 
+  @Override
+  public Definition visit(FloatTypeDefinition definition) {
+    return new FloatTypeDefinition(
+        expandExpr(definition.identifier),
+        copyLoc(definition.loc)
+    );
+  }
+
```
flofriday, line 723: **nitpick:** I would love to see some tests for this.

vadl-frontend/main/vadl/ast/SymbolTable.java
```diff
@@ -1178,6 +1179,16 @@ public Void visit(InstructionSetDefinition definition) {
       return null;
     }
 
+    @Override
+    public Void visit(FloatTypeDefinition definition) {
+      beforeTravel(definition);
+      // FloatTypeDefinition has no @Child fields, so it's not in NodeChildrenRegistry,
```
flofriday, line 1185: **question:** How does the absence in the NodeChildregistry change the code here?

Even if it were there with an empty list of children this code would look the same, right?

That we need to visit the annotations manually here does look absurd, I guess we could move that into the beforeTravel of definitions.

vadl/main/vadl/types/FloatType.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public class FloatType extends Type {
+
+  protected FloatType() { }
+
+  @Override
+  public String name() {
+    return "FloatType";
+  }
+}
```
flofriday, line 44: **question:** Wouldn't that mean that all float types are interchangeable? Maybe we should bind them to a name so that we have name-based identity.

vadl/main/vadl/types/BuiltInTable.java
flofriday, line 1: **nitpick:** I don't love how the float changes pollutes all of our previous assumptions about the typesystem and builtins :/

vadl/main/vadl/types/Type.java
```diff
@@ -188,6 +202,25 @@ public static StructType struct(String field1, Type type1,
...
+                                  String field5, Type type5) {
+    var fields = new LinkedHashMap<String, Type>();
+    fields.put(field1, type1);
+    fields.put(field2, type2);
+    fields.put(field3, type3);
+    fields.put(field4, type4);
+    fields.put(field5, type5);
+    return struct(fields);
+  }
+
```
flofriday: **nitpick:**  This helper seems oddly specific and is only used one. I'm not sure it's that's helpful.

vadl/main/vadl/types/Type.java
```diff
@@ -403,6 +480,7 @@ public static GroupType group(Type elementType, UIntType lengthType, UIntType bi
   /// These are all the builtin types that exist in the language.
   /// Some of them cannot be initialized like them since they require a size, like `SInt<16>`
   /// which is why they are named bases.
-  public static final Set<String> builtinTypeBases =
-      Set.of("Bool", "String", "Bits", "UInt", "SInt", "Instruction", "FetchResult");
+  public static final Set<String> builtinTypeBases = Set.of(
+      "Bool", "String", "Bits", "UInt", "SInt", "Instruction", "FetchResult"
+  );
 }
```
flofriday: **question:** Was there a specific reason for this reformat?

vadl-frontend/main/vadl/ast/TypeChecker.java
```diff
@@ -936,21 +938,28 @@ List<Expr> applyCastToArgs(List<Expr> args) {
...
       var inner = store.get(builtIn);
       if (inner == null) {
         return null;
       }
-      return inner.get(argTypes);
+      var innerInner = inner.get(constArgs);
+      if (innerInner == null) {
+        return null;
+      }
+      return innerInner.get(argTypes);
```
flofriday: **nitpick:** innerInner is an especially unuseful name. Maybe we can pick better names here.

vadl-frontend/main/vadl/ast/TypeChecker.java
flofriday, line 1: **minor:** I really don't love how builtin selection now requires constArgs (I would maybe call them typearg). 
It does make the code more complex especially since we only use it for float types.

flofriday (commented):
It's quite impressive how fast you got something working, especially with how many things you had to touch. 💪

I still think that the implementation got quite a bit more complicated now, and maybe it's worth going back to the language design stage to find a simpler implementation. I know I added quite a few comments here but that doesn't mean that it's bad.
Shoot me a DM if you want to go through all this stuff in a meeting.

vadl-frontend/main/vadl/ast/SymbolTable.java
BadGraphixD: The traversal of the annotations is normally also done by the code in the NodeChildRegistry, because the field storing the annotations is annotated with `@Child` in Definition.class. So you would expect NodeChildRegistry to also traverse the annotations of FloatTypeDefinition, but it doesn't. My guess is because FloatTypeDefinition does not have any fields annotated with `@Child` (even though its parent class, Definition.class has), and is thus ignored during the generation of NodeChildRegistry.

vadl-frontend/main/vadl/ast/AnnotationTable.java
BadGraphixD: Yes. The new FloatFlagAnnotation has much in common with the old FormatFieldAnnotation, so i made both the new FloatFlagAnnotation and the new ExecutionStateAnnotation inherit from FormatFieldAnnotation.

vadl/main/vadl/types/Type.java
BadGraphixD: No, its a remain of some things I tried and later deleted. I will remove the change.

vadl/main/vadl/types/Type.java
BadGraphixD: Agreed. It now uses the `StructType struct(Map<String, Type> types)` helper.

vadl-frontend/main/vadl/ast/TypeChecker.java
BadGraphixD: Agreed. I have rewritten the method a bit, I hope the new version is clearer to read.

vadl-frontend/main/vadl/ast/nodes/FloatTypeDefinition.java
BadGraphixD: I have also had this idea. The original idea was to have a definition for float formats, where annotations can be attached to specify various aspects of float formats. Since float formats are numerous and diverse, we did not want to settle for a fixed syntax for float format definitions. But since encoding bit-size is universal for float formats, it makes sense to include it into the definition.

vadl/main/vadl/types/FloatType.java
BadGraphixD: I think my documentation is quite lacking. I have added JavaDocs to the FloatType class, which hopefully clarify things.

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidBuiltitinCalls.vadl
BadGraphixD: It was intentional, because the error message can now also print the given types of the constant args, which are printed in angle brackets like this: `<const, arg, types, ...>(normal, arg, types, ...)`. If no const args are present, only the parentheses are printed. If you don't like this change, we could just print all args like before (just separated by commas), but I think when constant args are present, it makes things clearer. What do you think?

BadGraphixD: Removed the brackets. All parameters and given types are now in a list separated with commas.

vadl-frontend/main/vadl/ast/nodes/FloatTypeDefinition.java
BadGraphixD: We decided to leave it this way for now (on 11.08.26).

vadl-frontend/main/vadl/ast/MacroExpander.java
BadGraphixD: See `frontend-snapshots/macro/floatTypeTest.vadl`

vadl-frontend/main/vadl/ast/SymbolTable.java
BadGraphixD: See #1053

flofriday (approved):
Looks good!

Issue #1022, opened 2026-06-09
viam: Inline functions only when necessary


BadGraphixD:
Currently the `FunctionInlinerPass` inlines every behaviour that is used by a `DefProp.WithBehaviour`. This means that all `Function`s are inlined. But for some functions, this is unnecessary (e.g. when they are only used by `assembly` definitions). We want to avoid inlining these functions, as it generally only increases graph sizes with no benefit (in case of the `aarch32`, it hurts performance significantly).

PR #1051, opened 2026-08-06
lcb: Inline instructions assembly functions

Tbswss:
This PR introduces a new pass in the GCB pipeline, which inlines assembly functions of instructions. The inlining of assembly-behaviours was removed in https://github.com/OpenVADL/openvadl/issues/1022 and caused problems in the assembly-parser-generator.

~This PR fixes the generation of the LLVM instruction assembly printer. Functions defined in VADL and used in the assembly-definition of an instruction were not inlined correctly anymore, which caused "undefined reference" errors while compiling the generated compiler.~
~This is a regression introduced in https://github.com/OpenVADL/openvadl/issues/1022.~

### DIFF
--- a/vadl/main/vadl/lcb/codegen/assembly/AssemblyInstructionPrinterCodeGenerator.java
+++ b/vadl/main/vadl/lcb/codegen/assembly/AssemblyInstructionPrinterCodeGenerator.java
@@ -24,7 +24,6 @@ import vadl.lcb.passes.llvmLowering.tablegen.model.tableGenOperand.ReferencesImm
 import vadl.pass.PassResults;
 import vadl.viam.PrintableInstruction;
 import vadl.viam.graph.control.ReturnNode;
-import vadl.viam.passes.functionInliner.Inliner;
 
 /**
  * Wrapper class for the visitor.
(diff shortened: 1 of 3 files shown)


vadl/main/vadl/lcb/codegen/assembly/AssemblyInstructionPrinterCodeGenerator.java
```diff
@@ -55,8 +56,10 @@ public CppFunctionCode generateFunctionBody(
     var handler2 =
         new AssemblyInstructionPrinterLabelHandler(passResults, instruction, tableGenInstruction);
 
-    var returnNodes =
-        instruction.assembly().function().behavior().getNodes(ReturnNode.class).toList();
+    var assemblyBehaviorCopy = instruction.assembly().function().behavior().copy();
```
kper: Do you need the copy? Or why is it there?

kper (approved):

vadl/main/vadl/lcb/codegen/assembly/AssemblyInstructionPrinterCodeGenerator.java
Tbswss: I didn't want to modify the original behaviour in place, but the copy is indeed not necessary. I have removed it.

Tbswss:
I have found another issue introduced by the linked Issue / PR, I will update the branch tomorrow with a more general / appropriate fix.

Tbswss:
I have updated the PR @kper, as I found other issues caused by the linked PR / issue.

kper (approved):

Issue #1016, opened 2026-06-07
lsp: The model problem


flofriday:
During normal compilation models are evaluated/expanded during parsing and afterwards all references to them are pruned out of the AST.

During development it _would_ be nice if the LSP could provide also help inside models.
However as far as I can see this is really hard because models depend highly on the context they are used.

For example:
```
model abc(): Ex = {
   baum + 1
}
```

There is nothing we can say for certain about the body of this model. The type of the expression is completely dependend on the context it is called on and what `baum` binds to.

 ### Idea 0: Ignore model bodies
~~This is what we currently do~~, and for normal compilation what we want since there is nothing to be gained from analysing them

**Update:** In the meeting today (2026-06-08) we figured out that the LSP uses the values from the first invocation.

### Idea 1: Island model
Since we don't know the context of the models we could always treat them as independent islands without any context.
Obviously in many cases like the example above this would not help since we still don't know what `baum` refers to but in some other cases it could help.

```
model abc(): Defs = {
   constant apfel = 3 as Bits<8>
   constant strudel = let teig = 6 in apfel + teig 
}
```

Like here all the necessary context to provide goto definitions and type information and signature information is contained inside the model and could work.

To be fair this is quite a bit of work, especially for someone who new to the frontend, because we now have to symbol resolve and typecheck code where we have to assume that it might not be valid code and we still cannot report any errors for the macro bodies.

### Idea 2: Analyse all macro call sites.

I think you maybe _could_ analyse all macro call sites and if they are all in the same scope or have some common ancestor this could be used as a scope.

Obviously this would provide the best results but it is also the hardest implementation and I'm not yet sure how this would be implemented.
This is hardent because the Symbol resolver doesn't really implement multiple ancestors but instead manges them with a hack (by copying all symbols in one scope into the current scope, emulating multiple ancestry.

Also if another file imports the macro you need to be aware of that, because that could minimize the common ancestor logic.

AndreasKrall:
I think that idea 1 does not help much. So only 0 and 2 are viable options. But I believe that currently we do not have the man power to tackle idea 2 and have higher priority tasks.

flofriday:
### Idea 3: Multiple results of multiple invocations
In the meeting (2026-06-08) @Jozott00 suggested that we just take expanded content and if the same node evaluates to different types in different invocations we display multiple types in the popup.

As far as implementation goes this should be quite easy, since the LSP simply no longer has to return a single result but possibly returns multiple. 

[...]

PR #1052, opened 2026-08-09
lsp/frontend: Fix Model SourceLocation

jaraonthe-dot-net:
This attempts to fix various bogus hover data provided by the language server; by doing two things:
- Ignoring entities situated inside model definitions, as we cannot know if their type is only valid for one particular model invocation.
- Have proper `ExpandedLocation` data for all entities that a model invocation has been expanded to. This is the only way to tell that an entity is connected to a model invocation.

I had to fix `SourceLocation.join()` and the `MacroExpander` - I do not know if my changes cause issues elsewhere.

For testing it's best to use VSCode as lsp client, as it properly highlights the area that is being hovered on - in IntelliJ I see no such indication.

Some examples where the behavior change is apparent (all in `rv3264im.vadl`):
- line 73: `$zero45` no longer has hover
- line 231 & 511: no longer have hover that spans from model definition to model invocation
- also on line 231: `addr` still has hover showing a type that is provided by the model (CTRL-Click on it, you'll see what I mean)
- on line 20, change to `Arch64` -> lines 250-280 no longer trigger a hover ranging over the entire model invocation

### DIFF
--- a/vadl-frontend/main/vadl/ast/MacroExpander.java
+++ b/vadl-frontend/main/vadl/ast/MacroExpander.java
@@ -157,6 +157,7 @@ import vadl.ast.nodes.SymbolExpr;
 import vadl.ast.nodes.TemplateParam;
 import vadl.ast.nodes.TypeLiteral;
 import vadl.ast.nodes.TypedFormatField;
+import vadl.ast.nodes.UnOp;
 import vadl.ast.nodes.UnaryExpr;
 import vadl.ast.nodes.UsingDefinition;
 import vadl.ast.nodes.WildcardLiteral;
@@ -316,12 +317,14 @@ class MacroExpander
         for (var entry : recordInstance.entries) {
           entries.add(expandNode(entry));
         }
-        yield new RecordInstance(recordInstance.type, entries, recordInstance.sourceLocation);
+        yield new RecordInstance(recordInstance.type, entries,
+            copyLoc(recordInstance.sourceLocation));
       }
       case EncodingDefinition.EncsNode encs -> expandEncs(encs);
       case PlaceholderNode placeholderNode -> expand(placeholderNode);
       case MacroInstanceNode macroInstanceNode -> expand(macroInstanceNode);
       case MacroMatchNode macroMatchNode -> expand(macroMatchNode);
+      case UnOp unOp -> new UnOp(unOp.operator, copyLoc(unOp.location));
       case null, default -> node;
     };
   }
@@ -1066,8 +1069,8 @@ class MacroExpander
   public Definition visit(ProcessDefinition processDefinition) {
     var templateParams = new ArrayList<TemplateParam>(processDefinition.templateParams.size());
     for (var templateParam : processDefinition.templateParams) {
-      templateParams.add(new TemplateParam(templateParam.identifier(),
-          templateParam.type,
+      templateParams.add(new TemplateParam(expandExpr(templateParam.identifier()),
+          expandExpr(templateParam.type),
           templateParam.value == null ? null : expandExpr(templateParam.value)));
     }
 
@@ -1128,7 +1131,7 @@ class MacroExpander
   public Definition visit(SpecialPurposeRegisterDefinition definition) {
     return new SpecialPurposeRegisterDefinition(
         definition.purpose,
-        definition.exprs,
+        expandExprs(definition.exprs),
         copyLoc(definition.loc)
     );
   }
@@ -1281,12 +1284,12 @@ class MacroExpander
   @Override
   public Definition visit(AsmDescriptionDefinition definition) {
     return new AsmDescriptionDefinition(
-        definition.id,
-        definition.abi,
-        definition.modifiers,
-        definition.directives,
-        definition.rules,
-        definition.commonDefinitions,
+        expandExpr(definition.id),
+        expandExpr(definition.abi),
+        expandDefinitions(definition.modifiers),
+        expandDefinitions(definition.directives),
+        expandDefinitions(definition.rules),
+        expandDefinitions(definition.commonDefinitions),
         copyLoc(definition.loc)
     );
   }
@@ -1729,14 +1732,14 @@ class MacroExpander
   private List<Parameter> expandParams(List<Parameter> params) {
     var expandedParams = new ArrayList<>(params);
     expandedParams.replaceAll(param ->
-        new Parameter(param.identifier(), expandExpr(param.typeLiteral)));
+        new Parameter(expandExpr(param.identifier()), expandExpr(param.typeLiteral)));
     return expandedParams;
   }
 
   private List<StageOutputDefinition> expandStageOutputs(List<StageOutputDefinition> outputs) {
     var expanded = new ArrayList<>(outputs);
     expanded.replaceAll(param ->
-        new StageOutputDefinition(param.identifier(), expandExpr(param.typeLiteral)));
+        new StageOutputDefinition(expandExpr(param.identifier()), expandExpr(param.typeLiteral)));
     return expanded;
   }
 
--- a/vadl-frontend/main/vadl/ast/MacroExpander.java
+++ b/vadl-frontend/main/vadl/ast/MacroExpander.java
@@ -46,6 +46,7 @@ import vadl.ast.nodes.AsmModifierDefinition;
 import vadl.ast.nodes.AssemblyDefinition;
 import vadl.ast.nodes.AssignmentStatement;
 import vadl.ast.nodes.BasicSyntaxType;
+import vadl.ast.nodes.BinOp;
 import vadl.ast.nodes.BinaryExpr;
 import vadl.ast.nodes.BinaryLiteral;
 import vadl.ast.nodes.BlockStatement;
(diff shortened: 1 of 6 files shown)


vadl-frontend/main/vadl/ast/MacroExpander.java
```diff
@@ -1114,7 +1117,7 @@ public Definition visit(AbiSequenceDefinition definition) {
   public Definition visit(SpecialPurposeRegisterDefinition definition) {
     return new SpecialPurposeRegisterDefinition(
         definition.purpose,
-        definition.exprs,
```
flofriday, line 1131: **praise:** Great catch.

vadl-frontend/main/vadl/ast/MacroExpander.java
```diff
@@ -315,12 +316,14 @@ private <T extends Node> T expandNode(T node) {
...
         }
-        yield new RecordInstance(recordInstance.type, entries, recordInstance.sourceLocation);
+        yield new RecordInstance(recordInstance.type, entries,
+            copyLoc(recordInstance.sourceLocation));
       }
       case EncodingDefinition.EncsNode encs -> expandEncs(encs);
       case PlaceholderNode placeholderNode -> expand(placeholderNode);
       case MacroInstanceNode macroInstanceNode -> expand(macroInstanceNode);
       case MacroMatchNode macroMatchNode -> expand(macroMatchNode);
+      case UnOp unOp -> new UnOp(unOp.operator, copyLoc(unOp.location));
```
flofriday, line 329: **suggestion:** Let's also add BinOp here.

vadl-frontend/main/vadl/ast/MacroExpander.java
flofriday, line 1: **praise:** Thanks for the fixes here :)

vadl-lsp/main/vadl/lsp/AstFinderByPosition.java
```diff
@@ -109,6 +109,23 @@ public abstract class AstFinderByPosition<N extends Node> extends RecursiveAstVi
...
+          vadl.utils.RopeList<SourceLocation.DirectLocation> expandedFrom
+        )
+      ) {
+        var outerLocation = expandedFrom.toList().getLast();
+        if (!primaryLocation.begin().isWithin(outerLocation)
+            || !primaryLocation.end().isWithin(outerLocation)) {
+          return false;
+        }
+      }
+
```
flofriday, line 107: **question:** Can you explain when the primary location would be contained within the outer location?

In my mind that should never be the case, so maybe can you provide an example here?

Because in my intuitive opinion the code should just look like:

```java
if (n.location() instanceof SourceLocation.ExpandedLocation) {
    return false;
}
```

But in general we said we're not going to ignore the definitions from macro definitions, but instead collect all the possible types and display them, like mentioned in this issue: https://github.com/OpenVADL/openvadl/issues/1016#issuecomment-4646829968

flofriday:
Can you please explain the changes in SourceLocation?
I still don't quite get the kinds of problems the new implementation solves.
Also we should probably add tests for those cases in SourceLocationTest.

vadl-lsp/main/vadl/lsp/AstFinderByPosition.java
jaraonthe-dot-net: I want to keep all considerations from #1016 out-of-scope here. But thanks for pointing out the issue, I must have missed it.

> Can you explain when the primary location would be contained within the outer location?

As far as I've seen, this applies to entities that are part of **arguments** passed to the macro invocation.
Examples:
- `rv3264im.vadl:231` -> `addr` (example from description above). This variable name refers to the `let` definition in the model definition, but as we are hovering the model invocation, the type definitely applies here (as opposed to entities located within model definitions, where this is not so straight-forward).
- The last example above (`Arch64` on line 20), then go to the `pseudo instruction` on line 274 - here various type hovers work just fine (e.g. instruction parameters). The entire thing (and all child entities) have a primary location that is situated within the expanded location.

If you prefer, of course I can simplify the filter to ignore these cases too, but I saw no reason to do that, as I'm not aware that the shown type information is incorrect or otherwise problematic.

jaraonthe-dot-net:
> Can you please explain the changes in SourceLocation? I still don't quite get the kinds of problems the new implementation solves.

Firstly, `SourceLocation.join()` is relevant because there are AST entities that don't contain their own dedicated location field, instead compute their effective location from joining two child entities' locations.

My changes fix the behavior when two locations are joined that have a differing expansion stack. Part of the expansion stack (from the outermost model invocation inwards) may be the same, and in that case I need that part of the expansion stack to be kept in the result (to properly detect this entity is within a model definition/invocation in the first place). 
Instead, the previous implementation had the tendency to throw out the `expandedFrom` altogether, issuing a `DirectLocation` only (because it only includes the `expandedFrom` if the location stacks of both inputs are equal (or the innermost parts(!) are equal), and quite often (if not always) that is not the case).

This affects detection of locations that span from model definition to model invocation (see examples in description above): an entity joins the locations of two children together where one is part of the model definition, and the other is part of a model invocation argument, creating these weird `primaryLocation`s, but by dropping the expansion stack, it is very difficult to figure this out (and effectively ignore them).

Further example: `rv3264im.vadl:495`:
`let MinInt = ((1 as Regs) << ((match : Id ($opTy = SIntW => WLen; _  => MLen)) - 1)) as $opTy in`
The `match` statement creates one more entry in the expansion stack (for the entry that is the result of this `match`), which is then joined with the following `1` literal to make up the location of the subtraction operation (i.e. location with 3 `expandedFrom` locations joined with another that has only 2 (see screenshot)). The resulting mangled location has a weird `primaryLocation` and no `expandedFrom` (spanning yet again from within the model definition to within the model invocation).

If you need more convincing, we should probably discuss this in a meeting.

**EDIT**:
Two Screenshots illustrating the last example - this is without my changes to `SourceLocation`. `BinExpr` joins its left and right operand locations together, the resulting location (at the very bottom) does not contain any `expandedFrom` anymore, thus not being recognized as being part of a model invocation, leading to the bogus hover.
Without fixing SourceLocation Screenshot_20260811_225830
Without fixing SourceLocation Screenshot_20260811_230353

jaraonthe-dot-net:
@flofriday I've added your suggestions (see Fixup commit). Please take another look.

flofriday:
Now that I've a better tool to visualize/debug sourcelocations was the intent of this PR to fix cases like this (from the old version where it clearly is broken):

Screenshot 2026-08-19 at 20 24 58

Any other cases this intends to fix?

jaraonthe-dot-net:
> [...] was the intent of this PR to fix cases like this [...]

Without looking at the SourceLocations involved I am not sure why this example behaves like that (either the `expandedFrom` stacks diverge -> `join()`; or some location is not properly set by `MacroExpander`). But yeah, locations like that are what I want to filter out.

> Any other cases this intends to fix?

I'm trying to fix these issues:
- SourceLocation spanning from model definition to model invocation (some of those cannot be helped (i.e. an entity that joins two children where one is part of the definition and the other an invocation argument), thus the explicit filtering of these). (Your example falls into this category)
- SourceLocations that should have an `expandedFrom` stack but don't (because `join()` drops it). This is a very big problem because it makes it impossible to detect that a model is being used in the first place.
- The outermost `expandedFrom` becoming the effective location (my last example in the description).

vadl-lsp/main/vadl/lsp/AstFinderByPosition.java
flofriday: Yeah that makes sense, nice catch.

vadl/main/vadl/utils/SourceLocation.java
```diff
@@ -161,47 +161,42 @@ default SourceLocation join(SourceLocation other) {
...
+
+        return SourceLocation.of(firstLocation.path(), begin, end, expandedFrom);
       }
     }
 
-    if (!Objects.equals(firstLocation.path(), secondLocation.path())) {
+    // All minSize outer layers are the same for both locations; let's return them
+    if (!Objects.equals(thisStack.get(minSize - 1).path(), otherStack.get(minSize - 1).path())) {
       throw new IllegalArgumentException("Cannot join source locations from different files.");
     }
```
flofriday: **question:** Is this condition necessary, doesn't the loop already check if for all i < minSize?

flofriday:
I've went over the fix again now (with the Advanced AST Dump merged in #1067) and your implementation seems to be quite correct. I cannot come up with any counter-example and all the edge cases I tested are now correct or at least not more broken than before, well done ✨

However, I want to have more and easier to read tests for example this code is a minimal example of a previously really wrongly reported error message. That would make a nice addition to frontend  snapshottests (probably in the typechecker).
```
// Before #1052 this code reported a wrong source location.
model abc() : Ex = { 0 }

model xyz(): Defs = {
  constant flo = (5 + $abc) / $abc
}

$xyz()
```

On master this reports as:
```
error: Constant value required
     ╭── file:///Users/flo/src/open-vadl/example.vadl:5:19
     │
   5 │>   constant flo = (5 + $abc) / $abc
   6 │> }
   7 │> 
   8 │> $xyz()
     │    Division by zero cannot be computed.
     │
```

which is not correct and also dropped the expanded-from stack.
But with your change it reports only the division as expected 🎉


**Update:** While playing around I discovered another issue in the model expander which I wrote down here: #1075
If you want you can also try to fix it in this PR or in another one.

jaraonthe-dot-net:
@flofriday I've added the snapshot test you wanted me to add. Anything else before we can finish this PR?

flofriday:
Yes, there is still an unresolved conversation about a condition.

jaraonthe-dot-net:
> Yes, there is still an unresolved conversation about a condition.

I don't see any unresolved conversations on this PR. What are you referring to?

flofriday:
You're right I forgot to publish them, my bad sorry.

vadl/main/vadl/utils/SourceLocation.java
jaraonthe-dot-net: You're right, it's not necessary. Removed now.

flofriday (approved):
LGTM

PR #1054, opened 2026-08-11
frontend: Add float built-ins

BadGraphixD:
Adds built-in float functions necessary to fully specify the RISC-V F and D extension:
- arithmetic
- comparison
- conversion (float-float, float-int and int-float)
- classification

Also adds an example implementation for a built-in (VADL::adds) which also returns a float status. Since these variants are not required for RISC-V, I have not implemented them (they will be required ISAs which have trapping float exceptions).

### DIFF
--- a/vadl/main/vadl/types/BuiltInTable.java
+++ b/vadl/main/vadl/types/BuiltInTable.java
@@ -1040,6 +1040,317 @@ public class BuiltInTable {
           .build();
 
 
+  ///// FLOAT ARITHMETIC //////
+
+  /**
+   * {@code function fsqrt::< t : FloatType >( a : Bits<t.size>, rm : Bits<3> ) -> Bits<t.size> }
+   */
+  public static final BuiltIn FSQRT =
+      func("VADL::fsqrt",
+          Type.relation(List.of(BitsType.class, BitsType.class),
+              List.of(FloatType.class), BitsType.class))
+          .takesFloatArgsAndFrm(1)
+          .returnsFirstFloatType()
+          .build();
+
+  /**
+   * {@code function fadd::< t : FloatType >( a : Bits<t.size>, b : Bits<t.size>, rm : Bits<3> ) ->
+   *   Bits<t.size> }
+   */
+  public static final BuiltIn FADD =
+      func("VADL::fadd",
+          Type.relation(List.of(BitsType.class, BitsType.class, BitsType.class),
+              List.of(FloatType.class), BitsType.class))
+          .takesFloatArgsAndFrm(2)
+          .returnsFirstFloatType()
+          .build();
+
+  /**
+   * {@code function fadds::< t : FloatType >( a : Bits<t.size>, b : Bits<t.size>, rm : Bits<3> ) ->
+   *   ( Bits<t.size>, FloatStatus ) }
+   */
+  public static final BuiltIn FADDS =
+      func("VADL::fadds",
+          Type.relation(List.of(BitsType.class, BitsType.class, BitsType.class),
+              List.of(FloatType.class), StructType.class))
+          .takesFloatArgsAndFrm(2)
+          // TODO: returns function
+          .returnsFirstFloatType()
+          .build();
+
+  /**
+   * {@code function fsub::< t : FloatType >( a : Bits<t.size>, b : Bits<t.size>, rm : Bits<3> ) ->
+   *   Bits<t.size> }
+   */
+  public static final BuiltIn FSUB =
+      func("VADL::fsub",
+          Type.relation(List.of(BitsType.class, BitsType.class, BitsType.class),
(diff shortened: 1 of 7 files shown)


vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -1040,6 +1040,330 @@ public class BuiltInTable {
...
+   * {@code function fadd< t : FloatType >( a : Bits<t.size>, b : Bits<t.size>, rm : Bits<3> ) ->
+   *   Bits<t.size> }
+   */
+  public static final BuiltIn FADD =
+      func("VADL::fadd",
+          Type.relation(List.of(BitsType.class, BitsType.class, BitsType.class),
+              List.of(FloatType.class), BitsType.class))
+          .takesFloatArgsAndFrm(2)
+          .returnsFirstBitWidth(BitsType.class)
+          .returnsFirstFloatType()
```
Jozott00: Uses two different `returns` builder methods.

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -1040,6 +1040,330 @@ public class BuiltInTable {
...
+  // TODO: I think we want a status variant for float built-ins, similar to other built-ins.
+  //       But how to handle these for ISS?
+  //       When doing 64- or even 128-bit ops, returning the status from helpers is afaik not
+  //       possible. So we'd need to store them in a reg or an env variable and load them later.
+  //       But we have to make sure that other float ops are not scheduled in between!
+  /**
+   * {@code function fadds< t : FloatType >( a : Bits<t.size>, b : Bits<t.size>, rm : Bits<3> ) ->
+   *   ( Bits<t.size>, FloatStatus ) }
+   */
+  public static final BuiltIn FADDS =
```
Jozott00: Yes, I think the easiest way would be to store it in some "return" properties in the CPU state. The return property is then loaded into a TCG variable directly after the float built-in call. So the ISS always emits a status float built-in call together with a respective CPU env load operation. 

What do you think?

(note that we use a similar approach for exception arguments, which are also stored in the CPU state before calling the exception helper)

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -1040,6 +1040,330 @@ public class BuiltInTable {
...
+          .takesFloatArgs(2)
+          .returns(Type.bool())
+          .build();
+
+  ///// FLOAT CONVERSION //////
+
+  // TODO: currently the system allows conversion between all float types. We need to restrict
+  //       that to a set of supported conversions (currently ieee 32->64 and 64->32). Should this
+  //       happen during type-checking or in the backend? Different backends might be at different
+  //       stages of implementation and may not support everything other backends do...
```
Jozott00: I think language level conversion restrictions should be handled by the frontend, while backends should throw diagnostics if they don't yet support a valid conversion.

For the restriction check we could just add some kind of check method to the `BuiltIn` interface which is called by the frontend, together with the computed return type, which allows the built-in to throw diagnostics for such restrictions.

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -1040,6 +1040,330 @@ public class BuiltInTable {
...
+   */
+  public static final BuiltIn FCVT =
+      func("VADL::fcvt",
+          Type.relation(List.of(BitsType.class, BitsType.class),
+              List.of(FloatType.class, FloatType.class), BitsType.class))
+          .takesFloatArgsAndFrm(1)
+          .returnsSecondFloatType()
+          .build();
+
+  // TODO: Integer results have to be truncated by the IssNormalizationPass
```
Jozott00: This should be added to the `IssNormalizationPass`.

Jozott00 (changes requested):
Looks good to me! Just a couple minor things.

Please also add the built-ins to the documentation.

vadl/main/vadl/types/BuiltInTable.java
BadGraphixD: On second thought, I think the truncation should happen in the helper call. The size of the returned integer also affects when float exceptions are thrown, which means the helper must be aware of the returned size anyway.

vadl/main/vadl/types/BuiltInTable.java
BadGraphixD: I think that will work. I will add that in a separate PR.

BadGraphixD: see #1058

vadl/main/vadl/types/BuiltInTable.java
BadGraphixD: I think that will work. See #1059

Jozott00 (approved):
LGTM

PR #1056, opened 2026-08-12
lsp: Fix targetSelectionRange

jaraonthe-dot-net:
- Bug: Definition targetSelectionRange may lay outside targetRange (LSP specs don't allow this)

- This also meant that VSCode refused to jump to definition.
Example: `rv3264im.vadl:209`: `ADD` now jumps to the entire `instruction` definition inside the `RtypeInstr` model (cannot jump to identifier as that points back to `ADD` on line 209).

(EDIT: Correct line number)


flofriday (commented):
Looks good.

Can you add a minimal test for it with an example that didn't work before but does now.

Also I think you might have looked into an old version of rv3264 because the lines you mention don't contain any `ADD`.

jaraonthe-dot-net:
> Can you add a minimal test for it with an example that didn't work before but does now.

Done.

> Also I think you might have looked into an old version of rv3264 because the lines you mention don't contain any `ADD`.

Ah, my bad, that was a typo. I meant line **209**.

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.GotoDefinitionTest/modelArgument.snapshot
```diff
@@ -0,0 +1,76 @@
...
+      format Simple : Bits<1> = {b [0]}
+      model TestModel (name : Id, val : Bin) : IsaDefs = {
+        <TARGET SELECTION RANGE><TARGET RANGE>instruction $name : Simple = {}</TARGET SELECTION RANGE></TARGET RANGE>
+        encoding $name = { b = $val }
+        assembly $name = (mnemonic)
+      }
+    
+      $TestModel(<ORIGIN SELECTION RANGE>ZERO</ORIGIN SELECTION RANGE> ; 0b0)
+      $TestModel(ONE ; 0b1)
+    }
```
flofriday: **thought:** Looking at this, I don't think this is what a user would expect to be happening.

@Jozott00, @rascmatt I think you wrote more code but would you expect this behaivor?
For explaination someone CTRL-Clicked the `origin selection range` here and the LSP takes you to `target selection range` and highlights briefly  the `target range`.

Jozott00: Hmm, I think this is fine. In most cases I would like to go to the respective instruction definition logic. Technically, there is an ambiguity between the all occurrences of `$name`.

I guess in the future, we could have a dropdown, where the user can select between all `$name` instances, like it is the case for usages (if the LSP protocol itself supports this). So essentially instead of having a goto definition, it is a jump to a selected usage of `$name`.

The only thing I'm not sure about is whether we should highlight the whole `instruction` definition or just the instruction definition's `$name`. The latter one would align more with the just described ultimate goal.

jaraonthe-dot-net: In the weekly meeting today we discussed that it is not that clear what the best entity to jump to is. Maybe from the model invocation argument we should jump to the corresponding parameter definition in the model definition head (but that is technically not possible, as the AST does not contain the model invocation); or maybe it would be best to jump to the model definition as a whole (but that would be out-of-scope here).

We decided to keep this issue simple: Let's not offer a Goto Definition in this case. That change will still fix the non-compliant lsp behavior.

flofriday (approved):
LGTM

PR #1057, opened 2026-08-13
lsp: Hover expanded model invocations

jaraonthe-dot-net:
Hover shows what a model invocation expands to.

Screenshot_20260813_044154

Note that syntax highlighting of the hover snippet is not perfect because in our TextMate grammar many keywords are only allowed in specific contexts.

Based on unmerged work in #1052. I will remove those commits once that PR has been merged.

(EDIT: Correct based-on PR number)


flofriday:
This is pretty awesome 💪

Just a quick question, have you tried it on the really huge models yet?
Do the IDE's shrink the popup then automatically or do you think we should do something here to avoid beeing too distracting?

jaraonthe-dot-net:
> Do the IDE's shrink the popup then automatically or do you think we should do something here to avoid beeing too distracting?

Both VSCode and IntelliJ limit the size of the popup, adding scroll bars if necessary:
Screenshot_20260814_000739
Screenshot_20260814_001000

The hover is not shown immediately, but only after staying on the same position for a while, and the editors get rid of the popup if the user does something else (e.g. start typing).

flofriday:
Does this PR have to be based on #1065? If you drop the commits from that PR, I'm ready to merge this as is.

vadl-lsp/main/vadl/lsp/AstFinder.java
```diff
@@ -0,0 +1,209 @@
...
+   * given source code position.
+   *
+   * @param path The source code file to search in
+   * @param position The position to search for (within the file identified by {@code path}
+   * @return List of all found nodes. May be empty
+   */
+  static List<Node> findExpandedNodes(
+      Ast ast, Path path, SourceLocation.Position position) {
+
+    var visitor = new AstFinder<>(ast, (n) -> {
```
flofriday: **ntipick:** Let's use a more descriptive name here than n.

jaraonthe-dot-net:
> Does this PR have to be based on #1065? If you drop the commits from that PR, I'm ready to merge this as is.

I misspoke, it is based on #1052.

jaraonthe-dot-net:
@flofriday #1052 has been merged and removed from this pr. Should be fine to merge now.

vadl-lsp/test/resources/snapshots/vadl.lsp.integration.HoverTest/modelInvocation.snapshot
```diff
@@ -0,0 +1,58 @@
...
+    }
+  ================
+
+- returned Hover: 
+  Hover [
+    contents = Either [
+      left = null
+      right = MarkupContent [
+      kind = "markdown"
+      value = "Expanded to:  \n```vadl\ninstruction ONE : Simple = {\n}\nencoding ONE =\n{ b = 0b1 as Bits<1>\n}\nassembly ONE = (mnemonic)\n```"
```
flofriday: **super nitpick:** Great that this is in tests but now that the hover content get's more complicated it might be a good idea to render it in the tests in a way that is closer to how it's rendered in the editor and easier to verify at a glance.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -233,34 +233,64 @@ public CompletableFuture<Hover> hover(HoverParams params) {
...
+          }
+        }
 
-      if (node == null) {
-        return hoverResult(null, null);
-      }
+        {
+          // Show expanded code (for model invocations)
+          var nodes = AstFinder.findExpandedNodes(ast, path, position);
+          if (!nodes.isEmpty()) {
```
flofriday: **suggestion:** I think this code (and some other places) would be nicer if you inverted the condition and added an early return.

**Instead of:**

```java
if (cond) {
  // Many lines of code
  return result
}
return fallback;
```

**Return early:**
```java
if (!cond) {
  return fallback;
}

// Many lines of code
return result;
```

This code is often preferred because it puts the condition and returned value next to each other and it avoid deeply nested code.
If you like a deep dive from someone with great presentation skills I can recommend [Why You Shouldn't Nest Your Code](https://www.youtube.com/watch?v=CFRhGnuXG-4&pp=ygUSSSdtIGEgbmV2ZXIgbmVzdGVy)


**On Logs:**
One problem you mentioned before is that you often log something before you return. And then you'd have to write multiple log statements. I think that that isn't too bad but I can see how it bekomes easier to miss a branch.

But that is quite simple solvable with another layer of functions. 
So you only log in the top-level function the LSP4J framework calls. That way it is also clearer who has the responsibility to log and then the actual implementation that get's the result doesn't have to think about this anymore.

```java
  @Override
  public CompletableFuture<Hover> hover(HoverParams params) {
    log.debug(">> hover: {}", params);
    var result = hoverImpl(params);    // Real implementation now in hoverImpl.
    log.debug("<<- hover: {}", result);
    return result;
}
```

Or if you want to get fancy you could also get creative with higher order functions:

```java
  @Override
  public CompletableFuture<Hover> hover(HoverParams params) {
   return withLogging("hover", () -> {hoverImpl(params)})
}
```

vadl-lsp/main/vadl/lsp/AstFinder.java
```diff
@@ -0,0 +1,209 @@
...
+import vadl.ast.Ast;
+import vadl.ast.nodes.Definition;
+import vadl.ast.nodes.Expr;
+import vadl.ast.nodes.Node;
+import vadl.ast.nodes.RecursiveAstVisitor;
+import vadl.ast.nodes.Statement;
+import vadl.utils.SourceLocation;
+
+/**
+ * Finds one or several AST nodes based on some selection criteria (which can be quite complex).
```
flofriday: **nitpick:** I think we can find better wording like _"... based on an arbritary caller provided selection criteria"_.

vadl-lsp/main/vadl/lsp/AstFinder.java
```diff
@@ -0,0 +1,209 @@
...
+   * @return List of all found nodes. May be empty
+   */
+  static List<Node> findExpandedNodes(
+      Ast ast, Path path, SourceLocation.Position position) {
+
+    var visitor = new AstFinder<>(ast, (n) -> {
+      // Before
+      if (!(n.location() instanceof SourceLocation.ExpandedLocation expandedLocation)) {
+        return Choice.STOP_IF_SOME_FOUND;
+      }
```
flofriday: **suggestion:** Let's add a comment here why this is the stop condition, because it isn't clear at a glance.

vadl-lsp/main/vadl/lsp/AstFinder.java
```diff
@@ -0,0 +1,209 @@
...
+
+      var outermostLocation = expandedLocation.expandedFromStack().getLast();
+      if (path.equals(outermostLocation.path()) && position.isWithin(outermostLocation)) {
+        return Choice.SELECT_AND_SKIP_CHILDREN;
+      }
+      return Choice.STOP_IF_SOME_FOUND;
```
flofriday, line 60: **suggestion:** Let's also explain here why we are stopping if one was already found.

vadl-lsp/main/vadl/lsp/AstFinder.java
```diff
@@ -0,0 +1,209 @@
...
+      // Nothing
+    }
+    return selectedNodes;
+  }
+
+  private void beforeNode(Node node) {
+    depth++;
+    if (depth > ignoreUntilDepth) {
+      // Skipping children by ignoring them - but this still traverses through all these nodes.
+      // To skip traversing them altogether we'd have to override all(!) RecursiveAstVisitor.visit()
```
flofriday: **suggestion:** I think this got recently fixed with the `DefaultAstVisitor` which overrides all visit methods and provides a simple API to implement a default handling for all nodes. Actually skipping all the children would be quite useful. 
Can you try to implement it upon DefaultAstVisitor and if it get's too complex let's defer it for another PR.

flofriday (commented):
Awesome feature,  just some small things to tackle.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -233,34 +233,64 @@ public CompletableFuture<Hover> hover(HoverParams params) {
...
+            var range = document.calculateUtf16Range(
+                nodes.getFirst().location().expandedFromStack().getLast());
+            var prettyPrinted = new ArrayList<String>(nodes.size());
+            for (var node : nodes) {
+              var builder = new StringBuilder();
+              node.prettyPrint(0, builder);
+              prettyPrinted.add(builder.toString().trim());
+            }
+            return hoverResult("Expanded to:", String.join("\n", prettyPrinted), range);
```
flofriday: **thought:** Let's explicitly name that this is about model here, something like "This model expands to:".

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
jaraonthe-dot-net: I don't agree that the code in `hover()` would look nicer with this change.
There are two different features integrated here, one to provide type information and the other to show expanded code. The logical way to read the method, as it is right now, is that feature A is attempted, then feature B is attempted, then the fallback occurs. Pushing the fallback case to the front of feature B destroys this structure.
Or in other words: The condition belongs to feature B, not to the fallback (and the fallback's condition isn't simply the logical negation).

But I'll improve `hoverResult()` according to your suggestion.

vadl-lsp/main/vadl/lsp/AstFinder.java
jaraonthe-dot-net: FYI: I've tried your suggestion and it turns out that no parent class is needed for that approach (they don't simplify the implementation of `AstFinder` at all).

jaraonthe-dot-net:
@flofriday I've applied your suggestions (except where noted otherwise).

vadl-lsp/main/vadl/lsp/AstFinder.java
flofriday: I've looked over the new implementation and it does look neater than before, and it now actually skips all the children. 👍

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
flofriday: I see, to be fair in the snapshot of the review it looked like they were both called in lambdas, but the wrapping curly braces, are just there to provide a new scope.
I think this is a good point to simplify the code and put each feature into it's own function and keep the `hover()` function as the coordinator, managing the order and the fallback.

jaraonthe-dot-net: Done. Please take another look.

flofriday (approved):

PR #1062, opened 2026-08-17
riscv: add F and D extension ISA specs

BadGraphixD:
Adds the RISC-V F and D float extension specs. Contains `ISA` and `processor` definition.


AndreasKrall (approved):
looks good to me. Some identifiers are quite short, e.g. for some enumerations, could be more verbose. Please test with a subset of the RISC-V test suite the correctness. I assume some tests regarding NaNs must me moved into the future.

PR #1063, opened 2026-08-17
frontend: Add built-in check function

### LINKED ISSUE #1058: frontend: Add validity checks for built-ins
Some built-ins, like float-to-float conversion `fcvt` currently allow all input and output sizes. But conversions like 32-bit to 32-bit make no sense and should be forbidden. We should allow built-ins to throw diagnostics for that.

BadGraphixD:
Adds an optional check function to built-ins, which the type-checker calls. This check function can return a `Diagnostic`, which is thrown by the type-checker. This is currently only used to verify that the two float-types passed to `VADL::fcvt` do not have the same encoding, and that converting between float and int only involves 32- or 64-bit integers.

This check function will also be used after the transition to type parameters.

Also changes how float-types are stored in the AST. Previously, only the bit-size was stored in the AST. Now, the whole `FloatEncoding` is stored there.

Closes #1058

### DIFF
--- a/vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidFloatConvertBuiltinParams.vadl
+++ b/vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidFloatConvertBuiltinParams.vadl
@@ -1,0 +1,157 @@
+// INCLUDE-AST-DUMP
+
+[ IEEE : 32 ]
+float-type ieee32
+
+function fcvt(a: Bits<32>) -> Bits<32> = VADL::fcvt::<ieee32, ieee32>(a, 0 as Bits<3>)
+function fcvtfs(a: Bits<32>) -> SInt<7> = VADL::fcvtfs::<ieee32, 7>(a, 0 as Bits<3>)
+function fcvtfu(a: Bits<32>) -> UInt<7> = VADL::fcvtfu::<ieee32, 7>(a, 0 as Bits<3>)
+function fcvtsf(a: SInt<7>) -> Bits<32> = VADL::fcvtsf::<ieee32>(a, 0 as Bits<3>)
+function fcvtuf(a: SInt<7>) -> Bits<32> = VADL::fcvtuf::<ieee32>(a, 0 as Bits<3>)
+
+
+//  Reported Diagnostics:
+//
+//  error: Cannot convert to same float encoding
+//       ╭── test/resources/frontend-snapshots/typechecker/invalidFloatConvertBuiltinParams.vadl:6:42
+//       │
+//     6 │ function fcvt(a: Bits<32>) -> Bits<32> = VADL::fcvt::<ieee32, ieee32>(a, 0 as Bits<3>)
+//       │                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//       │
+//
+//  error: Can only convert to 32 or 64 bit integers
+//       ╭── test/resources/frontend-snapshots/typechecker/invalidFloatConvertBuiltinParams.vadl:7:43
+//       │
+//     7 │ function fcvtfs(a: Bits<32>) -> SInt<7> = VADL::fcvtfs::<ieee32, 7>(a, 0 as Bits<3>)
+//       │                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//       │
+//
+//  error: Can only convert to 32 or 64 bit integers
+//       ╭── test/resources/frontend-snapshots/typechecker/invalidFloatConvertBuiltinParams.vadl:8:43
+//       │
+//     8 │ function fcvtfu(a: Bits<32>) -> UInt<7> = VADL::fcvtfu::<ieee32, 7>(a, 0 as Bits<3>)
+//       │                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//       │
+//
+//  error: Can only convert from 32 or 64 bit integers
+//       ╭── test/resources/frontend-snapshots/typechecker/invalidFloatConvertBuiltinParams.vadl:9:43
+//       │
+//     9 │ function fcvtsf(a: SInt<7>) -> Bits<32> = VADL::fcvtsf::<ieee32>(a, 0 as Bits<3>)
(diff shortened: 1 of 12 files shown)


flofriday:
The quality of typechecker builitin diagnostics is a real problem but I don't love the approach here.

For one `check` and `takes` are two very generic sounding functions and even if the docs explain how they need to be used in tandem, nobody reads every function doc in a codereview. Also I mentioned them now in the wrong order but their names don't indicate any order. Making their names more complex (and ugly) to reflect their semantics would already help something (like `takesConstArgsAndArgsButNotCheckTheirValidityToEachOther` and `checkConstArgsAndArgsAndTheirValidityToEachOther`)

The bigger issue I see is that I don't really get why should have two functions in the first place. Right now it's really hard to tell why we would have two steps, and (as the example show above) for me how to name them. I think we can merge both functions, and always return a diagnostic.

This is a problem we had for a longer time, thus in the Typechecker this longstanding TODO stands to remind us:
```java
    if (!builtIn.takes(constArgs, argTypes)) {
      // FIXME: Further improve these error messages.
      var calledParamsAndTypes = String.join(", ", Stream.concat(
          constArgs.stream().map(Constant::toString),
          argTypes.stream().map(Type::toString)
      ).toList());
      var areSomeConst = originalArgTypes.stream().anyMatch(ConstantType.class::isInstance);
      addErrorAndStopChecking(
          error("Type Mismatch", location)
              .locationDescription(location, "The builtin has the signature `%s` but got `%s`.",
                  builtIn.signature(), calledParamsAndTypes)
              .applyIf(areSomeConst, b -> b.locationHelp(location,
                  "Try casting some of the constant arguments to explicit types."))
              .build());
    }
```

I'd suggest to merge `take` and `check` with the two only differing in the type they return. So for everyone who already uses `take` the API will still exist but under the hood there is just a single implementation that verifies the constArgs and Args, and a single call to either is enough to determin if the builtin call is semantically correct.f

All the newly added diagnostics for this can be supper generic and improved later but we should get the infrastructure right.

If this explodes in work, please DM me and we figure something out how to split this work across the two of us or defer it for later.

Sorry for the long reply. 😅

BadGraphixD:
@flofriday So you want the API to contain a `takes` and a `check` method, which both check the same thing and use the same functionality underneath, but `takes` simply returns a boolean result, while `check` creates diagnostics right? So the type-checker would only use `check`, and the `ViamVerificationPass` could then only use `takes` right?

I am all for this idea, but I'd like to implement it in a separate PR after #1074 is merged, since doing it here would require me to rewrite the behaviour twice (here and in #1074).

flofriday:
Yes exactly. 😅

Yeah fair, that would require some git shenanigains, to still be able to rebase the other branch on this one (no history rewrites).
So yeah I'm fine with doing that after #1047 is merged.

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidFloatConvertBuiltinParams.vadl
flofriday, line 1: **praise:** Well done test 🐿️🎉

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -2419,16 +2442,61 @@ public BuiltInBuilder returnsFromFirstTwoFloatSizes(
...
+        if (floatTypeEncoding(constArgs.get(0)) == floatTypeEncoding(constArgs.get(1))) {
+          return Optional.of(
+              Diagnostic.error("Cannot convert to same float encoding", location).build()
+          );
+        }
+        return Optional.empty();
+      });
+    }
+
+    public BuiltInBuilder checkConvertingTo3264Int() {
```
flofriday, line 2476: **minor:** Let's smuggle a _float_  in the name here somewhere so it's clear what we are convertig here.

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -2419,16 +2442,61 @@ public BuiltInBuilder returnsFromFirstTwoFloatSizes(
...
+        if (intBitSize != 32 && intBitSize != 64) {
+          return Optional.of(
+              Diagnostic.error("Can only convert to 32 or 64 bit integers", location).build()
+          );
+        }
+        return Optional.empty();
+      });
+    }
+
+    public BuiltInBuilder checkConvertingFrom3264Int() {
```
flofriday, line 2488: **minor:** Same float smuggeling opperation here

vadl/main/vadl/types/FloatEncoding.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /**
+   * Returns the IEEE encoding for the given bit-size.
+   */
+  public static @Nullable FloatEncoding ieee(int size) {
+    return switch (size) {
+      case 32 -> IEEE32;
+      case 64 -> IEEE64;
+      default -> null;
```
flofriday: **question:** Sure this is better than loudly failing with a `IllegalStateException`?

flofriday (commented):
Almost done, just minor stuff 🎬🎉

vadl/main/vadl/types/BuiltInTable.java
BadGraphixD: In #1074 I am consolidating all `fcvt*` functions into a single builtin (that's possible due to the type parameters). Due to that I am also merging all the conversion checks into a single method called `checkFloatConversion()`.

vadl-frontend/test/resources/frontend-snapshots/typechecker/invalidFloatConvertBuiltinParams.vadl
BadGraphixD: Thanks ❤️

vadl/main/vadl/types/BuiltInTable.java
BadGraphixD: Same as above

vadl/main/vadl/types/FloatEncoding.java
BadGraphixD: I am returning `null` here so I can later check against that an throw a Diagnostic in AnnotationTable.java:343. But you're right, I'll add a separate method for checking whether the size is valid.

flofriday (approved):
LGTM 🐎🎉

flofriday (approved):

PR #1064, opened 2026-08-19
frontend: Add NewLabelStatement and LabelType

### LINKED ISSUE #148: lcb: Wrong label for `%pcrel_lo` in -fPIC

benjaminkasper99:
This PR implements `Solution 1` proposed in the comments of #148.

The introduced `NewLabelStatement` is lowered to the existing VIAM `NewLabelNode`, which is used by the LCB to emit assembly labels when expanding instruction sequences. This allows us to finally remove the `HardcodeLGALabelPass`.

Note that the spec changes of this PR break the `rv32im` assembler and linker in the sense that they produce incorrect binaries. Because of this I disabled two CodeEmitter tests. The issue will be fixed and the tests re-enabled with the follow-up PR #1071.


vadl-frontend/main/vadl/ast/nodes/InstructionSequenceStatement.java
```diff
@@ -0,0 +1,24 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+package vadl.ast.nodes;
+
+/**
+ * Supertype for all statements that can appear in instruction sequences
+ * like {@link PseudoInstructionDefinition} and {@link AbiSequenceDefinition}.
+ */
+public abstract class InstructionSequenceStatement extends Statement {
+}
```
flofriday: **question:** Why is it important to have this empty shared super class? What was hard to do if they just shared Statement and from context it was clear which classes appear there?

vadl-frontend/main/vadl/ast/nodes/NewLabelStatement.java
```diff
@@ -0,0 +1,72 @@
...
+package vadl.ast.nodes;
+
+import java.util.Objects;
+import vadl.javaannotations.ast.Child;
+import vadl.utils.SourceLocation;
+
+/**
+ * A statement that defines a new label in an instruction sequence.
+ */
+public class NewLabelStatement extends InstructionSequenceStatement {
```
flofriday: **praise:** Looks good 👍

vadl-frontend/main/vadl/ast/nodes/PseudoInstructionDefinition.java
```diff
@@ -36,9 +36,9 @@ public class PseudoInstructionDefinition extends InstructionSequenceDefinition
   public AssemblyDefinition assemblyDefinition;
 
   public PseudoInstructionDefinition(IdentifierOrPlaceholder identifier, PseudoInstrKind kind,
```
flofriday, line 38: **nitpick:** I don't really see the point of `InstructionSequenceStatement`, they could have just been Statement here and we trust the Parser to only provide the couple of statements that are allowed.
But correct me if I'm wrong 😅

vadl-frontend/main/vadl/ast/BehaviorLowering.java
```diff
@@ -1068,6 +1071,13 @@ private ExpressionNode visitIdentifiable(Expr expr) {
           getViamType(expr.type()));
     }
 
+    // NewLabel statement in an instruction sequence
+    if (computedTarget instanceof NewLabelStatement) {
+      var labelNode = currentGraph.getNodes(NewLabelNode.class)
+          .filter(n -> n.labelNode().labelName().equals(innerName)).toList().getFirst();
+      return labelNode.labelNode();
```
flofriday: **minor:** I think we should  use `fetch` here.

It computes the label on demand (if it isn't computed yet, like when the label is defined after it's usage. And if it is already computed it uses a cache instead of traversing the graph.

This should look similar to the function call, so something like:
```
if (computedTarget instanceof NewLabelStatement labelStmt) {
      var labelNode = (NewLabelNode) fetch(labelStmt);
      return labelNode.labelNode();
}
```

vadl-frontend/main/vadl/ast/BehaviorLowering.java
```diff
@@ -2033,6 +2043,15 @@ private static Type forallIndexType(Type type) {
     return Type.bits(type.asDataType().bitWidth());
   }
 
+  @Override
+  public SubgraphContext visit(NewLabelStatement statement) {
+    var newLabelNode =
+        new NewLabelNode(new LabelNode(statement.labelId().name, statement.labelId().type()));
```
flofriday: **nitpick:** I know it won't make a difference here but for hygenic reasons maybe we should wrap the type here in `getViamType`.

vadl-frontend/main/vadl/ast/BehaviorLowering.java
flofriday: **issue:** The source-locations of this label will be wrong.

@Jozott00 can you verify this and maybe tell if there are other problems in the VIAM with this approach?

So everytime a label is used we return the same instance with `labelNode.labelNode()` and I don't think I've seen such a pattern in the VIAM yet.

One reason against it is that multiple labels usages, even if they point to the same label should have different locations (where they are used). So this code doesn't explicitly add a location but the behaivor lowering will add a best-efford location afterwards. Therefore the first usage of a label will modify the node and set a sourcelocation and the second usage will now already contain that location and not be overwritten. So all usages will point to the first usage.

vadl/main/vadl/pass/order/ViamPassOrder.java
```diff
@@ -87,7 +86,6 @@ public static PassOrder create(GeneralConfiguration configuration) throws IOExce
     order.add(new AlgebraicSimplificationPass(configuration));
     order.add(new BehaviorRewritePass(configuration));
     order.add(new InstructionResourceAccessAnalysisPass(configuration));
-    order.add(new HardcodeLGALabelPass(configuration));
```
flofriday, line 90: **nitpick:** Instead of removing the line I would have commented it out together and added a comment linking to the issue/pr that works on reenabling it.

vadl/main/vadl/types/LabelType.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public String name() {
+    return "LabelType";
+  }
+
+  @Override
+  public boolean isTrivialCastTo(Type other) {
+    return this == other || other instanceof DataType;
```
flofriday, line 37: **issue:** I don't think you can cast a Label to a UInt<4> or any other datatype.

vadl/main/vadl/types/LabelType.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return this == obj || obj instanceof DataType;
```
flofriday, line 42: **issue:** This also doesn't look right.

flofriday (commented):
Overall quite good, some issues need to be fixed but shouldn't be too much.

In the future, could you maybe turn off the automatic whole-file-formatting in your editor? The reformats were all fine so I left them now, but for one it makes reviewing harder because it's not initially clear why you even touched that code and secondly it clutters the git history.

I know it's a shame that our checkstyle rules aren't strict enough to not have catched them in the first place.

vadl-frontend/main/vadl/ast/nodes/InstructionSequenceStatement.java
benjaminkasper99: You are right, just using Statement also works. My intention was to make it explicit that only a certain subset of statements is allowed in InstructionSequences.

vadl-frontend/main/vadl/ast/nodes/PseudoInstructionDefinition.java
benjaminkasper99: see above

vadl/main/vadl/pass/order/ViamPassOrder.java
benjaminkasper99: I think it's fine to remove the pass. It will not be re-enabled, since the whole point of this current PR is to get rid of this hardcoded pass. This comment in the pass also references the issue, which will be closed by merging this PR:
>Temporary pass for inserting labels into a pseudo instruction.
> See {@code https://github.com/OpenVADL/openvadl/issues/148}.

vadl/main/vadl/types/LabelType.java
benjaminkasper99: I agree that this is weird, but I'm not quite sure how else to handle the LabelType.
When for example `pcrel_lo( hi_label )` is typechecked, I need `hi_label` to work as argument no matter what the actual declared type of the parameter of `pcrel_lo()` is.

In the TypeChecker the comparision is done via `equals` here:
```
private Expr tryWrapImplicitCast(Expr inner, Type to) {  
  if (inner.type == null) {  
    throw new IllegalStateException("The type of the inner expression must be known.");  
  }  
  var wrapped = wrapImplicitCast(inner, to);  
  if (!wrapped.type().equals(to)) {  
    addErrorAndStopChecking(typeMismatchError(inner, to, inner.type()));  
  }  
  return wrapped;  
}
```

vadl/main/vadl/types/LabelType.java
benjaminkasper99: Similar as below I need a LabelType argmuent to fit any parameter. `isTrivialToCastTo` is used in the `verifyState` of `FuncCallNode`:
```
for (int i = 0; i < args.size(); i++) {  
  var arg = args.get(i);  
  var param = params[i];  
  ensure(arg.type().isTrivialCastTo(param.type()),  
      "Argument does not match type of param %s, %s vs %s", param.simpleName(), param.type(),  
      arg.type());  
}
```

vadl-frontend/main/vadl/ast/BehaviorLowering.java
benjaminkasper99: I updated it to create a new `labelNode` for each usage.

vadl-frontend/main/vadl/ast/nodes/InstructionSequenceStatement.java
flofriday: For that use case a sealed abstract class would be a great fit, but since we don't take advantage of any features of this subset in the code I think we can get rid of it.

vadl/main/vadl/pass/order/ViamPassOrder.java
flofriday: Yess you are totally right, I confused this with tests somehow, sorry 😅

vadl/main/vadl/types/LabelType.java
flofriday: >  I need hi_label to work as argument no matter what the actual declared type of the parameter of pcrel_lo() is.

I'm sure I don't quite understand how labels work, but why wouldn't pcrel_lo expect a label type?

vadl-frontend/main/vadl/ast/BehaviorLowering.java
Jozott00: No, I think this should be fine. As all usages reference the same label, they should share the same dependency object.

vadl/main/vadl/viam/graph/dependency/LabelNode.java
```diff
@@ -26,18 +26,26 @@
  * {@link LabelNode}.
  */
 public class LabelNode extends ExpressionNode {
-  public LabelNode(Type type) {
+
+  private final String labelName;
```
Jozott00, line 33: The name should be annotated with `@DataValue` so the graph framework can distinct it from other labels with a different name.

vadl/main/vadl/types/LabelType.java
benjaminkasper99: A label is just a placeholder for value of a certain DataType, that is only finally resolved at link time. I think it makes sense to define relocations with the actual DataType and let the LabelType be compatibile with any DataType while typechecking.

If we were to the define relocations with a LabelType, we would have to handle operations where the operand is a LabelType. An example would be the addition here:
```
relocation pcrel_lo( symbol : Label ) -> SInt<12> = (symbol + 4) as SInt<12>
													 ^^^^^^^^^^ adding to a LabelType?
```


I also thought about just setting the type of a label so that it fits its usage. 
```
relocation pcrel_lo( symbol : Bits<32> ) -> SInt<12> = (symbol + 4) as SInt<12>

pseudo instruction CALL( symbol : Bits<32> ) =
{
    newlabel hi_label
    AUIPC{ rd = 1 as Bits5, imm = pcrel_hi( symbol ) }
    JALR{ rd = 1 as Bits5, rs1 = 1 as Bits5, imm = pcrel_lo( hi_label ) }
}
```
In this example we would somehow set the type of `hi_label` to `Bits<32>` before the expression `pcrel_lo( hi_label )` is typechecked.

flofriday (approved):
LGTM 🎂

flofriday (approved):

PR #1065, opened 2026-08-19
[#1061] Add utf-8 support for VADL source files

### LINKED ISSUE #1061: Source locations in error messages are wrong
See https://github.com/OpenVADL/openvadl/pull/1042/changes#r3748287616

rascmatt:
Fixes #1061 

With the introduction/use of the `∈` operator (#1042), the first unicode character, it has become apparent that the Scanner/Buffer did not correctly handle unicode.

The `Scanner.frame` in Coco/R (Java) already ships with an utf-8 buffer, however it is only conditionally selected, based on the File's utf-8 preamble (or 'byte order mark'/BOM).

Since Coco/R bootstraps itself, i.e. the vadl.ATG grammar is parsed using the same default `Scanner.frame` we were using until now, the symbols in the VADL grammar have not been interpreted correctly as unicode. This caused the scanner to interpret `∈` symbol as a 3-byte sequence, and Coco/R generated a scanner to detect those 3 bytes as an appropriate token (-kind) exactly. 

So, while we were able to detect the operator, the `length` of the token was `3`, which resulted in the source locations to be incorrectly shifted.

-- 

By adding the utf-8 preamble to the `vadl.ATG` (`ZWNBSP`, resp. `U+FEFF`) the Coco/R library now interprets the symbols in the grammar as utf-8 as well.

In addition, this PR now always interprets `*.vadl` source files in utf-8, regardless of whether they contain the BOM or not.

### DIFF
--- a/vadl-frontend/main/vadl/ast/Scanner.frame
+++ b/vadl-frontend/main/vadl/ast/Scanner.frame
@@ -310,7 +310,7 @@ public class Scanner {
 -->initialization
 	}
 	
-	public Scanner (String fileName) {
+	public Scanner(String fileName) {
 		buffer = new Buffer(fileName);
 		Init();
 	}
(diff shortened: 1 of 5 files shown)


flofriday (approved):
This is AWESOME 🎉🚀

flofriday (approved):

rascmatt:
Rebased and fixed 3 tests

flofriday (approved):

PR #1067, opened 2026-08-19
frontend: Add advanced AST dump

flofriday:
**Showcase:**

https://github.com/user-attachments/assets/9505de79-ba1a-4b3a-ac5d-bcc27ed1bcc7

Previously we only had text based dumps and while they worked, it often was hard to figure out which AST node corresponds to which construct in the language.

To pull this off I had to rewrite some parts of the AST Dumper and the templating setup.


rascmatt (approved):
Nice, looks very good to me :)

Jozott00 (changes requested):
I think backward referencing from VADL sources to AST would be good to have.

flofriday:
@Jozott00 that is a lovely idea, and if you wouldn't have already suggested it 5 times sitting right next to me right now I would seriously consider it 😜

Jozott00 (approved):
hhhmmmmm

PR #1071, opened 2026-08-24
linker: Add indirect relocations

benjaminkasper99:
Add the capability to generate indirect relocations to the OpenVADL linker.

An example of such a relocation is `pcrel_lo` of RISCV, where the relocation's type depends on the type of its paired relocation.

```
.Lpcrel_hi0:
   auipc   a0, %got_pcrel_hi(Coeff)
   ld      a0, %pcrel_lo(.Lpcrel_hi0)(a0)}
```
Here the `%pcrel_lo` relocation is paired with a global offset table (GOT) relocation, therefore `%pcrel_lo` is also GOT relative.

```
.Lpcrel_hi0:
   auipc   a0, %pcrel_hi(Coeff)
   ld      a0, %pcrel_lo(.Lpcrel_hi0)(a0)}
```
In this case `%pcrel_lo` is paired with a PC relative relocation, therefore  `%pcrel_lo` also is just PC relative.

Notable changes in this PR:
- Added the `paired` annotation to relocation definitions, to declare which relocations should behave as described above
- Added two LLD templates, in which the relocation indirection is implemented
- Moved pseudo expansion within the assembler from code emitter to parser, since access to a streamer is needed to emit labels. This also required to adjust the asm parser filecheck tests to check the expanded state of instructions.
- Added a Riscv64 LCB integration test (`LcbRiscv64SimulationTest`)
- Re-enabled two code emitter tests (disabled in #1064)
- Adjusted label names in LLVM IR filecheck tests

### DIFF
--- a/sys/risc-v/rv3264im.vadl
+++ b/sys/risc-v/rv3264im.vadl
@@ -296,7 +296,7 @@ $Arch3264 ( // ((((((((((((((((((((((((((((((((((((((((
   relocation pcrel_hi( symbol : Bits<32> ) -> UInt<20> = ( ( symbol + 0x800 as Bits<32> ) >> 12 ) as UInt<20>
   [relative]
   [syntax : "%{}({})"]
-  relocation pcrel_lo( symbol : Bits<32> ) -> SInt<12> = (symbol + 4) as SInt<12>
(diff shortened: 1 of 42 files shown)


Jozott00 (approved):
LGTM, although i just did a structural check rather than a semantic one

PR #1074, opened 2026-08-26
frontend: Change constant parameters to type parameters

BadGraphixD:
- Builtin functions can now take type parameters using the following syntax:
  - `VADL::builtin< first type param >< second type param >( normal parameters... )`
- Float types are now real types
  - They can be declared using `float-type`
  - `FloatType` is a subtype of `DataType`
  - All instances of `FloatType` are unique and contain a float format
  - All `FloatType` can only be explicitly cast from and to `BitsType`
    - Such casts DO NOT cause implicit sign/zero extension
- Adapts VIAM and ISS to the changes
- Adapts the RISC-V F/D spec to the changes
- Adapts the refmanual to the changes

**Note:** this branch is based on and contains a copy the commit of #1063. That PR should be reviewed and merged first.


vadl/main/vadl/types/FloatType.java
```diff
@@ -16,29 +16,68 @@
...
  *
- * <strong>Note:</strong> FloatType is NOT a {@link DataType}. It is NOT the type of float
- * expressions. Float expressions are {@link BitsType}. There exist no conversions between this
- * and any other type.
+ * // Example usage: converts from `Float` to `Double`
+ * VADL::fcvt<Double>(... as Float, ...)
+ * }</pre>
  */
-public class FloatType extends Type {
+public class FloatType extends BitsType {
```
flofriday: **thought:** I'm not totally certain that this inheritance model makes sense.

I _fear_ that possibly _somewhere_ in the code especially in the typechecker we might make checks based on is `instanceof` instead of checking the specific class (I found 23 such checks). In which case we now also accept the float type which other conditions have not expected.
Maybe these checks are fine. Honestly I'm not totally sure what to do here because intuetively it does make sense that it is a Bits type but maybe being more conservative and use a DataType here makes more sense.

@Jozott00 do you have any thoughts on this?

vadl-test/resources/cosim_scripts/riscv/compiler.py
```diff
@@ -50,9 +50,6 @@ def link(ld_cmd: str, linker_script: Path, obj_in: Path, elf_out: Path) -> None:
 def build_assembly(id: str, core: str) -> Path:
     asm_out = _tmp_file(id, f"asm-{id}.s")
 
-    # We load the tests into the RAM region at 0x40000000, because loading into the
-    # firmware region does not work, because aarch32/virt.vadl already has firmware.
-
```
flofriday, line 55: **question:** Is this simply no longer a problem?

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -1044,315 +1044,258 @@ public class BuiltInTable {
...
-          Type.relation(List.of(BitsType.class, BitsType.class, BitsType.class),
-              List.of(FloatType.class), StructType.class))
-          .takesFloatArgsAndFrm(2)
+          Type.relation(List.of(FloatType.class, FloatType.class, BitsType.class),
+              StructType.class))
+          .takesDefault()
+          .checkFrmAndFloatTypesEqual(2)
           // TODO: returns function
-          .returnsFirstFloatType()
+          .returnsFirstArgType()
```
flofriday, line 1080: **question:** Is this right because it should return a tuple, or am I missing something?

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -2416,85 +2322,109 @@ public BuiltInBuilder returnsFromDataTypes(Function<List<DataType>, Type> return
...
-      return floatTypeEncoding(arg).size;
+    public BuiltInBuilder checkFloatTypesEqual(int floatArgCount) {
+      return check((typeParams, argTypes, location) ->
+          checkFloatTypesEqualInternal(argTypes, floatArgCount, location));
     }
 
-    private int constInt(Constant arg) {
-      ensure(arg instanceof Constant.Value, "Expected a value type, but found %s", arg);
-      return ((Constant.Value) arg).intValue();
+    public BuiltInBuilder checkFrmAndFloatTypesEqual(int frmArgIdx) {
```
flofriday, line 2357: **minor:** Let's add a docstring here.

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -2416,85 +2322,109 @@ public BuiltInBuilder returnsFromDataTypes(Function<List<DataType>, Type> return
...
+      return check((typeParams, argTypes, location) ->
+          checkFrmInternal(argTypes.get(frmArgIdx), location).or(() ->
+              checkFloatTypesEqualInternal(argTypes, frmArgIdx, location)));
     }
 
-    public BuiltInBuilder check(
-        TriFunction<List<Constant>, List<Type>, WithLocation, Optional<Diagnostic>> checkFunction) {
-      this.checkFunction = checkFunction;
-      return this;
+    private Optional<Diagnostic> checkFrmInternal(Type type, WithLocation loc) {
```
flofriday, line 2368: **minor:** Also docstring.

vadl/main/vadl/types/BuiltInTable.java
```diff
@@ -2416,85 +2322,109 @@ public BuiltInBuilder returnsFromDataTypes(Function<List<DataType>, Type> return
...
-      return check((constArgs, argTypes, location) -> {
-        var intBitSize = argTypes.getFirst().asDataType().bitWidth();
-        if (intBitSize != 32 && intBitSize != 64) {
-          return Optional.of(
-              Diagnostic.error("Can only convert from 32 or 64 bit integers", location).build()
-          );
-        }
-        return Optional.empty();
-      });
+    private Optional<Diagnostic> checkFcvtIntegerTypeInternal(Type type, String fromTo,
```
flofriday, line 2434: **minor:** Also docstring.

sys/risc-v/rv64fd.vadl
flofriday, line 1: **question:** Do you think it makes sense to add this file to the `FrontendIntegrationTests`?

vadl-frontend/main/vadl/ast/ConstantEvaluator.java
```diff
@@ -485,7 +481,7 @@ public ConstantValue visit(CallIndexExpr expr) {
         throw new EvaluationError(
             "The constant evaluator cannot handle subcalls or indexing/slicing", expr);
       }
-      result = evalBuiltin(builtin, args.stream().map(this::eval).toList(), expr);
+      result = evalBuiltin(builtin, typeParams, args.stream().map(this::eval).toList(), expr);
```
flofriday, line 484: **question:** Can we actually ever eval anything that has non-empty typeParams?

vadl-frontend/main/vadl/ast/ParserUtils.java
```diff
@@ -1045,6 +1046,32 @@ static List<ExpandedSequenceCallExpr> expandSequenceCalls(Parser parser,
     return expandedCalls;
   }
 
+  static Expr createSymExprOrBinOp(IsId path, List<Expr> terms, boolean isLtOp,
```
flofriday, line 1063: **thought:** I'm fine with this, but can we add a comment explaining this hack and why we need it.

vadl-frontend/main/vadl/ast/TypeChecker.java
```diff
@@ -724,7 +725,7 @@ private static boolean canImplicitCast(Type from, Type to) {
         return availableWidth >= fromConstant.requiredBitWidth();
       }
 
-      if (to.getClass() == BitsType.class) {
+      if (to.getClass() == BitsType.class || to.getClass() == FloatType.class) {
```
flofriday: **question:** Does it make sense to cast a bitpattern implecitly to a float?
Because it will not have the same value as the bit pattern.

vadl-frontend/main/vadl/ast/TypeChecker.java
```diff
@@ -758,6 +759,13 @@ private static boolean canImplicitCast(Type from, Type to) {
       }
     }
 
+    // FloatType => Bits<n>
+    if (from.getClass() == FloatType.class) {
+      if (to.getClass() == BitsType.class) {
+        return ((FloatType) from).bitWidth() == ((BitsType) to).bitWidth();
+      }
+    }
+
```
flofriday: **question:** Same implicit debate here.

vadl/main/vadl/types/BuiltInTable.java
BadGraphixD: This is not yet implemented. I just add a return method so that no errors are thrown because of a missing return method. The float `*s` built-ins will be needed in the future and I have added an example implementation. I should probably make the `TODO` explain this.

vadl-test/resources/cosim_scripts/riscv/compiler.py
BadGraphixD: No, this comment is a result of copy-paste and should not be here. Andreas asked me to remove it.

sys/risc-v/rv64fd.vadl
BadGraphixD: I have added `rv64f.vadl` and `rv64d.vadl`. Both extend `rv64fd.vadl`.

vadl-frontend/main/vadl/ast/ConstantEvaluator.java
BadGraphixD: No. Currently only `fcvt` has type params, and all float builtins cannot be constant evaluated. They could be, but I see no point.

vadl-frontend/main/vadl/ast/TypeChecker.java
BadGraphixD: No, `FloatType` should always be explicitly cast to `Bits`

vadl-frontend/main/vadl/ast/TypeChecker.java
BadGraphixD: Same answer as above.

vadl/main/vadl/types/FloatType.java
Jozott00: Hmm, personally I'm fine with both. Considering that `BoolType` does also not inherit from `BitsType` it might make sense to do the same for `FloatType`.

However, I would definitely make `FloatType` extend `DataType`, which is the common superclass of all types that have some bit representation and can be stored in registers.

BadGraphixD: `FloatType` now extends `DataType`.

vadl-frontend/test/vadl/ast/FrontendIntegrationTest.java
```diff
@@ -43,6 +43,8 @@ public class FrontendIntegrationTest {
       "../sys/risc-v/rv64im.vadl",
       "../sys/risc-v/rv64v.vadl",
       "../sys/risc-v/rvcsr.vadl",
+      "../sys/risc-v/rv64f.vadl",
+      "../sys/risc-v/rv64d.vadl",
```
flofriday, line 47: **praise:** Nice addition

flofriday (approved):
LGTM

flofriday (approved):

PR #1077, opened 2026-08-27
frontend: Support reverse search in advanced AST dump

Jozott00:
Allows to hover over source code on right side to get the corresponding AST node on the left side.

Screenshot 2026-08-27 at 16 17 07

The Java side rendered builds a sorted list of source locations and their corresponding AST node ids as multiple arrays, which are traversed using binary search at hover time using javascript. The representation design was used to minimize loading time when opening the HTML dump in the browser.


vadl-cli/main/vadl/cli/BaseCommand.java
flofriday, line 1: **praise:** Nice catch

flofriday (approved):
LGTM 🔩🎉

PR #1078, opened 2026-08-27
frontend: Optimize advanced AST dump

Jozott00:
Lazily loads expansion locations in an optimized way and optimizes AST layout that is not visible on screen.

Reduces big dumps by 50% and significantly reduces load and usages lags.


flofriday (approved):
LGTM, maybe this helps with the aarch32 dumps too

PR #1083, opened 2026-08-30
iss: Add lazy float exception flag computation

BadGraphixD:
Adds lazy registers (could also be named "virtual registers"). Lazy registers contain parts, which are not stored or updated in the TCG global of the register. These parts are instead sourced from other parts of the CPU-state (e.g. the float_status). Reading/writing these registers must go through the C getters and setters in `cpu.c`. Thus, translation functions must emit helper calls to read/write. This is an optimization to avoid emitting costly logic for updating registers, which are often written inside helpers, but are rarely read.

Lazy registers are now used to speed up float exception flags. Instead of converting between QEMU's flag format to RISC-V's format after every float instruction, we now just store QEMU's format in the CPU-state. When the flags are read, the FCSR register is constructed from that.

### DIFF
--- a/vadl/main/vadl/iss/passes/extensions/RegInfo.java
+++ b/vadl/main/vadl/iss/passes/extensions/RegInfo.java
@@ -189,18 +189,6 @@ public class RegInfo extends DefinitionExtension<RegisterTensor> implements Rend
     );
   }
 
-  private List<Map<String, String>> feFlags(
-      Function<FloatFlagAnnotation, Map<Integer, FloatExceptionFlag>> flags) {
-    if (!reg().hasAnnotation(FloatFlagAnnotation.class)) {
-      return List.of();
-    }
-    var ann = reg().expectAnnotation(FloatFlagAnnotation.class);
-    return flags.apply(ann).entrySet().stream().map(e -> Map.of(
-        "idx", Integer.toString(e.getKey()),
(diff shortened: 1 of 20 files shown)


vadl/main/vadl/iss/passes/extensions/RegInfo.java
```diff
@@ -92,6 +92,30 @@ public enum ExecClass {
...
   }
 
+  /**
+   * Determines how much of the register is lazy.
+   *
+   * <p>Some parts or the whole register may be lazy. Lazy parts are not stored or updated in
+   * the TCG global. Instead, they are sourced from other parts of the CPU-state (e.g. from
+   * the {@code float_status}). Thus, no TCG globals are created for registers containing lazy
+   * parts. They are read/written via CPU helpers. When a register is fully lazy, there is no
+   * need to generate a field for it in the CPU-state.
```
Jozott00, line 102: So lazy parts are computed on demand from other CPU states, but there is no "backing field" in the CPU state that are assigned to those parts?

Jozott00 (approved):
LGTM but I have a bit of a hard time image the actual output for lazy registers.

Could you add some examples to the PR description (read/write/storage/usage) that shows how they are emitted in comparison to the default approach (and where the exact runtime benefit comes from). 

Also I noticed that the openvadl documentation currently has no information about floating point and its annotations. Could you add this (probably in a separate PR)?

PR #1085, opened 2026-08-31
gcb: Add explicit diagnostic for missing ABI

AlexanderRipar:
Previously, when attempting to run gcb on a specification without ABI definition, it would attempt to `get()` an empty `Optional`. This commit still lets that operation fail, but now does so by throwing a proper `Diagnostic` - even if it does not include a source location, since it is reporting the absence of a definition.


Jozott00:
This is an long standing issue, as we never implemented the planned `compiler` definition which would serve as the entrypoint and would reference both, the ISA and ABI. 

I will create an issue for it, because this is something pretty fundamental.

AlexanderRipar:
Ok, so we can ignore this PR then?

Jozott00:
Yeah, essentially. I would instead implement #1086

AlexanderRipar:
Closing this for now in favour of a future implementation of #1086.

PR #1089, opened 2026-09-01
frontend: Rework handling of AST child nodes to no longer be annotation-based

AlexanderRipar:
This PR reworks the way nodes declare their children in the AST, replacing the `@Child` annotation with explicitly overridden `forEachChild` methods.

There are two main reasons for this:
- The annotation-based approach is not easy to follow when debugging.
- The annotation-based approach ignored `@Child` fields in classes that did not themselves declare `@Child` fields, leading to a bunch of ugly and hard-to-maintain workarounds.

The main downside of the new approach is that it is quite a bit more (but also simpler, less edgecase-prone) code to maintain.


flofriday:
Honestly, great work, seeing it all laid out like this, it really highlights all the workarounds I've learned to ignore over the years.

I'm also interested in the benchmark performance so I triggered a non-master run for the frontend benchmarks [here](https://github.com/OpenVADL/openvadl/actions/runs/33616722643).

Great that still all tests pass, so I'm quite confident that it works just as it did before.

flofriday (commented):
Reviewed IRL, ready to merge this  soon :)

vadl-frontend/main/vadl/ast/nodes/RecursiveAstVisitor.java
```diff
@@ -84,17 +84,4 @@ public final Void visitExpression(Expr expr) {
     afterTravel(expr);
     return null;
   }
-
-  @Override
-  public Void visit(FloatTypeDefinition definition) {
```
flofriday, line 89: **praise:** I'm so happy to get rid of this hack 🎉

flofriday (approved):
Yes, this is easier than the previous implementation, well done 🎉

PR #1090, opened 2026-09-02
cli: Add propper unicode support on windows (no `?`)

flofriday:
~~On windows the JVM replaced all unicode with question marks because it wasn't sure if the stdout stream can accept unicode. Now we force the JVM to let unicode characters through.~~

**Update:** the previous attempt setting the output stream to utf-8 didn't always work so now this PR implements a fallback with pure ASCII characters.
The fallback is automatically chosen if the output cannot encode unicode characters.


flofriday:
@AlexanderRipar can you verify, because I don't own a Windows machine :)

flofriday:
After an initial check this isn't quite it because the code page the terminal has might still not be prepared for UTF-8 and would require the user to execute `chcp` beforehand. 
Let's rethink that approach.

flofriday:
### For the reviewer. 
- [ ] In an old style cmd does it now use the ASCII fallbacks instead of the question marks form before?
- [ ]  After manually changing the page with `chcp 65001` does it use unicode? (Did it do that also before this change?)
- [ ] Do the tests still run on Windows? `gradle :vadl-frontend:test`

vadl/test/vadl/ArchitectureTest.java
```diff
@@ -60,7 +61,11 @@ void shouldComplyWithArchitectureDesign() {
   void shouldNotUseSystemOut() {
     JavaClasses jc = new ClassFileImporter()
         .withImportOption(new ImportOption.DoNotIncludeTests())
-        .importPackages("vadl");
+        .importPackages("vadl")
+        // Exception for diagnostic printer because even though it doesn't use System.out to print
+        // it uses it to detect fallback on Windows, when it cannot output unicode.
+        .that(are(not(equivalentTo(DiagnosticPrinter.class))));
+        ;
```
flofriday: **note:** I don't love this workaround but I haven't yet come up with a better structure.

vadl/main/vadl/error/DiagnosticPrinter.java
```diff
@@ -145,14 +146,17 @@ private void printHeader(Diagnostic diagnostic, StringBuilder builder) {
    */
   private void printMultiSourcePreview(Diagnostic diagnostic, StringBuilder builder) {
     // Print preview header
-    builder.append("     %s╭── %s\n".formatted(colors.cyan(),
+    var topBorder = unicodeOrFallback("╭──", "---");
```
AlexanderRipar: I think this could be "+--", to get a *"nice"* corner.

vadl/main/vadl/error/DiagnosticPrinter.java
```diff
@@ -183,10 +187,14 @@ private void printMultiSourcePreview(Diagnostic diagnostic, StringBuilder builde
    */
   private String sourceDelimiter(SourceLocation previous,
                                  SourceLocation next) {
+    var topBorder = unicodeOrFallback("╭─", "--");
```
AlexanderRipar: Same here, "+-"

AlexanderRipar (commented):
Looks good on my (windows) end, the frontend tests still pass as well, so the forceUnicode option appears to be doing its job too. Just the tiny formatting touchups I mentioned above (unless there is some reason against them).

vadl/main/vadl/error/DiagnosticPrinter.java
flofriday: fair, i wanted to replicate rusts formatting but i only did it halfway so yess I'll change it :)

AlexanderRipar (approved):
Looks Great!

PR #1091, opened 2026-09-02
util: Improve sourcelocation outermost expansion access

flofriday:
A easier (and faster) way to access the outermost location (where the outermost macro was called)

### DIFF
--- a/vadl/main/vadl/utils/SourceLocation.java
+++ b/vadl/main/vadl/utils/SourceLocation.java
@@ -196,16 +196,59 @@ public sealed interface SourceLocation extends WithLocation, Comparable<SourceLo
   }
 
   /**
-   * Strips a location of all expandedFrom information and returns it as a direct location.
+   * Strips a location of all its macro expansion information and returns it as a
+   * {@link DirectLocation}.
+   *
+   * <p><pre>{@code
+   *  model inner(): Ex = {
+   *    42
+   *  }
+   *
+   *  model outer(): Ex = {
+   *    $inner
+   *  }
+   *
+   *  $outer()
+   * }</pre>
+   * In this example the {@code innermostDirectLocation} of the integer literal would be the one in
+   * the second line inside the inner model.
    *
    * @return the direct location.
    */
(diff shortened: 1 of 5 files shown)


jaraonthe-dot-net:
It seems you have checkstyle warnings.

vadl/main/vadl/utils/SourceLocation.java
jaraonthe-dot-net: **todo**: Javadoc description should be updated to reflect the new method name (it is not incorrect per se, but it may not be obvious how this relates to the name).

vadl/main/vadl/utils/SourceLocation.java
```diff
@@ -200,13 +200,27 @@ default SourceLocation join(SourceLocation other) {
...
-  default DirectLocation asDirectLocation() {
+  default DirectLocation innermostDirectLocation() {
     return switch (this) {
       case DirectLocation direct -> direct;
       case ExpandedLocation expanded -> expanded.primaryLocation;
     };
   }
 
+  /**
+   * Get's the outer most macro invocation (the top-level call).
```
jaraonthe-dot-net: **nitpick**: *Gets

vadl/main/vadl/utils/SourceLocation.java
```diff
@@ -200,13 +200,27 @@ default SourceLocation join(SourceLocation other) {
...
+   * If the location is not from a macro call it is just the location.
```
jaraonthe-dot-net: **nitpick**: Not a fan of the phrasing. Consider something like: "If the location is not the result of a macro call this just returns the location itself.

jaraonthe-dot-net (changes requested):

vadl/main/vadl/utils/SourceLocation.java
flofriday: I've added examples to clarify, but if that's not it than I'm out of clues with what else could help 😅

jaraonthe-dot-net (approved):

PR #1092, opened 2026-09-02
lsp: Updates are delayed on Windows

### LINKED ISSUE #1087: lsp: Updates are Delayed on Windows
# Failure Mode

When running the OpenVADL LSP on Windows updates are heavily delayed. The delay does not seem to be purely timing-based, as updates are only (inconsistently) performed once additional changes are made. This failure mode does not reproduce on @flofriday's device, so we theorize that it might be related to the OS.

Note that using the VS Code plugin results in an error message `OpenVADL server terminated: The language server exited prematurely. spawn EINVAL`, followed by `OpenVADL Language Server client: couldn't create connection to server.`.

Some of the inspected files where LSP exhibits the described laggy behaviour in IntelliJ are:
- ./vadl-frontend/test/resources/frontend-snapshots/lowering/baselineIsa.vadl
- ./sys/risc-v/rv32i.vadl
- ./sys/risc-v/mia/rv_1stage.vadl
- ./sys/risc-v/mia/rv_3stage.vadl
- ./sys/risc-v/mia/rv_5stage.vadl

The connection attempts were made without using TCP.

# System

- **OpenVADL LSP:** OpenVADL 0.1.0
- **IntelliJ:** 2026.2.1 Build #IU-262.9437.185, built on August 9, 2026
- **VM:** OpenJDK 64-Bit Server VM by JetBrains s.r.o.
- **OS:** Windows 11 Home Version 10.0.26200 Build 26200

jaraonthe-dot-net:
Closes #1087

- Using Path instead of String URI throughout; this avoids encoding ambiguities that break equality of file URIs on Windows. E.g. "c:" may be encoded as "c%3A" in a URI.


jaraonthe-dot-net:
@AlexanderRipar Please verify if this fixes your Windows issues.
@flofriday Please do a regular review.

AlexanderRipar:
This seems to work great!

However, while testing whether it withstands a `%3A` in the filename (it does), I think I found another bug. When you rename a file without saving and then make changes to it, they get dropped, just like in the failure mode addressed by this PR.
I mention this here instead of in a new issue since I believe keying your map of changed files by a unique file identifier would sidestep the problem addressed by this issue - uri / path equality - while also addressing this rename problem since such unique ids remain constant even after a (proper) rename.
By unique file identifier I mean `inode` under *nix and `nFileIndex` + `dwVolumeSerialNumber` of [`BY_HANDLE_FILE_INFORMATION`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/ns-fileapi-by_handle_file_information) under Windows.

If you'd rather treat this new bug separately I'll of course create a separate issue for it.

flofriday:
I think the renaming is just slightly out of scope and this fix already makes the LSP behave better so I created a new issue for it: #1094

vadl-lsp/main/vadl/lsp/LspUtils.java
```diff
@@ -18,12 +18,11 @@
 
 import java.net.URI;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 
 class LspUtils {
 
   static Path toPath(String uri) {
-    return Paths.get(URI.create(uri));
+    return Path.of(URI.create(uri));
```
flofriday, line 25: **question:** Don't quite see how or why this one would create a fix for windows but I guess it's good to use the newer APIs.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -179,7 +179,7 @@ public void didSave(DidSaveTextDocumentParams params) {
       }
       var originSelectionRange = document.calculateUtf16Range(identifier.location());
 
-      return definitionResult(targetDocument.uri, targetRange, targetSelectionRange,
+      return definitionResult(toUri(targetDocument.path), targetRange, targetSelectionRange,
```
flofriday: **nitpick:** I think I would move the toUri into definitionResult as in the domain of the LS we now only deal with paths and only the definitionResult  builder would have to deal with URIs.

But you can go either way and resolve this comment.

flofriday (approved):
LGTM

Is there any way we can add a test to with the exact encoded windows path provided by the client that @AlexanderRipar got?

vadl-lsp/main/vadl/lsp/LspUtils.java
jaraonthe-dot-net: This does indeed not contribute to the bugfix. I just got the impression that `Path.of()` is the preferred method nowadays (`Paths.get()` forwards to it).

flofriday (approved):

PR #1095, opened 2026-09-04
lsp: Restrict useless concurrency

### LINKED ISSUE #995: lsp: Restrict useless concurrency

jaraonthe-dot-net:
Closes #995.


vadl-frontend/main/vadl/ast/Frontend.java
```diff
@@ -67,6 +69,86 @@ public static Ast compileToAst(Path path, VirtualFileSystem fileSystem) {
...
+
+  /**
+   * Part of {@link BestEffortCompilation}.
+   */
+  public enum AstPass {
+    NONE(0),
+    PARSED(1),
+    MODELS_REMOVED(2),
+    UNGROUPED(3),
+    PARTIALLY_TYPE_CHECKED(4),
```
flofriday: **minor:** I think the naming `completedPass` combinded with partial typechecking is potentially confusing.

The code could be cleaner by changing the semantics from completed pass to last pass and if the pass failed it can be indicated by diagnostics having a non-null value.

vadl-frontend/main/vadl/ast/Frontend.java
flofriday: **chore:** This was originally added for the LSP but since it didn't quite fit and the new implementation is better let's remove it.

flofriday:
I've started working on a review, but I'm having a hard time reviewing this.

Can you please explain how this change reduces the concurrency? Also what part of it is "useless"?
And how it tackled the problem Andi had months ago where due to a bug in the compiler where a compilation suddenly took minutes the LSP pinned all of his cores to 100%?

Also go a bit into the reasoning here for the new structure.

jaraonthe-dot-net:
> Can you please explain how this change reduces the concurrency? Also what part of it is "useless"?

A file is not recompiled for every LSP request, but the compilation result is kept as long as it is deemed unchanged (i.e. until the file or a file it depends on is modified). Especially on big files, this speeds up the LSP responses, avoids doing the exact same work (the exact same compiler run) more than once. The second part is that compiler runs may be interrupted if they are not finished yet but already obsolete (because a modification occurred in the meantime).
So we are avoiding "useless" work, and that's what the title refers to.

> And how it tackled the problem Andi had months ago where due to a bug in the compiler where a compilation suddenly took minutes the LSP pinned all of his cores to 100%?

The measures described above mitigate this issue, but may not fully solve it. This is because Java does not support pre-empting Threads, i.e. a thread cannot be aborted immediately from the outside; the thread has to check its own interrupted state and stop itself accordingly. This means adding a lot of these checks all over the codebase (ideally in locations that are executed many times, e.g. loops), which is a very tedious task. Thus, in a weekly meeting some weeks ago, it was decided to only do these checks in between compiler passes for now (see the new `Frontend.compileToAstBestEffort()`).

If this is not good enough, then further interrupted checks (see `Frontend.throwIfInterrupted()`) can  be added in various positions throughout the compiler, but that's all the change that's required as the ground work has already been done in this PR.

> Also go a bit into the reasoning here for the new structure.

The new `DocumentCompilation` class manages the current compilation for a particular file. It simply returns the current compilation result or triggers a new compiler run and makes threads wait for it to complete (via `Future.get()`). The `VadlTextDocumentService` tells `DocumentCompilation` when a document's compilation is outdated; this happens when the document itself or a document it depends on is modified. In that case the underlying Future is cancelled, which (as far as I understand it) immediately interrupts the threads waiting for this Future and also sets the Future's thread's interrupted state (with the caveat described above).

`DocumentCompilation` shall only be used by `VadlTextDocumentService`, but as that file is already rather large, I took the opportunity to extract this part into a new file. This does not change how `VadlTextDocumentService` manages the current content of documents (the immutable `Document` instances stored in `openDocuments`); I think this makes for better separation of concerns, and avoids having to iterate through the entire `documentCompilations` map every time a new `LspSnapshotFileSystem` is created.

Indeed, `VadlTextDocumentService` manages both `openDocuments` and `documentCompilations` in sync, updating the document contents in the first and clearing compilations in the second. `openDocuments` is used as lock for both; I've designed the `synchronized` blocks carefully to ensure data consistency and making deadlocks impossible (that's why the clearing of compilations is done outside of the `synchronized` blocks, otherwise `openDocuments` and the `DocumentCompilation` itself would be locked at the same time but in different order (the other path being `DocumentCompilation.startCompilation() -> VadlTextDocumentService.createSnapshotFileSystem()`)). There may be better ways to manage this concurrency in a non-locking fashion, but considering that everything is driven by user input which is comparatively slow and non-concurrent, the amount of locking should be fine.

Every Task uses a virtual thread, so thread churn is not a problem.

vadl-lsp/main/vadl/lsp/VadlTextDocumentService.java
```diff
@@ -84,38 +84,57 @@ public void didOpen(DidOpenTextDocumentParams params) {
...
+      documentCompilation = documentCompilations.remove(path);
+    }
+
+    if (documentCompilation == null) {
+      return;
     }
+    clearDependentCompilations(documentCompilation);
+    documentCompilation.clearCompilation();
+
+    publishDiagnosticsForDependentDocuments(documentCompilation);
```
flofriday: **question:** Why are we publishing diagnostics when a document is closed?

flofriday:
I give up reviewing this code.
Let's talk about this tomorrow or let's find someone else who will do it.
I can no longer reason which objects protects which ressources, and the tight coupling between classes has exceeded my comprehension.

I've spent multiple hours on it and I'm still not certain it's correct.

Personally, I would try to radically redesign the LSP:
- Reduce coupling by classes that do one thing and one thing well. Like a Compilation store that caches compilation results and lazily computes them when needed.
- Use concurrent datastructures where they make sense instead of reinventing them.
- Use explicit Mutex/Lock data structures otherwise instead of locking random object where their name doesn't indicate whether not they need to be locked to be used.
- Split the computation of results more from logging and LSP message creations.
- Maybe remove some features again like the dependency mapping, I (personally) think it's fine if only open documents are handled and OpenVADL currently doesn't have a great way to detect entrypoints so. Solving this first at a language level could remove a lot of complexity in the LSP for what always is a rough approximation anyway.
- Maybe even going so far as removing the whole concurrency story, focusing on the features needed first, writing proper request-response logging for benchmarking and adding the concurrency back in once we have data needed (you know the whole "Make it work, make it right, make it fast".

jaraonthe-dot-net:
> * Reduce coupling by classes that do one thing and one thing well. Like a Compilation store that caches compilation results and lazily computes them when needed.

I agree that `VadlTextDocumentService` is becoming unwieldy. It probably makes sense to extract some sort of DocumentStore, which takes over management of current Document state (incl. their Compilations). That would mean that the synchronized blocks are contained there and no longer affect the complexity of `VadlTextDocumentService`.

> * Maybe remove some features again like the dependency mapping, I (personally) think it's fine if only open documents are handled and OpenVADL currently doesn't have a great way to detect entrypoints so. Solving this first at a language level could remove a lot of complexity in the LSP for what always is a rough approximation anyway.

All this feature does is force recompilation (and regenerating diagnostics) of open files if files they import have been modified. So this only affects open documents, and doesn't depend on an entrypoint mechanism or other language specifics. IMO, the dependency map is working well, I see no reason to remove it.

Jozott00:
I read a bit over the LSP and agree with @flofriday that the LSP has quite a bit of coupling and its concurrency mechanism might be hard to validate and especially expand in future. 

Maybe we should go a step back and come up with an abstraction that allow us to expose concurrency boundaries cleanly with single responsibilities. This will help us reason about correctness and makes the LSP easier to extend. I already have some concepts in mind that allows us to 
1. Share compiler analysis between LSP invocations
2. Bound analysis computation to a limited number of threads
3. Making cancellation propagation easier
4. Providing a safe and minimal interface for concurrent invocations
5. Handling workspace/source versioning/snapshoting

I will try to create a detailed discussion this evening.

benjaminkasper99:
### First, regarding the issue (#995) this PR addresses:

The two measures implemented to reduce concurrenct tasks are well described in the comment by @jaraonthe-dot-net:
> A file is not recompiled for every LSP request, but the compilation result is kept as long as it is deemed unchanged (i.e. until the file or a file it depends on is modified). Especially on big files, this speeds up the LSP responses, avoids doing the exact same work (the exact same compiler run) more than once. The second part is that compiler runs may be interrupted if they are not finished yet but already obsolete (because a modification occurred in the meantime).

Simply put: results of compilation tasks are chached and compilation tasks can be early exited, if their result would be outdated anyways. In my understanding these two measures _do_ lower the risk of the LSP claiming all CPU cores, but don't eliminate the possibility of such a scenario. If a hard bound on the number of threads desired, I think we would have to use a `fixedThreadPool` (or manually set a `maxPoolSize`).


### Second, regarding the used concurrency constructs:
The `VadlTextDocumentService` uses `synchronized(openDocuments)` blocks to manage concurrent accesses to two maps `openDocuments` and `documentCompilations`. If I understand correctly, all seven blocks in this class ensure one or both of two properties:
1. Avoid corrupted state per map in case of concurrent writes / reads, e.g.:
```
synchronized (openDocuments) {
    return documentCompilations.get(path);
}
```
2. Keeping the two maps in "lockstep", e.g.:
```
synchronized (openDocuments) {
    openDocuments.remove(path);
    documentCompilation = documentCompilations.remove(path);
}
```

My suggestion would be to remove these `synchronized` blocks, by using a single `ConcurrentHashMap` where entries are records of a `DocumentState` type, something along the lines of:
```
record DocumentState(Document document, DocumentCompilation compilation);

ConcurrentHashMap<Path,DocumentState> documentStates = new ConcurrentHashMap<>();
```

With this approach property 1. is ensured by the `ConcurrentHashMap` implementation and property 2. is ensured by having a single record which keeps documents and compilations per path in "lockstep".


### Third, regarding maintainability of the LSP:
I agree with the consensus of the comments of this PR, that structure of the LSP could be improved.
For me the most pressing issue is the `VadlTextDocumentService`. This class is a mix of LSP message building and concurrent task management, two separate concerns which should therefore be in separate classes.

PR #1096, opened 2026-09-04
frontend: Make `ConstantAnnotation`s resolve identifiers in their value

AlexanderRipar:
Super-simple fix for the lack of name resolution in the values of annotations of type `ConstantAnnotation`.

### DIFF
--- a/vadl-frontend/main/vadl/ast/Annotation.java
+++ b/vadl-frontend/main/vadl/ast/Annotation.java
@@ -53,38 +53,18 @@ public abstract class Annotation implements AnnotationDeclaration, WithLocation 
   /**
    * Called by the symbol resolver to resolve the subparts of the annotation.
    *
-   * <p>Can be overwritten by subclasses to specify additional
-   * annotation-specific behavior to execute during name-resolution.
-   *
-   * <p>In general, subclasses overwriting this method should call
-   * {@code super.resolveName} so that names in the annotation's values are
-   * resolved. If this is explicitly *not* desired, the {@code super} call must
-   * be omitted.
-   *
    * @param definition The `AnnotationDefinition` corresponding to this `Annotation`.
    * @param resolver The active name resolver.
    */
-  void resolveName(AnnotationDefinition definition, SymbolTable.SymbolResolver resolver) {
-    definition.values.forEach(value -> value.accept(resolver));
-  }
+  abstract void resolveName(AnnotationDefinition definition, SymbolTable.SymbolResolver resolver);
 
   /**
    * Called by the type checker to type check the annotation.
    *
-   * <p>Can be overwritten by subclasses to specify additional
-   * annotation-specific behavior to execute during typechecking.
-   *
-   * <p>In general, subclasses overwriting this method should call
-   * {@code super.typeCheck} so that the annotation's values are also
-   * typechecked. If this is explicitly *not* desired, the {@code super} call
-   * must be omitted.
-   *
    * @param definition The `AnnotationDefinition` corresponding to this `Annotation`.
    * @param typeChecker the active typechecker.
    */
-  void typeCheck(AnnotationDefinition definition, TypeChecker typeChecker) {
-    definition.values.forEach(typeChecker::check);
-  }
+  abstract void typeCheck(AnnotationDefinition definition, TypeChecker typeChecker);
 
   @Override
   public String name() {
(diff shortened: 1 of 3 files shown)


flofriday (approved):
LGTM 🎉

Oh I guess, so far we only had expressions here without identifiers and so we never needed it.
Can you take a quick look if there are other annotations that are missing that step? And if so either do it also here or add an issue to keep track of them.

AlexanderRipar:
I've tried to ensure that all annotations that can resolve and typecheck their values do so by making `Annotation.resolveName` and `.typeCheck` non-abstract and implementing value traversal as the default. However the individual overrides still seem like an absolute mess. Not sure how to proceed, or whether further cleanup is even warranted.

Also just realized that the exceptions currently thrown from `resolveName` when the value count is not as expected seem to prevent the compiler resolving / typechecking, so maybe there is also a better place to throw those.

Jozott00:
It was a design decision to not apply type checking and symbol resolving by default.
Having a valid expression in an annotation is not the default case. Also there are annotation where we want to restrict the symbol scope. 

So I think letting the Annotation classes opt-in to type checking/symbol resolving their annotation is the better option.

EDIT: A better option would be to introduce an other abstract annotation class that explicitly enables this default behavior. Then every annotation class that wants this, could inherit from it.

AlexanderRipar:
> It was a design decision to not apply type checking and symbol resolving by default. Having a valid expression in an annotation is not the default case. Also there are annotation where we want to restrict the symbol scope.
> 
> So I think letting the Annotation classes opt-in to type checking/symbol resolving their annotation is the better option.
> 
> EDIT: A better option would be to introduce an other abstract annotation class that explicitly enables this default behavior. Then every annotation class that wants this, could inherit from it.

Ok, that sounds like a decent solution. I'll try and implement that.

EDIT: Nevermind, I went back to just running name resolution and typechecking separately for each annotation class. Adding another abstract class into the hierarchy just to implement two one-liners which then need to be called via `super` anyways seems like overkill to me.

vadl-frontend/main/vadl/ast/Annotation.java
```diff
@@ -51,18 +51,18 @@ public Annotation(boolean allowMultiple) {
   }
 
   /**
-   * Called by the symbol resolver to resolve parts of the annotation.
+   * Called by the symbol resolver to resolve the subparts of the annotation.
    *
-   * @param definition to be resolved.
-   * @param resolver   who resolves the annotation.
+   * @param definition The `AnnotationDefinition` corresponding to this `Annotation`.
+   * @param resolver The active name resolver.
```
flofriday, line 57: **question:** I might be biased here but what was confusing about _parts_ -> _subparts_ ?
And also why is it important to name _active_ here? 

Obviously I designed these things so it was obvious for me, but I think we never define what active is (mostly because there is never a need to create an inactive anything)

vadl-frontend/main/vadl/ast/AnnotationTable.java
```diff
@@ -1132,16 +1137,18 @@ public EnableAnnotation() {
   @Override
   void resolveName(AnnotationDefinition definition, SymbolTable.SymbolResolver resolver) {
     verifyValuesCntBetween(definition, 0, 1);
+
+    definition.values.forEach(value -> value.accept(resolver));
   }
 
   @Override
   void typeCheck(AnnotationDefinition definition, TypeChecker typeChecker) {
+    definition.values.forEach(typeChecker::check);
```
flofriday: **nitpick:** I don't love this change, I think it was clearer beforehand, since there ever can be 0-1 values here.

What motivated the change? If it's just for astetic reasons. I think we could improve this code by inverting the condition and adding an early return.

vadl-frontend/main/vadl/ast/Annotation.java
AlexanderRipar: This is just a leftover of a larger change that I ended up reverting, so there's no deeper meaning. I think "parts" and "subparts" are both equally expressive here, maybe just "parts" is a tiny bit clearer though ("subparts" somewhat implies parts of parts).

vadl-frontend/main/vadl/ast/AnnotationTable.java
AlexanderRipar: Ok, I think I agree. I'll change it back in the cases where we know the number of values beforehand.

AlexanderRipar: Done as of 6e267e2678208255b1d8b7b65c89f9b5f4633cab.

flofriday (approved):

PR #1110, opened 2026-09-17
frontend: Add `logic [reservation station]` and its annotations

AlexanderRipar:
This PR adds a new logic element, `logic [reservation station]` as a starting point for modelling Out-of-order execution. This logic already appears in older existing documentation, but has seen some changes to its annotations as outlined in #1046 and #1047. As of this PR, `reservation station` is only implemented in the frontend up to and including lowering to the VIAM.


AlexanderRipar:
I'm struggling a bit to get the desired semantics on some of the annotations. Namely, I need both `[size : N]` and `[source : STAGE.OUTPUT]` to be present on a reservation station, i.e. they are "required" annotations. Is there a mechanism for this yet?

I also want each annotation to be present at most once. I haven't found a good way to enforce that yet either. Is there one yet or do I need to implement?

These two points are why this is just a draft for now, as the newly added tests fail because of them.

vadl-frontend/test/resources/frontend-snapshots/annotation/reservationStationSource.vadl
```diff
@@ -0,0 +1,54 @@
...
+  // Incorrect: Wrong stage output type for value of `source` annotation.
+  [size : 1]
+  [source : Dummy.val]
+  logic [reservation station] RS4
+
+  // Incorrect: Missing mandatory `source` annotation.
+  [size : 1]
+  logic [reservation station] RS5
+
+  // Incorrect: Duplicate `source` annotation.
```
AlexanderRipar, line 32: @flofriday This is an example of an annotation that should only appear once being provided twice, and not being reported as an error.

flofriday:
Honestly, I thought the double annotation checking was already implemented, but I guess we must have missed it. 🤷‍♂️
I cannot think of any good idea why we should allow it so please feel free to try implement a check for that.

We already do have checks for groups of annotations but not the same annotation.
Like here only the first one fails but the second one is fine.

Screenshot 2026-09-20 at 12 03 06

vadl-frontend/test/resources/frontend-snapshots/annotation/reservationStationSource.vadl
flofriday: I see, I've answered in another comment.

flofriday:
As for _required annotations_ I personally really dislike the concept as a whole.
We designed the annotations as additional syntax for definitions that could also stand on their own.
So no, there is no way to make those work right now, and usually this indicates that the core definition should be expanded.
Can you mention this tomorrow in the meeting and we'll try to find a solution there? ☺️

AlexanderRipar:
Alright, I'll try my luck with implementing the no-duplicate-annotations logic in a separate PR.

I feel that avoiding required annotations makes sense in general, but at least for logic elements in the MiA they seem necessary, as there are often no sensible defaults for core attributes of logic elements, and annotations are the only place to provide values. I'll definitely bring it up tomorrow.

vadl-frontend/main/vadl/ast/AnnotationTable.java
```diff
@@ -498,6 +505,111 @@ public class AnnotationTable {
...
+    annotationOn(LogicDefinition.class, "size", ConstantAnnotation::new)
+        .applyViam((def, annotation, lowering) -> {
+          def.ensure(
+              def instanceof Logic.ReservationStation,
+              "`size` annotation is not supported on `%s` logic elements.",
+              def.getClass().getSimpleName()
+          );
+
+          annotation.verifyGreaterThan(BigInteger.ZERO);
+          annotation.verifyLessThanEqual(BigInteger.valueOf(Integer.MAX_VALUE));
```
flofriday: **minor:** I think this could be in the typechecking phase.
then we'd also get these errors in typechecking, so that it works in the LSP.

Also in the frontend we try to avoid `ensure` as it throws an annotation and that usually means we abort processing, which is bad for fault tolerance (as we would exit on the first error).

vadl-frontend/main/vadl/ast/AnnotationTable.java
```diff
@@ -498,6 +505,111 @@ public class AnnotationTable {
...
+    annotationOn(
+        LogicDefinition.class,
+        "filter",
+        () -> IdentifersAnnotation.single(OperationDefinition.class)
+    ).applyViam((def, annotation, lowering) -> {
+      def.ensure(
+          def instanceof Logic.ReservationStation,
+          "`filter` annotation is not supported on `%s` logic elements.",
+          def.getClass().getSimpleName()
+      );
```
flofriday: **minor:** Same here

vadl-frontend/main/vadl/ast/AnnotationTable.java
```diff
@@ -498,6 +505,111 @@ public class AnnotationTable {
...
+                    stageId,
+                    "First element of the value of the `source` annotation was `%s` must a `%s`.",
+                    requireNonNull(stageId.target).nodeName(),
+                    Node.nodeNameFor(StageDefinition.class)
+                ).locationDescription(
+                    stageId.target,
+                    "`%s` defined here.",
+                    stageId.name
+                ).build();
+          } else {
```
flofriday, line 571: **nitpick:** The `else if` and `else` _could_ be removed here.

This is mostly because we always throw here so just `if` could work here and _keep the happy path left_.
This is just taste so feel free to keep it this way ;)

vadl-frontend/main/vadl/ast/AnnotationTable.java
```diff
@@ -498,6 +505,111 @@ public class AnnotationTable {
...
+    annotationOn(LogicDefinition.class, "source", ExprAnnotation::new)
+        .check((def, annotation, tc) -> {
+          annotation.verifyExprType(tc, MicroArchitectureType.instruction());
+
+          if (!(annotation.expr instanceof CallIndexExpr expr)
+              || !(expr.target instanceof Identifier stageId)
+              || expr.subCalls.size() != 1
+              || !(expr.subCalls.getFirst().id instanceof Identifier outputId)
+          ) {
+            throw error("Invalid source Annotation Value", annotation.expr)
```
flofriday, line 554: **minor:** I don't think we should `throw` here as it aborts the checking process.
Look into `tc.addErrorAndStopChecking`, and how the typechecker handles annotation processing.

flofriday (commented):
Looks good, great work.

I think we shouldn't throw in the annotation checking, because it aborts the checking process. Instead you can use the `addErrorAndStopChecking` method which stops checking until the next definition.

But maybe that's a good point to also add an invalid reservationStation file that verifies that errors are thrown (maybe even multiple).

vadl-frontend/main/vadl/ast/AnnotationTable.java
AlexanderRipar: I would theoretically prefer it without the `else` as well, but it is required since the `else` body uses the `stage` from the previous branch's `instanceof`.

vadl-frontend/main/vadl/ast/AnnotationTable.java
AlexanderRipar: The diagnostics thrown from inside `check` do get caught by the `TypeChecker` and result in a `StopPartialCheckingSignal`, so the `throw` already does (mostly) what we want it to.

PR #1111, opened 2026-09-17
Chore: Refactor ~~symbol~~ name resolving

flofriday:
Recently @AndreasKrall mentioned that name resolution is the better name for what the front-end does so this PR renames it.
Also name resolution is conceptually its own pass so we also refactored it to its own pass.


.idea/checkstyle-idea.xml
```diff
@@ -9,8 +9,8 @@
     </option>
     <option name="locations">
       <list>
-        <ConfigurationLocation id="bundled-sun-checks" type="BUNDLED" scope="All" description="Sun Checks">(bundled)</ConfigurationLocation>
-        <ConfigurationLocation id="bundled-google-checks" type="BUNDLED" scope="All" description="Google Checks">(bundled)</ConfigurationLocation>
+        <ConfigurationLocation id="bundled-sun-checks" type="BUNDLED" scope="All" description="Sun Checks">bundled-sun-checks</ConfigurationLocation>
```
AlexanderRipar: Is this intentionally included in this PR? If so, it's fine, but I want to make sure.

AlexanderRipar (commented):
Other than the one potential bycatch I commented on this looks good to me (though I admit, my eyes glazed over somewhere in the middle of all the renames in the tests). If it is indeed intentional, then this can be merged.

.idea/checkstyle-idea.xml
flofriday: Not at all, nice catch, let me revert this 👍

AlexanderRipar: Luckily it was the first file in the diff ;)

AlexanderRipar (approved):
Looks good now :)

PR #1112, opened 2026-09-17
Feature/mia dependency graph

AlexanderRipar:
This PR attempts to make the VIAM representation of the MiA's stage graph more traversable.

The first reason for this is that such traversals are currently implemented in an ad-hoc manner in vadl/rtl/passes/StageOrderingPass.java as well as some code that is still local to one of my branches.

The second reason is that it will allow me to integrate logic elements into the stage graph in a more or less seamless way at a future point.

Lastly, this PR also offers a consistency check on the MiA's stage graph - namely that there is exactly one stage that has no inputs. This is already implicitly assumed by the rtl's StageOrderingPass. Once instruction routing is implemented, a check for whether all instructions can reach a stage on which they can be executed also seems reasonable, and I guess there are likely others as well. 

The design of the `MiaDependency` interface was chosen after a discussion today with @linushdot.


vadl/main/vadl/viam/MicroArchitecture.java
```diff
@@ -84,19 +95,140 @@ public MicroArchitecture(Identifier identifier,
...
+      });
+    }
+
+    if (rootStages.size() == 1) {
+      this.rootStage = rootStages.iterator().next();
+      this.potentialRootStages = null;
+    } else {
+      this.rootStage = null;
+      this.potentialRootStages = rootStages;
+    }
```
Jozott00: To me it looks like this constructor has too much initialization logic. Maybe we should extract this into a separate pass that fills/sets the `MiaDependencies`.

vadl/main/vadl/viam/MicroArchitecture.java
```diff
@@ -40,21 +49,17 @@ public class MicroArchitecture extends Definition {
...
-                           InstructionSetArchitecture instructionSetArchitecture,
-                           List<Stage> stages,
-                           List<Logic> logic) {
-    this(identifier, instructionSetArchitecture, stages, logic, new ArrayList<>(),
-        new ArrayList<>(), new ArrayList<>(), new ArrayList<>(), new ArrayList<>());
-  }
+  // Dependencies between MiA elements (stages and logic).
+  private final List<MiaDependency> allDependencies;
+  private final HashMap<Definition, List<MiaDependency>> dependenciesBySource;
+  private final HashMap<Definition, List<MiaDependency>> dependenciesByDestination;
```
Jozott00: Maybe we should add a new internal class called `MiaDependencies` (can be next to the MiaDependency in `MiaDependency.java`) that own those properties and contains the logic of keeping them in sync.

vadl/main/vadl/viam/MicroArchitecture.java
```diff
@@ -40,21 +49,17 @@ public class MicroArchitecture extends Definition {
...
+
+  // Only null if an error was encountered during construction.
+  private final @Nullable Stage rootStage;
+
+  // For error propagation from construction to verification. Only non-null if
+  // an error was encountered during construction.
+  private final @Nullable Collection<Stage> potentialRootStages;
```
Jozott00: The VIAM shouldn't be used for error state propagation. If we move construction to a separate pass, this isn't required. (see comment below)

Jozott00 (changes requested):
LGTM in general, however, I think the VIAM object contains too much logic. Lets split it into a separate dependency analysis pass that fills the dependencies in the VIAM definition.

vadl/main/vadl/viam/MicroArchitecture.java
AlexanderRipar: Sounds good, I'll do that. Regarding the pass order, does it matter where I put it? I don't believe so, but since all the passes in `ViamPassOrder` seem to be grouped into blocks separated by empty lines maybe there is some pattern I am missing.

Jozott00: It probably doesn't matter. However, maybe should add it to the RTL pass-order only, so it doesn't run for the other generators. We will have to factor out all hardware related passes that are used by the compiler or CAS later anyways.

vadl/main/vadl/viam/MicroArchitecture.java
AlexanderRipar: From the discussion with Linus I don't believe there (currently) needs to be logic for keeping the dependencies in sync with the MiA's contents, as the behaviour's stage reads are never changed. However, Linus also expressed some concern for this regarding future-proofness. I guess that the best solution would be to update the dependencies every time a stage's behaviour is changed. However, I don't see a good way of updating the dependencies every time this happens, since I think the behaviour graph can be freely modified. Am I missing something there?

Regarding a new MiaDependencies class, I believe that it would only be a good idea if we decide to go with syncing the stage behaviours and the dependencies.

EDIT: I just realized that `MiaDependency` might be a misleading name. It is actually just dependencies involving the routing of `Instruction`s or `FetchResult`s. Was that clear from the context or should I maybe consider a rename (Maybe `MiaInstructionFlow` or something along those lines)?

Jozott00: I can't say a lot regarding naming. The additional class was more about having a structural grouping of the three cache structures, so there handling is moved to a dedicated entity to improve separation of concerns. However, I'm also fine with the current approach.

vadl/main/vadl/viam/MicroArchitecture.java
AlexanderRipar: I need it in the lcb for instruction scheduling.

vadl/main/vadl/viam/MicroArchitecture.java
AlexanderRipar: Ok, in that case I'd prefer to leave the members where they are. If syncing turns out to be a concern after all, I'll move them then.

vadl/main/vadl/viam/MicroArchitecture.java
Jozott00: Yes. For me its fine to keep it temporarily in the common pass order, but we will have to move all passes used by multiple mia targets into a separate pass order (RTL, LCB and CAS). So those passes are not ran for the ISS.

AlexanderRipar: Ah, I see. Thanks for the clarification! I've put it in what I believe are the right spots now (c63d264c345cfe5870d94947ba5e2c039cf57d86).

Jozott00 (approved):
LGTM

vadl/main/vadl/viam/MicroArchitecture.java
linushdot: @AlexanderRipar I agree that the term dependency maybe is too ambiguous/generic and could relate to other kinds of dependencies the MiA also contains, `MiaInstructionFlow` definitely would be better.

linushdot (approved):
Looks good to me

